Issue #3, opened 2019-06-06
Searching broken within spec PDF output


mkinsner:
I've heard that people are having trouble searching for text within the SYCL 1.2.1 spec PDF.  Some of the issues appear to be on Mac specifically, but others are cross-platform.   The problem may be related to underscores - that's where there are known reproductions.

**Problem 1:** Searching for italicized text with an underscore misses matches, even on a PDF reader under Windows.  For example, SYCL1.2.1r5, p.76, the table has _cl_command_queue_info_ in the right column.  I can't make my PDF reader find any portion of the text, once I add an underscore to it.

**Problem 2:** Non-italicized hits missing on some Mac readers (Safari/Preview).  A reproducer of the problem is async_handler.  Searching with Safari/Preview on Mac apparently finds 8 results, while Acrobat reader on Mac finds 33 (which is consistent with what I see with a different reader in Windows).

oddhack:
We dealt with this in the GL specs as well. I have a vague recollection that the {underscore} package helped, but it's been a long time and I'm not certain.

keryell:
I am working on it... Several issues...

keryell:
Making some progress... "async_handler" now in Ubuntu Okular or Windows Acrobat reader is found 42 times instead of 33, but only 6 times by Mac Preview...
"cl_command_queue_info" is now found by the 3 tools.
Perhaps PDF support is deprecated on Mac and there is a new format to use. That would not surprise me as that sounds familiar...
Still diving...

keryell:
The current situation seems quite better.
Now there are *43* instances of `async_handler` found.

On the crazy part on Mac OS X:
- MacOS X Preview 10.1 (944.6.16.1) find the 43 instances, but when I change the PDF in its back, it just crashes;
- Safari 12.1.1 (14607.2.6.1.1) can somehow find all the 43 occurrences, but actually 33 are found
while searching forward while 10 are found while searching backward... :-( I find this... awkward :-) I would like to program this bug, I would not know how to do it... :-(

If there are people using Mac OS X who want to open some bug reports, go for it.
I prefer using Linux in the meantime... :-)

PR #4, opened 2019-06-11
Fix a lot of keyword-related issues with some PDF search impact

keryell:
This is a try to fix https://github.com/KhronosGroup/SYCL-Docs/issues/3

Replace most of the `\keyword` by `\codeinline` or `\gls`-like glossaries

This fix some wrong typing in the sense of computer programming.
It fixes also a lot of typesetting problems, such as PDF search due to
wrong "_" rendering.
The remaining \keyword is some technical debt about creating new
glossary entries in the specification.

Do not use `\textbf` for keywords.

The "_" are replaced by space in the PDF search otherwise...
Use `\codeinline` as expected instead.

Remove useless `\_` and fix some typesetting

A lot of escaping of `_` were actually useless and might cause some
trouble in the PDF output.

Use correct typesetting of `async_handler` in the glossary.

This makes the "_" to be searchable.

Now there are 43 instances of `async_handler` found.

Improve `async_handler` glossary encoding for MacOS X Preview search.

On the crazy part on Mac OS X:
- MacOS X Preview 10.1 (944.6.16.1) finds the 43 instances, but when I change the PDF in its back, it just crashes;
- Safari 12.1.1 (14607.2.6.1.1) can somehow find all the 43 occurrences, but actually 33 are found while searching forward while 10 are found while searching backward... :-( I find this... awkward :-) I would like to program this bug, I would not know how to do it... :-(

If there are people using Mac OS X who want to open some bug reports, go for it.

### DIFF
--- a/latex/accessors.tex
+++ b/latex/accessors.tex
@@ -29,9 +29,9 @@ access to the data within an execution schedule.
 
 Because of this the interface of the \codeinline{accessor} will
 be different depending on the possible combinations of those parameters. There
-are three main categories of accessor; buffer accesors (see Section
-~\ref{sub.section.accessors.buffer}), local accessors (see Section
-~\ref{sub.section.accessors.local}) and image accessors (see Section
+are three main categories of accessor; buffer accesors (see Section%
+~\ref{sub.section.accessors.buffer}), local accessors (see Section%
+~\ref{sub.section.accessors.local}) and image accessors (see Section%
 ~\ref{sub.section.accessors.image}).
 
 %*******************************************************************************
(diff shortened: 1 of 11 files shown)


latex/accessors.tex
```diff
@@ -29,9 +29,9 @@
 
 Because of this the interface of the \codeinline{accessor} will
 be different depending on the possible combinations of those parameters. There
-are three main categories of accessor; buffer accesors (see Section
-~\ref{sub.section.accessors.buffer}), local accessors (see Section
-~\ref{sub.section.accessors.local}) and image accessors (see Section
+are three main categories of accessor; buffer accesors (see Section%
```
mkinsner: Are these % intentional?  I can't figure out what the impact is.

keryell: Yes. It was a bug in the specification I fixed while I was reading it.
There was a potential double space because of the new-line (which is a space) + "~" which is already a non-breakable space...
So adding a % at the end 
Classical LaTeX trick, explaining why there are often a lot of % at some end of line when spaces matter.

PR #13, opened 2019-07-09
Add thread_local keyword to storage class restriction so that can be found with search

mkinsner:


latex/compiler_abi.tex
```diff
@@ -113,8 +113,8 @@ \section{Language restrictions for kernels}
     Recursion is not allowed in a SYCL kernel or any
     code called from the kernel.
   \item
-    Variables with thread storage duration are not allowed to be odr-used in
-    kernel code
+    Variables with thread storage duration (\lstinline{thread_local} storage class specifier)
```
keryell: `\codeinline` instead?
But I realize that there is some new `\lstinline` below already...

mkinsner: I used the local style.  Do you want me to change the set of `\lstinline` in this section to `\codeinline`?  I can easily do that if we prefer `\codeinline`.

keryell: Yes, since we have
```
\newcommand{\codeinline}[1]{\lstinline[basicstyle=\ttfamily\small]$#1$}
```
currently `\codeinline` should typeset the text in a smaller size.

So, if you can fix `\lstinline{noexcept}`, `\lstinline{const}` and `\lstinline{constexpr}` while you are working on it, it would be cleaner.

mkinsner: Pushed additional commit that changes to codeinline.  The PDF render looks correct.

Ruyk:
Looks good to us

PR #15, opened 2019-07-11
Fix bug where host_image accessors can't have a read_write access mode

mkinsner:
Makes SYCL 1.2.1 consistent with OpenCL 1.2, where read_write mode is allowed for accesses mapped on the host.

Resolves internal Khronos issue https://gitlab.khronos.org/sycl/Specification/issues/238


jbrodman:
@keryell  - review ping.

Issue #9, opened 2019-06-14
SLM use


kevin-harms:
https://github.com/KhronosGroup/SYCL-Docs/blob/4a2dff3a779ae6fa4a477b0f02002c6ab82a3088/latex/accessors.tex#L632

The shared local memory accessor is quite vague specifying what it actually does. I can understand this is to be flexible for vendor implementation but makes it difficult to use in practice. It is not clear if the accessor is allocating a block of memory that the user should split between workgroups or if the total memory allocated is per workgroup.

In the example code, does the runtime allocate 1K * 16 workgroups of SLM or just 1K of SLM and it will be distributed equally between the 16 workgroups?  I don't believe the spec is clear in this regard.

```
    queue.submit([&](cl::sycl::handler &cgh)
    {
        auto acc = cl::sycl::accessor<int, 1,  cl::sycl::access::mode::read_write, 
                           cl::sycl::access::target::local>(cl::sycl::range<1>(256), cgh);
        cgh.parallel_for_work_group(cl::sycl::range<1>(16), cl::sycl::range<1>(64),
            [=](cl::sycl::group<1> g)
            {
                g.parallel_for_work_item([&](cl::sycl::h_item<1> hi)
                    {
                        size_t i = hi.get_local_id(0);
                        acc[i] = hi.get_global_id(0); 
                    }
                );
            }
        );
    }
);
```

keryell:
Resolved by https://github.com/KhronosGroup/SYCL-Docs/pull/18

PR #18, opened 2019-07-18
Clarify local mem allocation size, and fix local accessor member definition bugs

mkinsner:
Proposal to address #9.  Also fixes what I believe are a number of bugs in the local accessor member descriptions (references to buffer instead of SLM)


fraggamuffin:
Aug 12:
*  may need to adjust review process

fraggamuffin:
Aug 12:
*  filed by Kevin Harms

keryell:
Merging as Ruyman is back!

PR #19, opened 2019-07-18
WIP: Should the accessor listings for local accessor, for example, be a partial specialization?

mkinsner:
If this makes sense to you Ronan, I can make the change for the variety of specializations.

### DIFF
--- a/latex/accessors.tex
+++ b/latex/accessors.tex
@@ -353,7 +353,7 @@ Tables~\ref{table.specialmembers.common.reference} and
   specialization}
 {table.constructors.accessor.buffer}
   \addRow
-    { accessor(buffer<dataT, 1> \&bufferRef) }
+    { accessor(buffer<dataT, 1, AllocatorT> \&bufferRef) }
     {
       Available only when: \codeinline{((isPlaceholder ==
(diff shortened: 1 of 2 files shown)


fraggamuffin:
Aug 12:
*  seems like a copy paste error, so need to be confirmed to see if there was actual reason behind this.

AerialMantis:
@keryell local accessors are not permitted to be placeholders, this is mentioned in Table 4.47, we could perhaps include this in the specialisation to make that clearer.

keryell:
> @keryell local accessors are not permitted to be placeholders, this is mentioned in Table 4.47, we could perhaps include this in the specialisation to make that clearer.

Yes we can specialize it with `access::placeholder::false_t`

keryell:
Still a WIP about specializing it with `access::placeholder::false_t`

keryell:
still waiting for https://github.com/KhronosGroup/SYCL-Docs/pull/17 to move first

keryell:
#### 2021/09/02 SYCL Khronos committee meeting
In SYCL 2020 we deprecated some targets and created specific types instead.
Just close without merging.

PR #27, opened 2019-07-25
Clarify that image and image accessor get_size() returns the same res…

jbrodman:
…ult.

Signed-off-by: James Brodman <james.brodman@intel.com>

### DIFF
--- a/latex/accessors.tex
+++ b/latex/accessors.tex
@@ -1019,8 +1019,10 @@ Tables~\ref{table.specialmembers.common.reference} and
   \addRow
     { size_t get_size() const }
     {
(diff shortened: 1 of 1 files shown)


jbrodman:
Actually.  I'm not sure this is the right solution.  

get_size() doesn't map to anything in OpenCL on the kernel side.

Instead, I think get_size() should be replaced with get_range() that returns the # of pixels in each dimension.  This lines up with the OCL builtin image queries.

get_range() should return the same as image::get_range().

fraggamuffin:
Aug 12:
*  discussed internally, waiting for MR from CP.

keryell:
Closed as a duplicate of https://github.com/KhronosGroup/SYCL-Docs/pull/31

PR #28, opened 2019-07-26
Host kernel return type restriction into the restrictions section

mkinsner:
The kernel return type requirement (must be void) is written in text form in the middle of a spec paragraph, but isn't easy to find.  This change adds it also to the list of kernel function restrictions, and mentions what to avoid to not violate the restriction.

### DIFF
--- a/latex/compiler_abi.tex
+++ b/latex/compiler_abi.tex
@@ -96,9 +96,9 @@ The following restrictions are applied to device functions and kernels:
     allocate storage are permitted.
   \item
     Kernel functions must always have a \codeinline{void} return type.  A kernel lambda
-    trailing-return-type that is not void is therefore illegal, as is a return statement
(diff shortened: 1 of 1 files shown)


fraggamuffin:
Aug 12: 
*  please review and upvote

latex/compiler_abi.tex
```diff
@@ -94,6 +94,11 @@ \section{Language restrictions for kernels}
     allocate storage are disallowed in a SYCL kernel. The placement
     \codeinline{new} operator and any user-defined overloads that do not
     allocate storage are permitted.
+  \item
+    Kernel functions must always have a \codeinline{void} return type.  A kernel lambda
+    trailing-return-type that is not void is therefore illegal, as is a return statement
+    (that would return from the kernel function) with an expression that does not convert
+    to void.
```
keryell: To be coherent, all the `void` should appear in a `\codeinline{void}`

PR #33, opened 2019-08-15
Remove OpenCL interop for stream class

AerialMantis:
This pull request removes OpenCL interop support from the `stream` class as there is no suitable OpenCL C equivelant and enforcing one would cause problems for printf-based implementations of the `stream` class.

### DIFF
--- a/latex/programming_interface.tex
+++ b/latex/programming_interface.tex
@@ -90,7 +90,7 @@ For more details regarding these facilities and considerations for their use see
 \subsection{Common reference semantics}
 \label{sec:reference-semantics}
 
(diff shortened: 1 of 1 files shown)


keryell:
@AerialMantis update the CTS too.

Look again at it after the CTS change.

bader:
https://github.com/KhronosGroup/SYCL-CTS/pull/9 should align CTS with the spec.

bader:
https://github.com/KhronosGroup/SYCL-CTS/pull/9 is merged.
@keryell, @Ruyk, @mkinsner, I think we can merge this PR now.

Ruyk:
Yes, this should be merged.

keryell:
Thanks for the great collaboration!

PR #38, opened 2019-08-23
Rule of Zero for Common by-value semantics classes (Gitlab issue 260)

nliber:
In section 4.3.3, we spell out that the five special member functions (copy/move constructor/assignment operator and destructor) are =default.  This violates the C++ Core Guideline C.20: If you can avoid defining default operations, do. We should use the Rule of Zero (and possibly recommend using the Rule of Five if any of them have to be explicitly declared).

### DIFF
--- a/latex/headers/common-byval.h
+++ b/latex/headers/common-byval.h
@@ -5,15 +5,17 @@ class T {
   ...
 
  public:
-  T(const T &rhs) = default;
+  // Rule of Zero: the following five special member functions should not be explicitly declared
 
-  T(T &&rhs) = default;
+  // T(const T &rhs) = default;
 
-  T &operator=(const T &rhs) = default;
+  // T(T &&rhs) = default;
 
-  T &operator=(T &&rhs) = default;
+  // T &operator=(const T &rhs) = default;
 
-  ~T() = default;
+  // T &operator=(T &&rhs) = default;
+
+  // ~T() = default;
 
   bool operator==(const T &rhs) const;
 
--- a/latex/headers/common-byval.h
+++ b/latex/headers/common-byval.h
@@ -20,13 +20,9 @@ class T {
 
  public:
   // If any of the following five special member functions are not
-  
   // public, inline or defaulted, then all five of them should be
-    
   // explicitly declared (see rule of five).
-    
   // Otherwise, none of them should be explicitly declared
-    
   // (see rule of zero).
 
   // T(const T &rhs);
--- a/latex/headers/common-byval.h
+++ b/latex/headers/common-byval.h
@@ -19,17 +19,25 @@ class T {
   ...
 
  public:
-  // Rule of Zero: the following five special member functions should not be explicitly declared
+  // If any of the following five special member functions are not
+  
+  // public, inline or defaulted, then all five of them should be
+    
+  // explicitly declared (see rule of five).
+    
+  // Otherwise, none of them should be explicitly declared
+    
+  // (see rule of zero).
 
-  // T(const T &rhs) = default;
+  // T(const T &rhs);
 
-  // T(T &&rhs) = default;
+  // T(T &&rhs);
 
-  // T &operator=(const T &rhs) = default;
+  // T &operator=(const T &rhs);
 
-  // T &operator=(T &&rhs) = default;
+  // T &operator=(T &&rhs);
 
-  // ~T() = default;
+  // ~T();
 
   ...
 
--- a/latex/programming_interface.tex
+++ b/latex/programming_interface.tex
@@ -187,7 +187,7 @@ Each of the following \gls{sycl-runtime} classes: \codeinline{id}, \codeinline{r
 
 Some \gls{sycl-runtime} classes will have additional behavior associated with copy, movement, assignment or destruction semantics. If these are specified they are in addition to those specified above unless stated otherwise.
 
-Each of the runtime classes mentioned above must provide a common interface of special member functions and member functions in order to fulfil the copy, move, destruction and equality requirements.  If all of the special member functions are inlined, public and defaulted, none of them should be explicitly declared (Rule of Zero); otherwise, all of them should be explicitly declared (Rule of Five).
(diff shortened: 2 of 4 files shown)


keryell:
@nliber can you sign the CLA so we can move on?

keryell:
I'd like to have a C++ lawyer from Codeplay on this. I put @mmha as a random guess, but @Ruyk feel free to nominate someone else. :-)

latex/programming_interface.tex
```diff
@@ -179,7 +179,8 @@ \subsection{Common by-value semantics}
 
 Some \gls{sycl-runtime} classes will have additional behavior associated with copy, movement, assignment or destruction semantics. If these are specified they are in addition to those specified above unless stated otherwise.
 
-Each of the runtime classes mentioned above must provide a common interface of special member functions and member functions in order to fulfil the copy, move, destruction and equality requirements.
+Each of the runtime classes mentioned above must provide a common interface of special member functions and member functions in order to fulfil the copy, move, destruction and equality requirements.  If all of the special member functions are inlined, public and defaulted, none of them should be explicitly declared (Rule of Zero); otherwise, all of them should be explicitly declared (Rule of Five).
```
Ruyk: I think we need a link to some reference for rule of Zero and rule of Five, not all SYCL readers are necessarily well versed in C++

Ruyk:
This look good to me, but I'd like @AerialMantis to do a check since he wrote the previous wording

latex/programming_interface.tex
AerialMantis: I agree we should define what these terms mean and the justification for them to avoid confusion. It is defined in the C++ core guidelines, but I don't believe they have a fixed revision we can reference, the ISO C++ standard is also hard to reference on this rule as the wording is scattered across a number of sections.

Perhaps we should have a glossary term that describes the rule of zero and the rule of five in more detail, then anywhere we mention it we can just use the glossary term?

latex/programming_interface.tex
```diff
@@ -189,29 +190,29 @@ \subsection{Common by-value semantics}
 \addFootNotes{Common special member functions for by-value semantics}
 {table.specialmembers.common.byval}
   \addRow
-    {T(const T \&rhs) = default}
+    {T(const T \&rhs);}
     {
-      Default copy constructor.  
+      Copy constructor.
```
AerialMantis: Perhaps we should also state here that these member functions can be implicit or explicitly defined as defaulted.

keryell:
This looks like a good test bench for ISO C++ SG20 which cites the rule of 0, rule of five, etc. @cjdb help! :-)

keryell:
In the meantime I attended ISO C++ SG20 in Belfast about teaching C++ and again it looks like this is really a question for @cjdb... Help! :-)

bader:
@nliber, could you make sure that comments from @keryell, @Ruyk and @AerialMantis are resolved?

AerialMantis:
Excerps from the C++ core guidelines that we could potentially lin to (though we should be wary that these section numbers are not guaranteed to be consistent):
* Rule of five: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c21-if-you-define-or-delete-any-default-operation-define-or-delete-them-all
* Rule of zero: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c20-if-you-can-avoid-defining-default-operations-do

latex/programming_interface.tex
nliber: I'm moving the definitions of "rule of five" and "rule of zero" to the glossary.

nliber: I don't think we should provide a link to the core guidelines, as this spec is for SYCL implementers.  How users write their code, other than meeting requirements, is beyond what this specification covers.

latex/headers/common-byval.h
```diff
@@ -5,15 +5,17 @@ class T {
   ...
 
  public:
-  T(const T &rhs) = default;
+  // Rule of Zero: the following five special member functions should not be explicitly declared
```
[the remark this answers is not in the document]
nliber: If they are to be uncommented, then we are not following the rule of zero, because it is never correct to have all five of those special member functions inlined, public and defaulted.

latex/programming_interface.tex
nliber: If we are following the rule of zero, it is incorrect for all five of these special member functions to be explicitly defined as defaulted and inlined.

keryell: We had already a similar discussion in the past about whether the SYCL spec should describe the real implementation or just some equivalent behaviour (such as "there are some constructors, synthesized or real, that behaves like if they where defaulted...").
For example since triSYCL relies aggressively on very modern C++, some constructors are not implemented at all, just synthesized and thus probably departs from the spec because of this...

latex/headers/common-byval.h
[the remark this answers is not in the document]
AerialMantis: SYCL working group: We are happy to leave the declarations commented out, though we should update the comment slightly to reflect both the rule of zero and the rule of five, i.e. they either have to all be declared or none should be declared.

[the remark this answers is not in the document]
nliber: The comments have been updated.

latex/programming_interface.tex
nliber: A reference to the rule of five and rule of zero has been added to the table.

latex/headers/common-byval.h
```diff
@@ -19,17 +19,25 @@ class T {
...
  public:
-  // Rule of Zero: the following five special member functions should not be explicitly declared
+  // If any of the following five special member functions are not
+  
+  // public, inline or defaulted, then all five of them should be
+    
+  // explicitly declared (see rule of five).
+    
+  // Otherwise, none of them should be explicitly declared
+
```
keryell: Why these 4 empty lines above in between?

nliber: There was a blank line between every line in the original.  I was just trying to be consistent, but it wasn't obvious what to do with multiline comments.  Anyway, I removed the blank lines.

PR #42, opened 2019-09-10
Issue259 hiddenfriends

nliber:
There is a definition of "hidden friend functions" in section 4.3.2.  I wasn't sure where else to put it.  

The definition is:

> A hidden friend function is a function first declared via a `friend` declaration with no additional out of class or namespace scope declarations. Hidden friend functions are only visible to ADL (Argument Dependent Lookup) and are hidden from qualified and unqualified lookup. Hidden friend functions have the benefits of avoiding accidental implicit conversions and faster compilation.

In the class synopses, I declared hidden friends in the form

```
friend ret operatorOP(const type& lhs, const type& rhs) { /* ... */ }
```

to indicate that the body of the function must be present (even if all that body does is call another function).  This is a requirement for hidden friends to work.

Anthony Williams does a great job explaining [The Power of Hidden Friends in C++](https://www.justsoftwaresolutions.co.uk/cplusplus/hidden-friends.html), including using them for unary operators.

With the exception of those operators that cannot be non-member functions, I changed all the operators to use hidden friends.  This affected reference semantics, by-value semantics, `multi_ptr`, `range`, `id` and `vec`.

### DIFF
--- a/latex/vec_class.tex
+++ b/latex/vec_class.tex
@@ -208,6 +208,42 @@ The constructors, member functions and non-member functions of the SYCL \codeinl
   {
     Stores the components of this SYCL \codeinline{vec} into the values at the address of \codeinline{ptr} offset in elements of type \codeinline{dataT} by \codeinline{numElements * offset}.
   }
+
+  \addRowTwoL
+    {vec<dataT, numElements> \&operator=(}
+    {  const vec<dataT, numElements> \&rhs)}
+    {
+      Assign each element of the \codeinline{rhs} SYCL \codeinline{vec} to each element of this SYCL \codeinline{vec} and return a reference to this SYCL \codeinline{vec}.
+    }
+  \addRowTwoL
+    {vec<dataT, numElements> \&operator=(}
+    {  const dataT \&rhs)}
+    {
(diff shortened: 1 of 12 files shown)


latex/vec_class.tex
```diff
@@ -208,216 +208,205 @@ \subsubsection{Vec interface}
   {
     Stores the components of this SYCL \codeinline{vec} into the values at the address of \codeinline{ptr} offset in elements of type \codeinline{dataT} by \codeinline{numElements * offset}.
   }
+
+  \addRowTwoL
```
keryell: Is this change relevant to the hidden friends too?

keryell: Ah OK.

AerialMantis:
SYCL working group:
@AerialMantis to take a look at this and merge it once it is approved.

mkinsner:
@nliber can you please resolve the merge conflict?  Otherwise this is ready to go.

nliber:
I know it says there is a merge conflict, but I'm not seeing any...

bader:
I suggest merging latest version of the master branch into you private branch.

nliber:
Hopefully now...

PR #46, opened 2019-09-27
Clarify sycl::stream parameters and behavior

mkinsner:


mkinsner:
I believe that all discussion is resolved with the current PR wording.

PR #50, opened 2019-10-16
Fix bug in group::get_group_range()

Pennycook:
get_group_range() should return number of work-groups, not their size.

This commit also moves the note that a group's local range can be provided by the programmer or chosen by the run-time to the description of get_local_range().

Resolves internal Khronos issue: https://gitlab.khronos.org/sycl/Specification/issues/270


mkinsner:
LGTM

PR #55, opened 2019-10-29
Fix operations on id & range

keryell:
Add <, >, <=, >= for id and range.
Add some missing || and &&.

This solve the non controversial part of internal discussion https://gitlab.khronos.org/sycl/Specification/issues/148 and https://gitlab.khronos.org/sycl/Specification/merge_requests/326

### DIFF
--- a/latex/expressingParallelism.tex
+++ b/latex/expressingParallelism.tex
@@ -119,6 +119,7 @@ A synopsis of the SYCL \codeinline{range} class is provided below. The construct
       Where \codeinline{OP} is: \codeinline{+}, \codeinline{-}, \codeinline{*},
       \codeinline{/}, \codeinline{\%}, \codeinline{<<}, \codeinline{>>},
       \codeinline{\&}, \codeinline{|}, \codeinline{^},
+      \codeinline{\&\&}, \codeinline{||},
       \codeinline{<}, \codeinline{>}, \codeinline{<=},
       \codeinline{>=}.
       \newline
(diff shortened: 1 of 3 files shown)


illuhad:
Can you elaborate what the issue is for `&&` and `||`? Why do they have to go?

latex/expressingParallelism.tex
```diff
@@ -118,8 +118,8 @@ \subsubsection{\texttt{range} class}
     {
       Where \codeinline{OP} is: \codeinline{+}, \codeinline{-}, \codeinline{*},
       \codeinline{/}, \codeinline{\%}, \codeinline{<<}, \codeinline{>>},
-      \codeinline{\&}, \codeinline{|}, \codeinline{^}, \codeinline{\&\&},
```
AerialMantis: Why does this change remove the `&&` and `||` operators?

keryell: Because there were half-baked in the current specification and it is not clear if we really want the short-cut semantics in a SIMT context...
What you would imagine otherwise? Just an implicit conversion to bool of all the arguments and then just use `&` or `|` accordingly?
Are there users or some CTS testing this today?

AerialMantis: The current behaviour is that you perform the operator element-wise and if the results are booleans they are converted to size_t.

```
Constructs and returns a new instance of the
SYCL id class template with the same dimensionality as this SYCL id, where each
element of the new SYCL id instance is the
result of an element-wise OP operator between each element of this SYCL id and
each element of the rhs id. If the operator
returns a bool the result is the cast to size_t.
```

This behaviour is tested in the CTS here - https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/tests/common/common.h#L557

I agree this should be improved in a later revision, but I'm wary fo removing features from the specification in an errata change.

nliber: If we decide not to remove it at this time, can we at least deprecate it?

keryell: Following the discussion I have added && and || in some missing places instead of removing the existing ones.

PR #56, opened 2019-10-29
Clarify mapping of unit stride dimension between SYCL and OpenCL

mkinsner:
Clarify mapping of unit stride dimension (and dimensions more generally) between SYCL and OpenCL.  Ensures stable mapping between SYCL and OpenCL interop enqueues.

Also change equations to be zero-based instead of one-based.

### DIFF
--- a/latex/expressingParallelism.tex
+++ b/latex/expressingParallelism.tex
@@ -2257,7 +2257,11 @@ user code should call \codeinline{clReleaseKernel} if the
 
 OpenCL and SYCL use opposite conventions for the unit stride dimension.  SYCL
 aligns with C++ conventions, which is important to understand from a performance
-perspective when porting code to SYCL.
+perspective when porting code to SYCL.  The unit stride dimension, at least for data, is implicit in the
+linearization equations in SYCL (Equation~\ref{row-major-equation-buffer}) and OpenCL.  SYCL aligns with C++ array subscript ordering
+\codeinline{arr[a][b][c]}, in that range constructor dimension ordering used to launch a kernel
+(e.g. \codeinline{range<3> R\{a,b,c\}}) and range and ID queries within a kernel, are ordered in
+the same way as the C++ multi-dimensional subscript operators (unit stride on the right).
 
 When specifying a \codeinline{range} as the global or local size
 in a \codeinline{parallel_for} that invokes an OpenCL interop kernel (through
@@ -2267,7 +2271,7 @@ the highest dimension of the range in SYCL will map to the
 lowest dimension within the OpenCL kernel.  That statement applies to both
 an underlying enqueue operation such as \codeinline{clEnqueueNDRangeKernel}
 in OpenCL, and also ID and size queries within the OpenCL kernel.
-For example, a global range specified in SYCL as:
+For example, a 3D global range specified in SYCL as:
 
 \begin{lstlisting}[style=nonumbers]
 range<3> R{r0,r1,r2};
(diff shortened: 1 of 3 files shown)


latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
 \codeinline{cl_kernel} is no longer needed in the calling context.
 
 
+%*******************************************************************************
+% OpenCL Kernel Conventions and SYCL
+%*******************************************************************************
+\subsection{OpenCL Kernel Conventions and SYCL}
+\label{sec:opencl_kern_conventions_and_sycl}
+
+OpenCL and SYCL use opposite conventions for the unit stride dimension.  SYCL
```
AerialMantis: I think we should outline here what the C++ and OpenCL conventions are to set the foundation for the rest of this section.

AerialMantis: I also think we should clarify that the OpenCL convention here is not madated by OpenCL, but is inferred and most commonly adopted by implementations.

latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
...
+Likewise, for a two dimensional global range \codeinline{range<2> R\{r0,r1\}}, OpenCL and SYCL kernel code queries relate to the range as:
+
+\startGenericThreeColTable{8cm}{3cm}{3cm}{SYCL kernel query}{OpenCL kernel query}{Returned Value}
+\genericThreeColRow {nd_item::get_global_range(0) / item::get_range(0)}{get_global_size(1)}{\codeinline{r0}}
+\genericThreeColRow {nd_item::get_global_range(1) / item::get_range(1)}{get_global_size(0)}{\codeinline{r1}}
+\genericThreeColRow {nd_item::get_global_id(0) / item::get_id(0)}{get_global_id(1)}{Value in range 0..(\codeinline{r0}-1)}
+\genericThreeColRow {nd_item::get_global_id(1) / item::get_id(1)}{get_global_id(0)}{Value in range 0..(\codeinline{r1}-1)}
+\completeGenericTabular
+\captionGenericTable{Example range mapping from SYCL enqueued two dimensional global \codeinline{range} to OpenCL and SYCL queries}
+\completeGenericTable
```
Ruyk: Don't particularly mind, but seems a bit verbose to have two separate tables - maybe collapse the two tables into one will make reading easier ?

latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
...
+\end{lstlisting}
+
+maps to an \codeinline{clEnqueueNDRangeKernel} \codeinline{global_work_size} argument
+of:
+
+\begin{lstlisting}[style=nonumbers]
+size_t cl_interop_range[3] = {r2,r1,r0};
+\end{lstlisting}
+
+Work-item and work-group ID and range queries have the same reversed convention for unit
```
AerialMantis: The way latex generates the following tables means the two text sections appear first and then the two tables show together, so it reads a bit strange. If we use links to the tables instead I think this will avoid any confusion.

latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
...
+aligns with C++ conventions, which is important to understand from a performance
+perspective when porting code to SYCL.
+
+When specifying a \codeinline{range} as the global or local size
+in a \codeinline{parallel_for} that invokes an OpenCL interop kernel (through
+\codeinline{cl_kernel} interop or \codeinline{compile_with_source}/
+\codeinline{build_with_source}),
+the highest dimension of the range in SYCL will map to the
+lowest dimension within the OpenCL kernel.  That statement applies to both
+an underlying enqueue operation such as \codeinline{clEnqueueNDRangeKernel}
```
Ruyk: Does apply to all enqueue operations, e.g. rectangular copies, or only to NDRangeKernel?

latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
...
+in OpenCL, and also ID and size queries within the OpenCL kernel.
+For example, a global range specified in SYCL as:
```
AerialMantis: I think we should specify the 2-dimensional case here as well to clarify that only two dimensions are used otherwise there can be some ambiguity in how they are passed to `clEnqueueNDRangeKernel` after the flip.

I.e. `{r0, r1}` -> `{r1, r0}`, rather than `{1, r1, r0}`.

latex/expressingParallelism.tex
```diff
@@ -2249,6 +2249,66 @@ \subsubsection{Defining kernels using OpenCL C kernel objects}
...
+% OpenCL Kernel Conventions and SYCL
+%*******************************************************************************
+\subsection{OpenCL Kernel Conventions and SYCL}
+\label{sec:opencl_kern_conventions_and_sycl}
+
+OpenCL and SYCL use opposite conventions for the unit stride dimension.  SYCL
+aligns with C++ conventions, which is important to understand from a performance
+perspective when porting code to SYCL.
+
+When specifying a \codeinline{range} as the global or local size
```
AerialMantis: This wording suggests that this only applies to the interop `parallel_for`, so the non-interop `parallel_for` API is free to choose how to map the unit stride, but the table further down appears to specify a mapping for non-interop SYCL kernels.

AerialMantis:
Very good clarification, thanks @mkinsner just some minor comments.

latex/expressingParallelism.tex
mkinsner: I tried to outline the convention and clarify.  Please let me know if you think it should be improved in any way.

latex/expressingParallelism.tex
mkinsner: I added text before the table describing when the columns apply.  The mapping reversal only applies with OpenCL interop kernels, while the SYCL column is for reference (already well defined in the spec).

latex/expressingParallelism.tex
mkinsner: Good point.  I added this to address:

> The mapping of highest dimension in SYCL to lowest dimension in OpenCL applies to all
operations where a multi-dimensional construct must be mapped, such as when mapping SYCL
explicit memory operations to OpenCL APIs like \codeinline{clEnqueueCopyBufferRect}.

latex/expressingParallelism.tex
mkinsner: Good point.  I switched to a table reference instead.

latex/expressingParallelism.tex
mkinsner: Agreed.  I merged into one table, and also added the 1D case for completeness.

latex/expressingParallelism.tex
AerialMantis: This addition looks good, that covers what I had in mind, thanks.

latex/expressingParallelism.tex
AerialMantis: This is covered by the new addition, thanks.

latex/expressingParallelism.tex
AerialMantis: This is much clearer now, thanks.

keryell:
This is related to internal https://gitlab.khronos.org/sycl/Specification/issues/72

PR #59, opened 2019-11-08
Fix SYCL definition of preferred_work_group_size_multiple

mkinsner:
Fix SYCL definition of preferred_work_group_size_multiple so that "multiple" is in the definition.  That word was accidentally missed in the definition, relative to the query name and also the equivalent OpenCL definition.  Also state that it shouldn't be more than the max legal size.

### DIFF
--- a/latex/expressingParallelism.tex
+++ b/latex/expressingParallelism.tex
@@ -1868,8 +1868,10 @@ descriptor.}
 {info::kernel_work_group::preferred_work_group_size_multiple}
 {size_t}
 {
-  Returns the preferred work-group size for executing a kernel on a particular
-  device.
+  Returns a value, of which work-group size is preferred to be a multiple,
+  for executing a kernel on a particular device.  This is a performance
+  hint.  The value must be less than or equal to that returned by
+  \codeinline{info::kernel_work_group::work_group_size}.
 }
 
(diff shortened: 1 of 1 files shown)


latex/expressingParallelism.tex
```diff
@@ -1868,8 +1868,10 @@ \subsection{Kernel class}
 {info::kernel_work_group::preferred_work_group_size_multiple}
 {size_t}
 {
-  Returns the preferred work-group size for executing a kernel on a particular
-  device.
+  Returns a value, of which work-group size is preferred to be a multiple,
+  for executing a kernel on a particular device.  This is a performance
+  hint.  The value must be less than or equal to that returned by
+  \codeinline{info::kernel_work_group::work_group_size}.
```
Ruyk: Unrelated to your change, but seems to me 

> info::kernel_work_group::work_group_size

should be named 

>info::kernel_work_group::max_work_group_size 

instead?

mkinsner: I thought so too, but SYCL currently matches OpenCL.  The device has CL_DEVICE_MAX_WORK_GROUP_SIZE, while kernels have CL_KERNEL_WORK_GROUP_SIZE which is the maximum.

PR #61, opened 2019-11-11
Update contributor list

mkinsner:


MathiasMagnus:
Seems good to me. (Máté)

Naghasan:
Could you add "Tomas Matheson, Codeplay" does not seem to be in the list

mkinsner:
> Could you add "Tomas Matheson, Codeplay" ? does not seem to be in the list

Done.

PR #16, opened 2019-07-15
Clarify image formats when multiple OpenCL implementations, and on the host device

mkinsner:
Resolves internal Khronos issue: https://gitlab.khronos.org/sycl/Specification/issues/219

PR #63, opened 2019-11-25
Add clarification of supported image formats

AerialMantis:
Add clarification as to the required image support for the host device. This pull request replaces https://github.com/KhronosGroup/SYCL-Docs/pull/16

* Add general clarification on required image format for images.
* Add description of the minimum image formats.
* Add clarification on requirement for the host device support to
support the minimum image formats.

### DIFF
--- a/latex/architecture.tex
+++ b/latex/architecture.tex
@@ -1044,12 +1044,52 @@ through specialized operations.
 Elements of a buffer object can be a scalar data type (such as an int, float),
 vector data type, or a user-defined structure. In SYCL, a \gls{buffer} object is a
 templated type (\codeinline{cl::sycl::buffer}), parameterized by the element
-type and number of dimensions. An \gls{image} object is stored in one of a limited
-number of formats. The elements of an image object are selected from a list of
-predefined image formats which are provided by an underlying OpenCL
-implementation.  Images are encapsulated in the \codeinline{cl::sycl::image}
-type, which is templated by the number of dimensions in the image. The minimum
-number of elements in a memory object is one.
+type and number of dimensions.
+
+An \gls{image} is represented by the \codeinline{cl::sycl::image} class
+template, parametrized by the dimensionality. The minimum number of elements in
+a memory object is one. Elements of an \gls{image}, are stored in memory
+according one of a limited number of image formats described by a combination of
+\codeinline{cl::sycl::image_channel_order} and
+\codeinline{cl::sycl::image_channel_type}. An \gls{image} can be constructed
+with any one of these image formats, however whether that \gls{image} can be
+passed to a \gls{sycl-kernel-function} is subject to the target \gls{device}
+supporting that image format. The image formats supported by a \gls{device} is
+implementation defined, however if the \gls{device} is reported as supporting
+images it must support the minimum image formats as described below:
+
+\begin{itemize}
+  \item
+    \codeinline{cl::sycl::image_channel_order::rgba}
+    \begin{itemize}
+      \item
+        \codeinline{cl::sycl::image_channel_type::unorm_int8}
+      \item
+        \codeinline{cl::sycl::image_channel_type::unorm_int16}
+      \item
+        \codeinline{cl::sycl::image_channel_type::signed_int8}
+      \item
+        \codeinline{cl::sycl::image_channel_type::signed_int16}
+      \item
+        \codeinline{cl::sycl::image_channel_type::signed_int32}
+      \item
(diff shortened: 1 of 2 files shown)


latex/architecture.tex
```diff
@@ -1044,12 +1044,52 @@ \section{Memory objects}
...
-predefined image formats which are provided by an underlying OpenCL
-implementation.  Images are encapsulated in the \codeinline{cl::sycl::image}
-type, which is templated by the number of dimensions in the image. The minimum
-number of elements in a memory object is one.
+type and number of dimensions.
+
+An \gls{image} is represented by the \codeinline{cl::sycl::image} class
+template, parametrized by the dimensionality. The minimum number of elements in
+a memory object is one. Elements of an \gls{image}, are stored in memory
+according one of a limited number of image formats described by a combination of
```
ProGTX:
```suggestion
according to one of a limited number of image formats described by a combination of
```

latex/device_class.tex
```diff
@@ -250,12 +250,14 @@ \subsubsection{Device information descriptors}
...
-
    \addInfoRow
      {info::device::image_support}
      {bool}
-     {Returns true if images are supported by this SYCL \codeinline{device} and false if they are not. }
-
+     {
+       Returns true if images are supported by this SYCL \codeinline{device} and
+       false if they are not. If this SYCL \codeinline{device} is a
+       \gls{host-device} image must be supported.
```
ProGTX:
```suggestion
       \gls{host-device}, images must be supported.
```

keryell:
We need to discuss whether we close https://github.com/KhronosGroup/SYCL-Docs/pull/16 while still working on this.

latex/architecture.tex
```diff
@@ -1044,12 +1044,52 @@ \section{Memory objects}
...
-number of formats. The elements of an image object are selected from a list of
-predefined image formats which are provided by an underlying OpenCL
-implementation.  Images are encapsulated in the \codeinline{cl::sycl::image}
-type, which is templated by the number of dimensions in the image. The minimum
-number of elements in a memory object is one.
+type and number of dimensions.
+
+An \gls{image} is represented by the \codeinline{cl::sycl::image} class
+template, parametrized by the dimensionality. The minimum number of elements in
+a memory object is one. Elements of an \gls{image}, are stored in memory
```
keryell: Spurious ","

latex/device_class.tex
```diff
@@ -250,12 +250,14 @@ \subsubsection{Device information descriptors}
...
-
    \addInfoRow
      {info::device::image_support}
      {bool}
-     {Returns true if images are supported by this SYCL \codeinline{device} and false if they are not. }
-
+     {
+       Returns true if images are supported by this SYCL \codeinline{device} and
+       false if they are not. If this SYCL \codeinline{device} is a
+       \gls{host-device}, images must be supported.
```
keryell: I do not know if this has to be restated here since already stated on line 1343-1344 but if so, it should be clarified by adding for example "and thus this function returns true."

latex/architecture.tex
AerialMantis: Fixed.

latex/device_class.tex
AerialMantis: That's a good point, I've added this clarification.

latex/architecture.tex
```diff
@@ -1044,12 +1044,52 @@ \section{Memory objects}
...
+      \item
+        \codeinline{cl::sycl::image_channel_type::unsigned_int16}
+      \item
+        \codeinline{cl::sycl::image_channel_type::unsigned_int32}
+      \item
+        \codeinline{cl::sycl::image_channel_type::fp16}
+      \item
+        \codeinline{cl::sycl::image_channel_type::fp32}
+    \end{itemize}
+  \item
```
mkinsner: Where does this `bgra` minimum requirement come from?  I don't see it in OpenCL 1.2 - am I missing it?

mkinsner: I see that the full profile requires it but not embedded.

mkinsner: I think that we should conform to the embedded profile with this change (no `bgra`).  Either that, or we need to factor profile into the required formats list.

AerialMantis: SYCL working group: We have opted for clarifying the minimum requirements to match the OpenCL embedded profile, so removing `bgra` from the list.

keryell: @AerialMantis so I guess we are waiting for you to remove the `bgra` item.

keryell:
Merged during 2020/02/18 meeting

PR #64, opened 2019-12-19
Clarify introduction of atomic class

Pennycook:
The introduction to the `atomic` class in 4.11 makes two claims that are contradicted by other parts of the specification:
1. That atomic operations have no default `memory_order`
2. That `atomic` can be used in host code

### DIFF
--- a/latex/programming_interface.tex
+++ b/latex/programming_interface.tex
@@ -3205,8 +3205,7 @@ The SYCL specification provides atomic operations based on the C++11
 library syntax. The only available ordering, due to constraints of the
 OpenCL 1.2 memory model, is \codeinline{memory_order_relaxed}.
 The SYCL atomic library may map directly to
-the underlying C++11 library in host code, and must interact safely
-with the host C++11 atomic library when used in host code. The SYCL
+the underlying C++11 library on the host device. The SYCL
 library must be used in device code to ensure that only the limited
 subset of functionality is available. SYCL 1.2.1 device compilers should
 give a compilation error on use of the \codeinline{std::atomic}
--- a/latex/programming_interface.tex
+++ b/latex/programming_interface.tex
@@ -3203,9 +3203,8 @@ A \gls{work-group-barrier} or \gls{work-group-mem-fence} may provide ordering se
 
 The SYCL specification provides atomic operations based on the C++11
 library syntax. The only available ordering, due to constraints of the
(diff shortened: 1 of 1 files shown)


keryell:
It is not clear to me whether it is still possible to use the explicit SYCL atomics in kernels that executes on the host device.

Pennycook:
@keryell I think "device code" in that paragraph is supposed to refer to all devices (including the host device).  Would this text be clearer:

> The SYCL library must be used in kernel functions to ensure that only the limited subset of functionality is available.  SYCL 1.2.1 device compilers should give a compilation error on use of the `std::atomic` classes and functions in SYCL kernels, but the SYCL atomic library may map directly to the underlying C++11 library on the host device.

The key point is that `sycl::atomic` can't be used outside of SYCL kernels.

AerialMantis:
We agree the wording on these is not very clear so this should be clarified.

The change here makes sense to us.

The intention was that the SYCL `atomic` class should be available within SYCL kernel functions and on the host via accessors with `access::mode::atomic`. Though the atomicity is limited to the scope of a single accessor and cannot be used between multiple accessors.

Pennycook:
> The intention was that the SYCL atomic class should be available within SYCL kernel functions and on the host via accessors with access::mode::atomic. Though the atomicity is limited to the scope of a single accessor and cannot be used between multiple accessors.

Ah.  I hadn't thought enough about host accessors -- that makes this more complicated than I realized.  Maybe it should say: "may map directly to the underlying C++11 library in SYCL kernels executing on the host device or host code using access::mode::atomic accessors."

But an additional complication is that Section 3.5.2.2 says:
> The cl::sycl::atomic<T> must be used for elements of a buffer to provide safe atomic access to the buffer from device code.

If the `atomic` class should only be available within kernel functions and on the host via accessors, should the line above also be removed?

keryell:
> If the `atomic` class should only be available within kernel functions and on the host via accessors, should the line above also be removed?

Actually I was not thinking to host accessors but just to device code, including the host device.

When using host accessors, it is just normal C++ and you can use for example C++20 `atomic_ref`.
Note that for host device, the implementation of atomic on the host does not require compiler support and could be implemented like C++20 `atomic_ref`.

keryell:
I am good with the current wording.
We do not have to introduce atomic mode for the host accessor because it is just C++ there, so outside of the scope of SYCL.

PR #67, opened 2020-01-14
Omit range when reinterpreting 1D

### LINKED ISSUE #65: Buffer reinterpret 1D range
When reinterpreting to a buffer of dimension 1, the 1.2.1 spec allows only one way to specify the range, because of the following paragraph (Member functions for the buffer class):

```
Must throw an invalid_object_error SYCL exception if the total size in bytes represented by the type and range of the reinterpreted SYCL buffer does not equal the total size in bytes represented by the type and range of this SYCL buffer.
```

Example:
```cpp
buffer<int, 2> initial{range<2>{5, 6}};
buffer<float, 1> linear = initial.reinterpret<float, 1>(range<1>{initial.get_size() / sizeof(float)});
```

In this case it should be possible to just omit the range, because the SYCL runtime knows exactly what to do. If we also allow the dimensions parameter to be defaulted to the same one as the original buffer, it would simplify 1D buffers even further:
```cpp
buffer<int> initial{range<1>{16}};
buffer<float> linear_float = initial.reinterpret<float>();
buffer<char> linear_char = initial.reinterpret<char>();
```

ProGTX:
* Uses same template parameters
* Only allowed when target dimension 1

Closes #65

### DIFF
--- a/latex/headers/buffer.h
+++ b/latex/headers/buffer.h
@@ -126,6 +126,12 @@ class buffer {
   buffer<ReinterpretT, ReinterpretDim, AllocatorT>
   reinterpret(range<ReinterpretDim> reinterpretRange) const;
 
+  // Only available when ReinterpretDim == 1
+  // or when (ReinterpretDim == dimensions) &&
+  //         (sizeof(ReinterpretT) == sizeof(T))
+  template <typename ReinterpretT, int ReinterpretDim = dimensions>
+  buffer<ReinterpretT, ReinterpretDim, AllocatorT>
+  reinterpret() const;
 };
 }  // namespace sycl
 }  // namespace cl
--- a/latex/programming_interface.tex
+++ b/latex/programming_interface.tex
@@ -1421,6 +1421,28 @@ The SYCL \codeinline{buffer} class template takes a template parameter \codeinli
        offset or size of the sub-buffer view (in bytes) relative to the parent
        \codeinline{buffer}.
      }
+   \addRowThreeL
+     { template <typename ReinterpretT, int ReinterpretDim = dimensions> }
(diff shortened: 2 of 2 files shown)


ProGTX:
I think that makes sense, possibly even for all cases where the underlying types are the same size, e.g. `int` to `float`. Should we allow `const int` -> `int`?

keryell:
Well, actually `const` was a bad example since it is still something to be clarified in the specification if we allow it or not. https://gitlab.khronos.org/sycl/Specification/issues/226

But yes for `int` and `unsigned int` and so on and more generally when you have the same size.

ProGTX:
Updated MR:

Omit range when reinterpreting buffer
* Uses same template parameters
    * Default to same dimensions
* Only allowed when at least one of:
    * Target dimensions equals 1
    * Sizes of types are the same

keryell:
still waiting for CTS

ProGTX:
I've slightly updated the wording:

Omit range when reinterpreting buffer
* Uses same template parameters
    * Default to same dimensions
* Only allowed when at least one of:
    * Target dimensions equals 1
    * _Dimensions are the same_ and
       sizes of types are the same

I've also created a MR for the CTS: https://github.com/KhronosGroup/SYCL-CTS/pull/47

keryell:
@bader are you OK with this?

fraggamuffin:
Mar 24:

- is this adding to existing feature, aka an improvement to existing api
- while it is a new overload 
- it is changing the api
- keep it open and cherry pick it in 2020

keryell:
In flight for future SYCL version with https://gitlab.khronos.org/sycl/Specification/merge_requests/416

PR #69, opened 2020-01-21
Clarify accessor mode union, and requirements for accessors without uses

mkinsner:


latex/architecture.tex
```diff
@@ -475,11 +475,12 @@ \subsection{SYCL Application Memory Model}
...
 
-To access memory objects from inside a kernel, the user must create an
+To access memory objects, the user must create an
 \gls{accessor} object which parameterizes the type of access
-the kernel requires. 
+that a kernel or the host requires.
 The \gls{accessor} object defines a requirement to access a memory
-object from a \gls{command-group}. 
+object, and this requirement is defined by
+construction of an accessor, regardless of whether there are any uses in a kernel.
```
keryell:
```suggestion
construction of an accessor, regardless of whether there are any uses in a kernel or by the host.
```
? Or just stopping the sentence after "uses"?

latex/architecture.tex
```diff
@@ -475,11 +475,12 @@ \subsection{SYCL Application Memory Model}
 consistent access to the memory from the different devices to the memory
 objects.
 
-To access memory objects from inside a kernel, the user must create an
+To access memory objects, the user must create an
 \gls{accessor} object which parameterizes the type of access
-the kernel requires. 
+that a kernel or the host requires.
```
keryell: Since you are working on it, the sentence should be clarify about the fact we need 1 accessor per memory object, not only one for all the memory objects.

mkinsner: Good point.  I think I've clarified this with the latest commit

Issue #66, opened 2020-01-09
Ambiguous queue constructors due to arbitrary property_list type


tomdeakin:
The following two queue constructors are not necessarily uniquely identifiable where the default `property_list` is used:
```
cl::sycl::queue::queue(const cl::sycl::device_selector&, const async_handler&, const cl::sycl::property_list& = {})
cl::sycl::queue::queue(const cl::sycl::device_selector&, const cl::sycl::property_list&)’
```

For instance, I saw that the use below was calling the second constructor, when I would expect the first:
```
  cl::sycl::queue *queue = new cl::sycl::queue(dev, [&](cl::sycl::exception_list l)
  {
    // omitted ...   
  }); 
```

@illuhad identified the issue is caused by the `property_list` constructor accepting arbitrary types, and so my lambda function is used to construct a `property_list` rather than an `async_handler`!

There are a couple of solutions:
1. Define a new `cl::sycl::property` type that the `property_list` constructors can accept instead of any type.
2. Add `explicit` to the constructors.
3. Rely on implementors to get it right! This would be bad, but it's what happens in TriSYCL currently.

Some more context can be found in the [hipSYCL Issue](https://github.com/illuhad/hipSYCL/issues/185) I raised. The resolution to the second constructor there was causing other compile errors later.

keryell:
This is a typical use case for C++20 concepts...
In the meantime it has been implemented with SFINAE in https://github.com/triSYCL/triSYCL/blob/dc50058eb5d83b8cdddd6a43c994f76ea04a8b0e/include/triSYCL/property_list.hpp#L44

tomdeakin:
@keryell So is that a vote for option 3 for now? Is there a compelling reason why we can't clarify the 1.2.1 spec within the bounds of C++11 that it has, and then change the constructors/spec for a future version if we want to exploit future C++ standards?

My main issue with this is that the compiler won't necessarily warn if it's made an implicit conversion, so we don't really know if the compiler has set up a queue with a `async_handler` or not.

keryell:
We need to clarify the specification in any of the 3 cases. Even for relying on the implementation (choice 3) would require specifying what is expected.
Perhaps the choice 2 in the simplest.
Choice 1 is possible but does not seem to me very generic compared to where C++ is going with concepts.

tomdeakin:
[...]

PR #78, opened 2020-03-09
Add explicit keyword to Queue constructors

tomdeakin:
Implicit type conversions were causing conflicts between different
constructors. In particular, it wasn't defined whether a given
argument was a property_list or an async_handler.

This would close Issue #66 .


tomdeakin:
I see what you mean @illuhad, but I think that's preferable to implicit casting resulting in a compiler-dependent choice of constructor.

I'm not sure of the benefit of only making some of the constructors `explicit` either.

PR #73, opened 2020-02-20
Generalized use of property list

steffenlarsen:
Adds optional `property_list` parameters to all user-constructible SYCL
object constructors that did not have it before.

This is the wording related [this proposal](https://github.com/codeplaysoftware/standards-proposals/blob/master/general_property_lists/general_property_lists.md) which has been discussed internally.

PR #81, opened 2020-03-18
Add clarification to property_list constructor and supporting traits

AerialMantis:
This change resolves a potential ambiguity in SYCL class constructors such as those which take an `async_handler` which can occur in some implementations. It also supports another clarification https://github.com/KhronosGroup/SYCL-Docs/pull/73 which extends support for properties to a number of other SYCL classes.

* Add minor fix to the properties example.
* Introduce `is_property` and `is_property_of` traits.
* Add clarification to `property_list` constructor such that it is only
available if all of the variadic parameters passed to it are of types
which are valid properties, evaluated via `is_property`.

### DIFF
--- a/latex/code/propertyExample.cpp
+++ b/latex/code/propertyExample.cpp
@@ -16,9 +16,9 @@
   context myContext;
 
   std::vector<buffer<int, 1>> bufferList {
-    buffer<int, 1>{},
-    buffer<int, 1>{property::use_host_ptr{}},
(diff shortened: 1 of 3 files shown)


fraggamuffin:
mar 19: 
- minor fix to an example, introduces traits that can detect if property is valid
- only when each of variadic type parameter evaluates to true for the type property
- this has to come before PR73
- CTS needed? there are new traits, seems yes; ok will put up CTS change
- OK likes the direction.

PR #82, opened 2020-03-23
clarified the buffer conversion to C++ pointer types

bso-intel:
According to https://github.com/KhronosGroup/SYCL-Docs/pull/44,
we already decided not to support coversion of a buffer to a C++ pointer type.
Now, the user can use the method `get_pointer()` instead.
This patch just clarifies this new decision in the architecture section.

Signed-off-by: Byoungro So <byoungro.so@intel.com>


keryell:
Thanks!

Issue #40, opened 2019-09-07
Namespace for half data type: cl::sycl or no namespace?


bader:
Khronos SYCL-CTS validates that SYCL implementation defines `half` data type in `cl::sycl` space, although it's not clearly stated in the spec that this type must be defined.
Here is all the relevant references in the spec regarding `half` data type.

> **4.10.1 Scalar data types.**
> ...
> Additional scalar data types which are supported by SYCL within the cl::sycl namespace are described in Table 4.93
> ...
> For the purpose of interoperability and portability, SYCL defines a set of aliases to C++ types within the cl::sycl namespace using the cl_ prefix. These aliases are described in Table 4.94
>...
> cl_half 
> Alias to a 16-bit floating-point. The half data type must conform to the IEEE 754-2008 half precision storage format. A SYCL feature_not_supported exception must be thrown if the half type is used in a SYCL kernel fucntion which executes on a SYCL device that does not support the extension khr_fp16

Should we extend the Table 4.93 with `half` data type?

`half` data type is mentioned in 

> **4.10.2.2 Aliases**
> SYCL provides aliases for vec<dataT, numElements> as <dataT><numElements> for the data types: char, short, int, long, float, double, half, cl_char, cl_uchar, cl_short, cl_ushort, cl_int, cl_uint, cl_long, cl_ulong, cl_float, cl_double and cl_half and the data types: signed char , unsigned char, unsigned short, unsigned int, unsigned long, long long and unsigned long long represented with the short hand schar, uchar, ushort, uint, ulong, longlong and ulonglong respectively, for number of elements: 2, 3, 4, 8, 16. For example the alias to vec<float, 4> would be float4.
> 

NOTE: it's not clear here whether `half` is expected to be defined in `cl::sycl` namespace or not. It doesn't seem so as this list includes C++ fundamental types, which are not in `cl::sycl` namespace.

> **5.1 Half Precision Floating-Point**
> The half scalar data type: half and the half vector data types: half1, half2, half3, half4, half8 and half16 must be available at compile-time. However if any of the above types are used in a SYCL kernel function, executing on a device which does not support the extension khr_fp16, the SYCL runtime must throw a feature_not_supported exception.
> 
> The conversion rules for half precision types follow the same rules as in the OpenCL 1.2 extensions specification [2, par. 9.5.1].
> 
[...]

PR #83, opened 2020-03-24
Move description of "half" to the right table

keryell:
half is not a C++ native type but an additional SYCL scalar data type.

This should close https://github.com/KhronosGroup/SYCL-Docs/issues/40


keryell:
> LGTM. I believe that size_t has the same problem and should be moved as well. It's not a fundamental type in C++.

Good point.

Since I was working on it I also removed `std::size_t` from the table about fundamental types.

Since it does exist in C++ already it does not to be declared under `cl::sycl::`.

PR #87, opened 2020-04-01
Clarify range of zero dimension accessors

gmlueck:
Clarify that zero dimension accessors to a buffer, access the first
element of the buffer.  Previously, it just said "a single element" of
the buffer.


keryell:
The committee feedback is good! Thanks.

PR #95, opened 2020-04-23
cl::sycl::event::get() method is supposed to be a constant.

ddpagan:
get() method is not supported from constant cl::sycl::event object, but it is supported from constant cl::sycl::context, cl::sycl::device, cl::sycl::kernel, cl::sycl::platform, and cl::sycl::program.

### DIFF
--- a/latex/headers/event.h
+++ b/latex/headers/event.h
@@ -23,7 +23,7 @@ class event {
 
   /* -- common interface members -- */
 
-  cl_event get();
+  cl_event get() const;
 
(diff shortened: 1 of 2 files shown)


bader:
Should we improve CTS to cover "the const case"?

keryell:
I do not sure. I have the feeling it would be some kind of negative testing.

bader:
> I do not sure. I have the feeling it would be some kind of negative testing.

Why do you think so?

PR #101, opened 2020-06-27
Non-uniform work-group sizes in SYCL

ddpagan:
SYCL 1.2.1 doesn't impose any restrictions for the relationship between global
and local sizes in nd_range<> but should because of OpenCL 1.2 restrictions.

### DIFF
--- a/latex/expressingParallelism.tex
+++ b/latex/expressingParallelism.tex
@@ -1411,7 +1411,10 @@ Table~\ref{table.members.handler.kernel} lists all the members of the
       Defines and invokes a \gls{sycl-kernel-function} as a lambda function
       or a named function object type,
       for the specified \gls{nd-range} and given an \gls{nd-item}
-      for indexing in the indexing space defined by the \gls{nd-range}.
+      for indexing in the indexing space defined by the \gls{nd-range}. If the
+      global size defined in the associated \gls{nd-range} is not evenly divisible
+      by the local size in each dimension then an \codeinline{nd_range_error} SYCL exception
+      must be thrown.
       If it is a named function object and the function object type is globally
       visible there is no need for the developer
       to provide a \gls{kernel-name} (\codeinline{typename KernelName}) for it,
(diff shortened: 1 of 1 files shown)


fraggamuffin:
LGTM
Wait for more approvals

keryell:
Besides updating this for SYCL 1.2.1, I am unsure about the status in the SYCL 2020 specification, since I have not seen anything about it.

gmlueck:
I think some discussion might be needed here.  Although OpenCL 1.2 does not support non-uniform work group sizes, OpenCL 2.x does.  Are we saying that SYCL cannot expose this feature to applications?  Maybe this should be an optional limitation that is imposed only by some backends?  If we go that route, we should provide some way for the application to query for this support, for example either by adding an `info::device` descriptor or by adding an `aspect`.

keryell:
> I think some discussion might be needed here. Although OpenCL 1.2 does not support non-uniform work group sizes, OpenCL 2.x does. Are we saying that SYCL cannot expose this feature to applications? Maybe this should be an optional limitation that is imposed only by some backends? If we go that route, we should provide some way for the application to query for this support, for example either by adding an `info::device` descriptor or by adding an `aspect`.

I think the clarification here is about SYCL 1.2.1 targeting a priori OpenCL 1.2, even of course there has been other implementations not targeting OpenCL 1.2, not even talking about SYCL 2020. But then this is kind of outside of the spec.
And we do not want to add new aspects to SYCL 1.2.1...
So do we agree this is a reasonable fix for SYCL 1.2.1?

gmlueck:
> So do we agree this is a reasonable fix for SYCL 1.2.1?

Yes, I agree.

Actually, this could be reasonable for SYCL 2020 also.  I was talking with @Pennycook and he pointed out that we would need other changes in SYCL 2020 to support non-uniform work group sizes.  Perhaps it is best to just say this is not supported in SYCL 2020 either, and we can consider better support in the next version of the spec.

ProGTX (approved):

keryell (approved):

fraggamuffin:
LGTM 1.2.1
AR for 2020
separate for NEXT

gmlueck:
I created #135 which makes the same change to SYCL 2020.

keryell:
Thanks for your insight!

PR #116, opened 2021-03-08
Minor fixes to spec constant API

gmlueck:
* The declaration of `specialization_id` defines a class type named
  `value_type` for the underlying type of the specialization constant.
  However, all the declarations for `get_specialization_constant()` and
  `set_specialization_constant()` referred to the type name as just
  `type`.  We need to use the same name in both places.  This commit
  uses the name `value_type` consistently, but I'm happy to use `type`
  instead of we prefer.

* During a previous update to the specialization constant API I removed
  the method `kernel_handler::has_specialization_constant()` because I
  did not think it could be implemented in a useful way.  However, I
  forgot to remove the declaration from the synopsis.  Do that now.

* The synopsis for the `handler` class was missing the declaration of
  `set_specialization_constant()` even though we described it later in
  the specification.  Add this missing declaration.

* Some of the member function declaration for specialization constants
  used the template parameter name `SpecName` while others used the
  template parameter name `S`.  Consistently use the name `SpecName` to
  make it clear that all these template parameters mean the same thing.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13778,7 +13778,7 @@ another invocation of the same <<command-group-function-object>>.
 ----
 template<auto& SpecName>
 void set_specialization_constant(
-    typename std::remove_reference_t<decltype(SpecName)>::type value);
+    typename std::remove_reference_t<decltype(SpecName)>::value_type value);
 ----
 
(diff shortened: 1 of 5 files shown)


keryell (approved):

ProGTX (approved):

fraggamuffin:
prefer value_type
LGTM

PR #117, opened 2021-03-09
Add Github Actions CI script 

oddhack:
To build HTML and PDF specs on push or manual selection.

Ready to merge - this has no impact on the actual spec.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,10 +1,13 @@
+# Copyright 2021 The Khronos Group, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
 # CI to build asciidoctor spec targets, on push or manually
 
 name: CI
 
 # Controls when the action will run.
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  # Triggers the workflow on push or manual dispatch
   push:
 
   # Allows you to run this workflow manually from the Actions tab
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,13 +22,17 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
+      # Unfortunately, asciidoctor-pdf gets pathname-specific errors
+      # building under the usual $GITHUB_WORKSPACE (/__w). As a workaround,
+      # generate the outputs in /tmp.
+
       - name: Build spec targets
         run: |
           cd adoc
-          make QUIET= html pdf
+          make OUTDIR=/tmp/out QUIET= html pdf
       - name: Archive generated files
         uses: actions/upload-artifact@v2
         with:
           name: spec-outputs
           path: |
-            adoc/out/
+            /tmp/out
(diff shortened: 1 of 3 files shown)


oddhack:
This is weird, Actions complains about not finding some of the intermediate image files and fails - but when I run exactly the same build on exactly the same branch using exactly the same Docker image locally, it works. It also works in internal gitlab CI using the same image. This may be painful to debug.

oddhack:
@keryell @gmlueck I put in a workaround to allow CI to work until there's a fix for the asciidoctor path problem.

The repo README is entirely incorrect as it refers to gitlab CI and a different Docker image (I have refactored and renamed the image to not imply Vulkan specific contents). If you want I can fix that up here before merging. We should probably have two sections, one for gitlab and one for github.

keryell:
> @keryell @gmlueck I put in a workaround to allow CI to work until there's a fix for the asciidoctor path problem.

It looks like it work!

> The repo README is entirely incorrect as it refers to gitlab CI and a different Docker image (I have refactored and renamed the image to not imply Vulkan specific contents). If you want I can fix that up here before merging. We should probably have two sections, one for gitlab and one for github.

Yes, probably having 2 sections is the way to go.
Thanks.

oddhack:
@keryell @gmlueck I've updated the README. This can be merged at any time.

oddhack:
@keryell note as well that the Docker image now being used has been updated to the current rouge gem, so should not need the 'dock' target.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,38 @@
+# CI to build asciidoctor spec targets, on push or manually
+
+name: CI
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
```
gmlueck, line 8: Is the comment correct here that the workflow is triggered on both "push" and "pull"?  I don't know this yaml file scheme, but since it just says "push:" here, it makes me think the action is triggered only on "push" requests.

Also out of curiosity, why is the workflow limited to just the "master" branch?  I don't see that branch name anywhere.  Is it the default?

oddhack: Good catch, thanks. Those are left over comments from a different branch I copied a starting point from.

oddhack: Good catch, thanks. Those are left over comments from a different branch I copied a starting point from.

oddhack: N.b. gitlab CI YAML is bad but github Actions is worse.

ProGTX (approved):

keryell (commented):
> @keryell note as well that the Docker image now being used has been updated to the current rouge gem, so should not need the 'dock' target.

Yes, I think it was not used on the CI, just for local developments.

oddhack:
> > @keryell note as well that the Docker image now being used has been updated to the current rouge gem, so should not need the 'dock' target.
> 
> Yes, I think it was not used on the CI, just for local developments.

The gem repository has updated to 3.26.0 - is something even newer needed?

keryell:
Thanks!

PR #118, opened 2021-03-11
Fix unary vector operator to take a const reference

tomdeakin:
Reported by Daniel Geon Park via the Khronos Slack.

> Without a `const` reference, it is impossible to do something like:
>```
>const sycl::float4 v{};
>-v; // Error
>```
>
>Or:
>```
>-sycl::float4{}; // Error
>```

This PR adds the const reference.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -16493,7 +16493,7 @@ Where [code]#OP# is: [code]#pass:[++]#, [code]#--#.
 a@
 [source]
 ----
-vec operatorOP(vec &v)
+vec operatorOP(consnt vec &v)
 ----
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -16493,7 +16493,7 @@ Where [code]#OP# is: [code]#pass:[++]#, [code]#--#.
 a@
 [source]
 ----
-vec operatorOP(vec &v)
+vec operatorOP(consnt vec &v)
```
gmlueck, line 16496:
```suggestion
vec operatorOP(const vec &v)
```

tomdeakin: Good stop - my keyboard has a dodgy `n` key.

keryell (approved):
Thanks for the fix.

ProGTX (approved):

PR #121, opened 2021-03-11
Clarify buffer reinterpret w/o range

gmlueck:
The overload of `buffer::reinterpret()` that takes no range parameter
was defined to throw an exception if the reinterpreted range does not
equal the original buffer's range.  However, since there is no range
parameter, this can only happen if the reinterpreted element size does
not evenly divide the buffer's size.  Clarify the spec by saying this
explicitly.


keryell (approved):

ProGTX (approved):

fraggamuffin:
LGTM

Issue #115, opened 2021-03-06
Unary plus and minus operators for sycl::vec takes a non-const reference to the value


mujjingun:
From SYCL 2020 specification (revision 2) page 351:

```cpp
// OP is unary +, -
friend vec operatorOP(vec &rhs) const { /* ... */ }
```

Is there a reason why this takes a non-const lvalue reference? This makes it impossible to do something like:

```cpp
const sycl::float4 v{};
-v; // Error: cannot bind const vec& to vec&
```

or 

```cpp
-sycl::float4{}; // Error: cannot bind vec&& to vec&
```

In my opinion, the line in the spec should be corrected to:

```cpp
// OP is unary +, -
friend vec operatorOP(const vec &val) { /* ... */ }
```

tomdeakin:
Fixed in #118

mujjingun:
@tomdeakin Shouldn't the `const` qualifier *after* the function signature be removed too? It doesn't make sense to have it in a non-member friend function.

tomdeakin:
Good catch, I'll re-open the PR.

keryell:
I think this is fixed now.

PR #122, opened 2021-03-12
Remove const function qualifier from non-member unary operator function

tomdeakin:
Further fix for #118 and #115.

We want to make sure the non-member unary operator is not a `const` function. It's not clear that this is legal C++. The fix removes the `const`.


keryell (approved):
We are converging! :-)

fraggamuffin:
LGTM

PR #124, opened 2021-03-16
Add read-write host unsampled image accessor

gmlueck:
A reviewer noted to me that SYCL 1.2.1 supported a read-write image
accessor for `target::host_image`.  SYCL 2020 inadvertently dropped
this functionality because `host_unsampled_image_accessor` does not
support read-write access mode.  Add that functionality back, so we
are again on parity with SYCL 1.2.1.

Adding a read-write mode also has a happy side-effect.  This is now
a reasonable default mode, so change the synopsis to show read-write
as the default (unless `dataT` is `const`).  Adding a default for
`accessMode` also allows us to add a default for `dimensions`, so
make that default to `1`, which matches the default for
`unsampled_image` and the default for the other accessors.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -8370,9 +8370,11 @@ within a <<host-task>>.  The latter provides access from host code that is
 outside of a <<host-task>>.
 
 The dimensionality of an unsampled image accessor must match the dimensionality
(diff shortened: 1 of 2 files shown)


gmlueck:
This addresses one of the issues reported in #126

keryell (approved):

ProGTX (approved):

fraggamuffin:
LGTM

PR #125, opened 2021-03-16
Replace std::dynamic_extent with sycl:: version

Pennycook:
Adds std::dynamic_extent to the list of pre-adopted C++20 features.
This was only implied before (by pre-adoption of std::span).


keryell (approved):
Good catch!

ProGTX (approved):

fraggamuffin:
LGTM

Issue #126, opened 2021-03-17
SYCL 2020 accessors issues.


bader:
Here is the summary of discussion between @gmlueck and @yuriykoch regarding issues in accessors section of SYCL 2020 specification.

1) An accessor with deprecated target `local` has method `is_placeholder()`. But there was no such method for local accessor in SYCL 1.2.1. We should remove this method from SYCL 2020 specification.

2) Par. 3.13.3 states that: 
    “Inside kernels, the functions and data types available are restricted by the underlying capabilities of SYCL backend devices.”

    That statement from section [3.13.3](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_functions_and_data_types_available_in_kernels) is not very useful because it doesn’t explain how an application can tell which data types are restricted by the backend.  We describe this better, though, in other parts of the spec.  For example, [4.6.4.3](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-aspects) states that the “sycl::half” type is only available on a device if the device has “aspect::fp16”. We suggest removing 3.13.3 paragraph.

3) local_accessor description states that: “The underlying dataT type can be any C++ type”. 

    We suggest changing this sentence to “The underlying dataT type can be any C++ type that the device supports.”

    The intent is to say that local accessors do not impose any further restrictions on the underlying type beyond those restrictions on the device code in general.

4) Accessor API changes missing in "What has changed ..." section. Looks like D.1 appendix may be extended with the following:
* There is no explicit mention of `accessor::value_type` and `accessor::reference` type dependency on the access mode. The spec says “Member functions of accessor which return a reference to an element have been changed to return a const reference for read-only accessors.”, but we should tweak that sentence to say that the member types are also changed for read-only accessors.

* There is no explicit mention of read_write access mode removal for image accessors. Even SYCL 1.2.1 did not allow read-write access mode for images, unless the target was “host_image”.  It was probably an oversight that “host_unsampled_image_accessor” does not provide this ability.

5) p. 180, get_range() description for the target ‘local’ has misprint:
                “the returned value is the the range”

gmlueck:
I think the issues reported here are addressed in the following PRs:

* #124
* #127
* #128
* #129

We can close this issue when those PRs are merged.

bader:
> I think the issues reported here are addressed in the following PRs:
> 
> * #124
> * #127
> * #128
> * #129
> 
[...]

PR #127, opened 2021-03-17
Local accessor should not have is_placeholder

gmlueck:
The deprecated form of `accessor` with `target::local` did not have a
member function named `is_placeholder` in SYCL 1.2.1 because there is
no such thing as a local placeholder accessor.  The SYCL 2020 spec
inadvertently added `is_placeholder` as part of the deprecated API,
which was not our intent.  Remove it.

This addresses one of the issues reported in #126.


keryell (approved):

ProGTX (approved):

fraggamuffin:
LGTM

PR #128, opened 2021-03-17
Remove vague architecture section

gmlueck:
This section from the architecture chapter seems too vague, and I
don't think it's entirely true.  My belief is that all backends are
supposed to support all core SYCL features, except those that are
specifically described as optional.  Those optional features are
already described better in section 4.6.4.3. "Device aspects" and
section 5.7. "Optional kernel features".

Since this section does not seem to specify any new requirement or
clarify anything, I propose that we just remove it.

This addresses one of the issues reported in #126.


keryell (approved):
Why not...

fraggamuffin:
LGTM

PR #129, opened 2021-03-17
A few minor fixes

gmlueck:
* Clarify that the underlying type of a local accessor can be any C++
  type **that the device supports**.

* Note that the accessor `value_type` and `reference` class types have
  changed since 1.2.1 for read-only accessors.

* Fix a grammatical typo.

These address issues reported in #126.


keryell (approved):

ProGTX (approved):

fraggamuffin:
LGTM

PR #130, opened 2021-03-17
Clarify accessor target of explicit copy ops

gmlueck:
The main change in this commit is to clarify that the `target` template
parameter for accessors used in the explicit copy operations must be
either `target::device` (aka `target::global_buffer`) or
`target::constant_buffer`.  Although, `target::constant_buffer` is
deprecated, I could not think of any reason to disallow it, so I
thought we should allow it for backward compatibility with 1.2.1.

Some other minor clarifications are:

* Remove the sentence saying that the accessor must be a "buffer
  accessor".  This is redundant now that we state that the only
  allowable values for the `target` parameter are `target::device` and
  `target::constant_buffer` (which are both buffer accessors).

* Use the word "must" instead of "may" when describing the allowable
  accessor modes.

Here is my justification for why the other `target` parameters should
not be allowed:

* The spec already disallows `target::host_buffer` because we state
  that this target only allows access outside of a "command", and we
  define "command" to include the explicit copy operations.  Therefore,
  `target::host_buffer` cannot be allowed.

* The spec already disallows `target::local` because that is not a
  "buffer accessor", and the spec stated that a buffer accessor was
  required.

* Allowing `target::host_task` does not make sense to me.  If the
  application wants to explicitly copy memory to the host, it should
  use a host pointer as a parameter to the explicit copy, or it should
  use the `update_host()` member function instead.  Since
  `target::host_task` is new in SYCL 2020, there is no issue of
  backwards compatibility.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13390,19 +13390,19 @@ Manual copy operations can be seen as specialized kernels executing on the
 device, except that typically this operations will be implemented using a
 host API that exists as part of a backend (e.g, OpenCL enqueue copy operations).
 
-The SYCL memory objects involved in a copy operation are specified using
-buffer accessors.
-Explicit copy operations have a source and a destination.
-When an accessor is the _source_ of the operation, the destination can be
-a host pointer or another accessor.
-The _source_ accessor can have either [code]#read# or
-[code]#read_write# access mode.
+These explicit copy operations have a source and a destination.  When an
+accessor is the _source_ of the operation, the destination can be a host
+pointer or another accessor.  The _source_ accessor must have either
+[code]#access_mode::read# or [code]#access_mode::read_write# access mode.  When
+an accessor is the _destination_ of the explicit copy operation, the source can
+be a host pointer or another accessor.  The _destination_ accessor must have
+either [code]#access_mode::write#, [code]#access_mode::read_write#,
+[code]#access_mode::discard_write# or [code]#access_mode::discard_read_write#
+access mode.
 
-When an accessor is the _destination_ of the explicit copy operation,
-the source can be a host pointer or another accessor.
-The _destination_ accessor can have either
-[code]#write#, [code]#read_write#, [code]#discard_write#,
-[code]#discard_read_write# access modes.
+When an accessor is used as a parameter to one of these explicit copy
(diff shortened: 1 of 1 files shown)


keryell (approved):

fraggamuffin:
Wait a week.

keryell:
On the other hand you can pass a host pointer you got from a `host_accessor`, so it could make sense to allow it by symmetry.

gmlueck:
If we were to allow `target::host_task`, I'm trying to think what that would mean.  If an accessor is the source of the copy, it seems like the `target` is meaningless.  The implementation will just copy the data from wherever it currently resides.  Therefore, if we allow `target::host_task`, I think we would document that there is no difference between `target::host_task` and `target::device` when used in the source accessor.

If an accessor is the destination of the copy and the target is `target::device`, then the semantic seems clear.  The data is copied to the queue's device.  If the target is `target::host_task`, I presume the semantic would be to copy the data to the host (so that a subsequent read from that buffer via either a host task accessor or `host_accessor` would be fast).

Originally, I was thinking that this provides no benefit beyond `update_host()`, but this is not quite true because `update_host()` only allows you to move the location of data that is already controlled by the accessor.  By contrast, `copy(<source>, accessor<target::host_task>)` can copy data from a different buffer into the accessor AND also move that data to the host.  It does seem like this could be useful.

So, I think I'm changing my mind about this now.  If others agree, I can update this PR.

BTW, I'm starting to wonder now about the semantics of `handler::memcpy()`.  What if the `dest` pointer is a USM pointer that is shared between host and device?  Is the expectation that the copied data is available for fast access from the queue's device?  What if the application wants fast access from the host instead?  (I'm assuming here that the implementation would do something different to move the data closer to the device vs. closer to the host in the two cases.)  Tagging @jbrodman

gmlueck:
After much discussion in the Khronos meeting today, we decided to disallow `target::host_task` as a possible target for these explicit copy functions.  Our reason was that the semantics of this target are not entirely clear, and we wanted more time to think about it.  Disallowing `target::host_task` is a conservative change because we can always allow it later if we want.

This PR is already written to disallow `target::host_task`, so I think it is ready to merge.

illuhad (approved):

fraggamuffin:
LGTM

PR #132, opened 2021-04-01
Fixes to reducers

nliber:
Internal issue #535

In the reducers section:

Made member functions public
Added aliases for template parameters
Fixed deleting copy/move constructors/assignment operators
Changed the return type of `combine` to `reducer&`
Made operators hidden friends and changed their return type to `reducer&`
Fixed constraints with `Dimensions` to be consistent between code and prose (`Dimensions` start at `0`) 
Fixed `combine` to be a non-const member function in prose

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12594,7 +12594,7 @@ a@
 ----
    a@ Equivalent to calling [code]#accum.combine(partial)#.
       Available only when: [code]#Dimensions == 0 &&
-      (is_same_v<BinaryOperation, plus<>> || is_same_v<BinaryOperation, plus<T>>)#.
+      (std::is_same_v<BinaryOperation, plus<>> || std::is_same_v<BinaryOperation, plus<T>>)#.
 
 a@
 [source]
@@ -12603,7 +12603,7 @@ a@
 ----
    a@ Equivalent to calling [code]#accum.combine(partial)#.
       Available only when: [code]#Dimensions == 0 &&
-      (is_same_v<BinaryOperation, multiplies<>> || is_same_v<BinaryOperation, multiplies<T>>)#.
+      (std::is_same_v<BinaryOperation, multiplies<>> || std::is_same_v<BinaryOperation, multiplies<T>>)#.
 
 a@
 [source]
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -12578,65 +12580,66 @@ T identity() const
...
 a@
 [source]
 ----
-template <typename T>
-    void operator+=(reducer<T,plus<T>,0>& accum, const T& partial)
+    reducer& operator+=(reducer& accum, const T& partial)
 ----
    a@ Equivalent to calling [code]#accum.combine(partial)#.
+      Available only when: [code]#Dimensions == 0 &&
+      (is_same_v<BinaryOperation, plus<>> || is_same_v<BinaryOperation, plus<T>>)#.
```
keryell, line 12597: Should we have everywhere `std::is_same_v`?

adoc/headers/reducer.h
```diff
@@ -4,39 +4,51 @@
...
+public:
 
-  reducer(const reducer<T,BinaryOperation,Dimensions>&) = delete;
-  reducer<T,BinaryOperation,Dimensions>& operator(const reducer<T,BinaryOperation,Dimensions>&) = delete;
+  using value_type = T;
+  using binary_operation = BinaryOperation;
+  static constexpr int dimensions = Dimensions;
+
+  reducer(const reducer&) = delete;
+  reducer(reducer&&) = delete;
```
ProGTX, line 14: Are we sure we want to make `reducer` non-movable? What's the benefit?

Pennycook: If I remember correctly, we wanted to permit implementations of reducer that looked like this:
```c++
template <typename T>
struct reducer {
  T _data;
};
```
...in which case, the developer should always pass around a reference to the reducer that was created by the SYCL implementation rather than creating copies of it.

I think deleting these constructors may also stop a developer from accidentally sharing a reducer between multiple work-items (e.g. by writing the reducer out to memory).

adoc/chapters/programming_interface.adoc
nliber: Should `plus`, etc. also be `std::plus`?

adoc/headers/reducer.h
nliber: It was clear that the intent was to make it non-copyable and non-moveable.

(I'm not saying we shouldn't discuss it; rather, that seems like more than a bug fix or minor improvement of something not previously considered.)

adoc/chapters/programming_interface.adoc
keryell: If we talk about something from the `std::`, yes.

keryell (approved):

ProGTX (approved):

fraggamuffin:
LGTM

PR #133, opened 2021-04-07
Clarify spec constant requirements

gmlueck:
This commit adjusts the requirements for declaring `specialization_id`
variables after doing a deeper analysis of how they can be implemented.
Following are the major changes:

* Add a requirement that the underlying type `T` must be device
  copyable.  This is necessary because the specialization constant's
  value could be copied from host to device in some implementations.
  We think this restriction was intended all along.

* Add a requirement that the `specialization_id` variable must be
  declared `constexpr`.  We intended all along to require the default
  value to be known at compile time, but the previous wording didn't
  state this.  By requiring the `specialization_id` variable to be
  `constexpr`, we also require the arguments to the constructor (the
  default value) to be known at compile time.

* We think that applications may want to use `marray` or `vec` as the
  underlying type of a `specialization_id` variable.  Make their constructors
  `constexpr` (where appropriate) to make this possible.

* Remove the requirement that the `specialization_id` must be declared
  with static storage duration.  This is redundant now that we require
  it to be declared `constexpr` since all namespace scope variables
  have static storage duration and all class scope `constexpr`
  variables also have static storage duration.

* Add a requirement that a `specialization_id` variable declared in
  class scope must be publicly accessible.  We think that a
  common implementation will generate a post-integration header file
  that is conceptually included at the end of the translation unit, and
  this header will reference the name of each `specialization_id`
  variable.  This is not possible if a class scope `specialization_id` variable
  has non-public access.

* Add a requirement that a `specialization_id` variable declared in
  namespace scope may not be "shadowed" by another variable
  with the same name declared in an `inline` namespace, such that
  the `specialization_id` variable is no longer accessible.  (The spec has
  an example showing this situation.)  Again, this situation would
  prevent the post-integration header from accessing the `specialization_id`
  variable.  We think this is a very weird corner case which will not
  affect any reasonable applications.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13725,16 +13725,12 @@ class with the following restrictions:
 
 * the template parameter [code]#T# must be a <<device copyable>> type;
 * the [code]#specialization_id# variable must be declared as [code]#constexpr#;
-* the [code]#specialization_id# variable must be declared in one of the
-  following scopes:
-  - namespace scope (including the anonymous namespace);
-  - class scope where the class is declared in namespace scope;
-  - class scope where the class is declared in some other class scope, and the
-    outermost class is declared at namespace scope;
+* the [code]#specialization_id# variable must be declared in either namespace
+  scope (including the anonymous namespace) or in class scope;
 * if the [code]#specialization_id# variable is declared in the anonymous
   namespace, its name must be unambiguous when referenced from the global
   namespace scope (i.e. it must not have the same name as some other variable
-  declared in a named namespace);
+  declared in different namespace);
 * if the [code]#specialization_id# variable is declared in class scope, it must
   have public accessibility when referenced from namespace scope.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13726,13 +13726,22 @@ class with the following restrictions:
 * the template parameter [code]#T# must be a <<device copyable>> type;
 * the [code]#specialization_id# variable must be declared as [code]#constexpr#;
 * the [code]#specialization_id# variable must be declared in either namespace
-  scope (including the anonymous namespace) or in class scope;
-* if the [code]#specialization_id# variable is declared in the anonymous
-  namespace, its name must be unambiguous when referenced from the global
-  namespace scope (i.e. it must not have the same name as some other variable
-  declared in different namespace);
+  scope or in class scope;
 * if the [code]#specialization_id# variable is declared in class scope, it must
-  have public accessibility when referenced from namespace scope.
+  have public accessibility when referenced from namespace scope;
+* the [code]#specialization_id# variable may not be shadowed by another
+  variable [code]#V# which has the same name and is declared in an
+  [code]#inline# namespace, such that the [code]#specialization_id# variable is
+  no longer accessible after the declaration of [code]#V#.
+
+[NOTE]
+====
+The expectation is that some implementations may conceptually insert code at
+the end of a translation unit which references each `specialization_id`
+variable that is declared in that translation unit.  The restrictions listed
+above make this possible by ensuring that these variables are accessible at the
+end of the translation unit.
+====
 
 The following example illustrates some of these restrictions:
 
@@ -16165,7 +16174,7 @@ vec()
 a@
 [source]
 ----
-explicit vec(const dataT &arg)
+explicit constexpr vec(const dataT &arg)
 ----
    a@ Construct a vector of element type [code]#dataT# and
       [code]#numElements# dimensions by setting each value to [code]#arg# by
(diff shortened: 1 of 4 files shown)


adoc/code/specialization_id.cpp
```diff
@@ -0,0 +1,35 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <sycl/sycl.hpp>
+using namespace sycl; // (optional) avoids need for "sycl::" before SYCL names
+
+struct Mixed {int i; float f;};
+
+constexpr specialization_id<int> a{1};            // OK
+static constexpr specialization_id<int> b{2};     // OK
+specialization_id<int> c{3};                      // ERROR: not constexpr
```
nliber, line 11: When we say "ERROR", does that mean we expect the compiler to diagnose it?

adoc/chapters/programming_interface.adoc
```diff
@@ -13721,12 +13721,31 @@ overhead associated with recompilation of the kernel's bundle.
...
+  following scopes:
+  - Namespace scope (including the anonymous namespace).
+  - Class scope where the class is declared in namespace scope.
+  - Class scope where the class is declared in some other class scope, and the
+    outermost class is declared at namespace scope.
+* If the [code]#specialization_id# variable is declared in the anonymous
+  namespace, its name must be unambiguous when referenced from the global
+  namespace scope (i.e. it must not have the same name as some other variable
+  declared in a named namespace).
+* If the [code]#specialization_id# variable is declared in class scope, it must
```
keryell, line 13738: Just for the coherence of the document typography:
```suggestion
* the template parameter [code]#T# must be a <<device copyable>> type;
* the [code]#specialization_id# variable must be declared as [code]#constexpr#;
* the [code]#specialization_id# variable must be declared in one of the
  following scopes:
  - namespace scope (including the anonymous namespace);
  - class scope where the class is declared in namespace scope;
  - class scope where the class is declared in some other class scope, and the
    outermost class is declared at namespace scope;
* if the [code]#specialization_id# variable is declared in the anonymous
  namespace, its name must be unambiguous when referenced from the global
  namespace scope (i.e. it must not have the same name as some other variable
  declared in a named namespace);
* if the [code]#specialization_id# variable is declared in class scope, it must
```

adoc/code/specialization_id.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace foo {
+  constexpr specialization_id<Mixed> c{1, 3.14};  // OK
+}
+namespace {
+  constexpr specialization_id<Mixed> d{2, 2.718}; // OK
+}
+
```
keryell, line 19: I have the feeling that
```c++
namespace toto {
  constexpr specialization_id<Mixed> d{3, 4};
}
```
can work since the first `d` can be moved outside of the integration headers while the second one remains as `toto::d`.

nliber: Have we resolved `constexpr` vs. `inline constexpr`?  I'd rather we just mark them `inline constexpr` when they are non-`static` or we can have examples for both.

adoc/code/specialization_id.cpp
gmlueck: I only want to convey that the declaration on this line in non-conformant.  The spec did not say before that the compiler was required to diagnose any errors with declarations of `specialization_id` variables, and I did not want to add that requirement now.

I see that the spec has an existing example which uses the word "illegal" to describe non-conformant code like this, so I think I should change these lines like so:

```
specialization_id<int> c{3};                      // ILLEGAL: not constexpr
```

adoc/code/specialization_id.cpp
gmlueck: @keryell: Are you saying that the application could have two `specialization_id` variables with the same name that are declared in different named namespaces?  Yes, I agree that is legal.  I can add an example to demonstrate this.

@nliber: We identified an implementation that allows `specialization_id` variables to be declared either `inline constepxr` or just `constexpr`.  I can add an example showing the `inline constexpr` case.

gmlueck: I added some more examples in 95154eb0cbcd03171e9adcd2601f491cb8f45864.

I also realized that the restriction on declaring a `specialization_id` variable in class scope could be simplified.  The description before said that it was OK to declare a `specialization_id` variable in class scope so long as the class was not declared in function scope.  However, I realize now that normal C++ rules forbid a `constexpr` variable from being declared in class scope if the class is in function scope.  Therefore, we can simplify the restriction about `specialization_id` variables to just say that it is OK to declare them in class scope.

keryell (approved):

adoc/code/specialization_id.cpp
gmlueck: Responding to @keryell's original question in this thread ... we identified an implementation which allows a `specialization_id` variable to be "shadowed" by an enclosed unnamed namespace, so we removed this restriction from the spec.

Unfortunately, we identified a different shadowing case with `inline` namespaces, so we added a restriction for that.  However, we think this `inline` shadowing is very weird, so we don't think it will be much of a burden.

fraggamuffin:
need more time, aim for next week.

keryell (approved):
Complex, the `inline namespace`.
I wonder whether it is not a can of worms we have ignored in some other parts of the spec...
Like shadowing some kernel names with some implementations?

gmlueck:
> Complex, the `inline namespace`.
> I wonder whether it is not a can of worms we have ignored in some other parts of the spec...
> Like shadowing some kernel names with some implementations?

Yes, I also think that the specification of kernel names could be clarified.  There are probably similar issues.

fraggamuffin:
LGTM

PR #136, opened 2021-04-14
Standard attribute position for non-lambda kernels

gmlueck:
We used to define the position of C++ attributes attached to kernel
functions such that they appertained to the "function-type" of the
kernel function.  This is necessary for lambda functions because the
C++17 standard does not define an attribute position that appertains to
the lambda function itself.  (This is expected to be fixed in C++23.)

However, non-lambda functions have a well-defined position for
attributes that appertain to the function itself.  It's not clear why
the spec did not use this position for these functions.

This commit changes the position of these attributes to the standard
place.

I also made some clarifications about the general restrictions of
these attributes, which were suggested by our front-end experts.  I
think these do not change the intent, but just tighten the wording.


gmlueck:
Our front-end team is asking why SYCL 2020 non-lambda function attributes are defined in a weird position.

I'm opening this draft PR to ask if we really want to define them in a weird position, or if we want to use the standard position for function attributes.  Feedback welcome here, or we can discuss in the next working group meeting.

keryell (approved):

adoc/code/attributes.cpp
```diff
@@ -1,19 +1,27 @@
 // Copyright (c) 2011-2021 The Khronos Group, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-// Kernel defined as a lambda
+// Kernel defined as a lambda.
```
keryell, line 4: I do not think we use `.` at the end of single-sentence comments in the specification.

gmlueck: Fixed here and below.

gmlueck:
Drawing attention to another semantic change in this PR:

Previously, we did not allow the same attribute to be applied twice to the same device function.  Our front end team
says this is more restrictive than standard C++ attributes.  The C++ standard does allow duplicate attributes like `[[attr]][[attr]]`.
Redundant occurrences of the attribute are just ignored in a case like this.

Apparently, C++ does disallow duplicates when they appeared like this: `[[attr,attr]]`.  However, WG21 removed this restriction in
(https://wg21.link/p2156) because they felt that restriction was silly given that duplicates are allowed via `[[attr]][[attr]]`. 
I changed SYCL to adopt the non-silly wording proposed in p2156.

AaronBallman:
Thank you for the changes, they look good to me!

adoc/chapters/device_compiler.adoc
```diff
@@ -456,30 +456,58 @@ to influence the code generated by the device compiler.  These attributes are
...
+kernel or device function in the same translation unit may optionally have the
+same attribute applied (so long as the attribute arguments are the same between
+the declarations), but this is not required.  The attribute remains in effect
+regardless of whether it appears in the redeclaration.
+
+Unless an attribute's description specifically allows it, a kernel or device
+function may not be declared with the more than one instance of the same
+attribute unless all instances have the same attribute arguments.  The compiler
+must issue a diagnostic for programs which violate this requirement.  When two
+or more instances of the same attribute appear on the declaration of a kernel
```
keryell, line 472: It would be interesting if we introduce some attributes like `[[toggle-property-A]] [[toggle-property-A]]`...
P2156 seems just to talk about standard attributes from the C++ which are idempotent.
But it does not say in the general case what is the semantics of repeating attributes and thus whether they are idempotent or not...
Is this sentence important? If not, can we just remove it?
@AaronBallman ?

gmlueck: Are you concerned only with this sentence @keryell?

> When two or more instances of the same attribute appear on the declaration of a kernel or device function, the effect is as though a single instance appeared (assuming that all instances have the same attribute arguments).

I agree that we would need to change that sentence if we ever added an attribute like "toggle", but we don't have any attributes like that now.

I was concerned that the spec wasn't clear about what happens when a attribute is duplicated, which is why I added that sentence.  Do you think it's clear enough if we remove that sentence?

AaronBallman: > I agree that we would need to change that sentence if we ever added an attribute like "toggle", but we don't have any attributes like that now.

I have no idea what you have in mind for toggle, but you may want to be aware of this: http://eel.is/c++draft/dcl.attr#grammar-4.sentence-4

AaronBallman: >But it does not say in the general case what is the semantics of repeating attributes and thus whether they are idempotent or not...
>Is this sentence important? If not, can we just remove it?
>@AaronBallman ?

For standard attributes, the standard says whether they can be repeated or not and what happens if they are. For example: http://eel.is/c++draft/dcl.attr.deprecated#3 and http://eel.is/c++draft/dcl.attr.deprecated#4

For vendor-specified attributes, it's up to the vendor to say what happens (so long as it conforms to the rest of the requirements of the standard like appertainment, etc).

As for why the sentence in SYCL is important -- definitions are also declarations, so you want to allow sensible things like:
```
[[attr(1)]] void whatever();
[[attr(1)]] void whatever() {} // This should not be disallowed
```
The quoted sentence helps to clarify that the above code is fine and behaves as though there's only one attribute associated with the declaration.

fraggamuffin:
LGTM

PR #145, opened 2021-06-02
Global memory and concurrently executing kernels applies only to USM

tomdeakin:
This updates the text on global memory to say that accessing same memory in concurrent kernels only possible with USM memory, and results can be made correct with atomics and mem-fences.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -905,9 +905,9 @@ Work-items executing in a kernel have access to three distinct address spaces
     Work-items can read from or write to any element of a global memory
     object. Reads and writes to global memory may be cached depending on the
     capabilities of the device. Global memory is persistent across kernel
-    invocations. There is no guarantee that two concurrently
-    executing kernels can simultaneously write to the same USM allocation and
-    expect correct results unless <<mem-fence>> and atomic operations are used.
+    invocations. Concurrent access to a location in an USM allocation by two or more executing
+    kernels where at least one kernel modifies that location is a data race; there is no guarantee
+    of correct results unless <<mem-fence>> and atomic operations are used.
   * <<local-memory,Local-memory>> is accessible to all work-items in a single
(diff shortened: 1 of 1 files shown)


gmlueck (commented):
LGTM

adoc/chapters/architecture.adoc
```diff
@@ -905,9 +905,9 @@ Work-items executing in a kernel have access to three distinct address spaces
     Work-items can read from or write to any element of a global memory
     object. Reads and writes to global memory may be cached depending on the
     capabilities of the device. Global memory is persistent across kernel
-    invocations, however there is no guarantee that two concurrently
-    executing kernels can simultaneously write to the same memory object and
-    expect correct results.
+    invocations. There is no guarantee that two concurrently
+    executing kernels can simultaneously write to the same USM allocation and
```
Pennycook, line 909: This may be nit-picking, but this affects more than writes.  For example, there is no guarantee if one kernel writes to a location while another kernel reads from the same location.

I wonder if it's worth trying to align with the language from http://eel.is/c++draft/intro.races and/or http://eel.is/c++draft/atomics here.  Something like the below, maybe?
```
Concurrent access to the same USM allocation by two or more executing kernels is a data race,
and there is no guarantee of correct results unless <<mem-fence>> and atomic operations are used.
```

tomdeakin: You're correct, that needs clarifying. I think we need to be careful that it's only a data race if there is at least one write. If two kernels both read the same location in memory, then that should be fine. This might need further clarification around it's a location in a USM allocation - I'm not sure I meant to imply the entire allocation.

tomdeakin: 78f222c should address the above.

Pennycook (approved):

ProGTX (approved):

keryell (approved):

PR #147, opened 2021-06-10
Adopt consistent naming convention for template arguments

keryell:
Start addressing
https://gitlab.khronos.org/sycl/Specification/-/issues/312


keryell:
This should close https://gitlab.khronos.org/sycl/Specification/-/issues/312

PR #149, opened 2021-06-10
Clarify when host_task events are considered complete

### LINKED ISSUE #140: Clarify behavior of event::wait regarding asynchronicity in interop host_tasks
I couldn't find any information on how `sycl::event::wait` behaves regarding `host_task`s that use an `interop_handle` to submit asynchronous operations to the backend queue underlying the SYCL queue. Consider this example:

```c++
auto e = my_queue.submit([&](sycl::handler& cgh) {
    cgh.host_task([=](sycl::interop_handle& ih) {
        auto stream = ih.get_native_queue<sycl::backend::cuda>();
        cudaMemcpyAsync(dst_ptr, src_ptr, count, cudaMemcpyDefault, stream);
        // cudaStreamSynchronize(stream); // Required?
    };
});

// Does this wait for the callback to return, or for `cudaMemcpyAsync` to be finished?
e.wait();
```

Basically the example says it already, there are two different behaviors one could expect, either:
1) `e.wait()` returns as soon as the callback provided to `handler::host_task` has returned, or
2) it returns only once the callback has returned, and all work on the underlying queue (in this example, the CUDA stream) has finished.

Personally, I would find option 2) much more useful, as option 1) would effectively mean that any form of interop would have to be blocking.

psalz:
Following internal discussion on #140, this clarifies that the `sycl::event` returned by the submission of a `host_task` is considered complete once the invocation of the provided callable has returned.

Resolves #140.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13934,6 +13934,10 @@ invoked directly by the SYCL runtime, regardless of which <<device>> the
 
 A <<host-task>> is enqueued on a <<queue>> via the [code]#host_task#
 member function of the [code]#handler# class.
+The <<event>> returned by the submission of the associated <<command group>> is
+considered complete (corresponding to a status of
+[code]#info::event_command_status::complete#) once the invocation of the
+provided {cpp} callable has returned.
 
 A <<host-task>> can optionally be used to interoperate with the
 <<native-backend-object,native backend objects>> associated with the <<queue>> executing the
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

keryell (approved):

keryell:
If someone put some code in the invocable having whatever side effect, it seems a little bit outside of the scope of SYCL. So I am not sure we need to say more about it.

keryell:
Meeting 2021/07/08:

To clarify in term of state transition.

psalz:
I've reworded it to say
>  The event returned by the submission of the associated command group **enters the completed state** (corresponding to a status of info::event_command_status::complete) once the invocation of the provided C++ callable has returned.

using this wording to be in line with section 4.16.2. Synchronization.

PR #155, opened 2021-07-01
Remove references to independent forward progress

Pennycook:
The info::device::sub_group_independent_forward_progress query and
associated mentions of "independent forward progress" were inherited
from OpenCL during the introduction of sub-groups.

The ISO C++ specification defines its own classes of forward progress
(e.g. "weakly parallel forward progress") and going forward it is
important that SYCL aligns with this terminology.

Removing all mentions of "independent forward progress" from SYCL 2020
does not significantly change functionality or intent:
- OpenCL backends can query sub_group_independent_forward_progress
- Other backends can provide their own queries (via interoperability)

Providing a replacement mechanism to query the ability of work-items
to make forward progress is deferred.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1112,8 +1112,8 @@ supported if and only if the following conditions are true:
   * the work-items being synchronized are guaranteed to make forward progress
     with respect to one another.
 
-The ability of work-items to make forward progress with respect to work-items
-in other work-groups is implementation-defined.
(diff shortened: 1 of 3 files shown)


ProGTX (approved):

keryell (approved):

keryell:
The Khronos Memory Model work group will start discussing this probably in 1 month from now.

SYCL requires to describe more the execution model in the future, like what we have in the C++ specification.

There is no forward progress guarantee yet.

PR #160, opened 2021-07-02
parallel_for_work_group can take function object

gmlueck:
Clarify that the `kernelFunc` parameter to `parallel_for_work_group()`
can be a named function object, just as with `parallel_for()`.

This closes internal issue 233.


keryell (approved):
Nice catch

ProGTX (approved):

PR #161, opened 2021-07-02
Fix group algorithm example

gmlueck:
One of our developers discovered a bug in the example showing how to
use the group algorithm "reduce" function:

* There is no such function `nd_item::get_linear_id()`.  We meant to
  use `get_global_linear_id()` instead.

* The partial sum computed by the example should be 120
  (= 0 + 1 + ... + 15) not 136 (= 1 + 2 + ... + 16).

Also fix the formatting, so there is no line wrap in the PDF render.

### DIFF
--- a/adoc/code/algorithms.cpp
+++ b/adoc/code/algorithms.cpp
@@ -17,19 +17,22 @@ myQueue.submit([&](handler & cgh) {
   cgh.parallel_for(nd_range<1>(range<1>(16), range<1>(16)),
     [=] (nd_item<1> it) {
 
-      // Apply a group algorithm to any number of values, described by an iterator range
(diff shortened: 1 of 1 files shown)


gmlueck:
@Pennycook: Can you please check that this is correct?

Pennycook:
LGTM.

I think I can shed some light on the 136.  For correctness tests I think it's dangerous (in general) to include 0 in the inputs, since there's no way to be sure that the reduction actually included the contribution from work-item 0.  I'm mentioning this only in case this test (or something similar) ends up making its way into the CTS -- including a 0 in this simple example is fine, but otherwise having `std::iota` start from 1 might be a good idea.

keryell (approved):
You can thank one of your developers! :-)

ProGTX (approved):

PR #168, opened 2021-07-20
Invert order of src/dest parameters for USM copy

gmlueck:
We noticed that the USM `copy()` functions had a parameter order of
`(dest, src)`, which is opposite of the `copy()` functions for
accessors.  This seemed very confusing, so we decided to change the
order to match.  Since these USM functions are new in SYCL 2020 and
because no major implementation has added them yet, we think it is
still safe to make this change.

This closes internal issue #557

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3384,14 +3384,14 @@ a@
 [source]
 ----
 template <typename T>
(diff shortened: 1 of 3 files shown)


mkinsner:
LGTM

ProGTX (approved):

keryell (approved):
Quite more coherent.

PR #170, opened 2021-07-22
Clarify equality of "kernel_id" objects

gmlueck:
Clarify that two `kernel_id` objects compare equal if and only if they
refer to the same kernel.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -14373,6 +14373,12 @@ functions and common member functions that are listed in
 <<sec:reference-semantics>> in <<table.specialmembers.common.reference>> and
 <<table.hiddenfriends.common.reference>>, respectively.
 
+As with all SYCL objects that have the common reference semantics, kernel
+identifiers are equality comparable.  Two [code]#kernel_id# objects that refer
+to the same application kernel or to the same device built-in kernel are
+guaranteed to compare equal.  Two [code]#kernel_id# objects that refer to
+different kernels are guaranteed to compare unequal
+
 There is no public default constructor for this class.
 
 [source,,linenums]
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -14374,10 +14374,9 @@ functions and common member functions that are listed in
 <<table.hiddenfriends.common.reference>>, respectively.
 
 As with all SYCL objects that have the common reference semantics, kernel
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -14373,6 +14373,12 @@ functions and common member functions that are listed in
 <<sec:reference-semantics>> in <<table.specialmembers.common.reference>> and
 <<table.hiddenfriends.common.reference>>, respectively.
 
+As with all SYCL objects that have the common reference semantics, kernel
+identifiers are equality comparable.  Two [code]#kernel_id# objects that refer
+to the same application kernel or to the same device built-in kernel are
+guaranteed to compare equal.  Two [code]#kernel_id# objects that refer to
+different kernels are guaranteed to compare unequal
```
keryell, line 14380:
```suggestion
different kernels are guaranteed to compare unequal.
```

keryell: Perhaps we could condense this wording by using a mathematical `if and only if` instead?

keryell (changes requested):
Missing ".".

adoc/chapters/programming_interface.adoc
gmlueck: Good idea.  Wording simplified in 3086fb3d52186a319cbc998ab0bb3f639fc72955.

adoc/chapters/programming_interface.adoc
```diff
@@ -14373,6 +14373,11 @@ functions and common member functions that are listed in
 <<sec:reference-semantics>> in <<table.specialmembers.common.reference>> and
 <<table.hiddenfriends.common.reference>>, respectively.
 
+As with all SYCL objects that have the common reference semantics, kernel
+identifiers are equality comparable.  Two [code]#kernel_id# objects compare
+equal if an only if they refer to the same application kernel or to the same
```
keryell, line 14378:
```suggestion
equal if and only if they refer to the same application kernel or to the same
```

keryell (changes requested):

adoc/chapters/programming_interface.adoc
gmlueck: Fixed in 67abc57aed6aa818c374e044dbb387f6625c9502.

keryell (approved):

PR #172, opened 2021-07-23
Clarify reduction identity behavior

Pennycook:
This PR removes constraints from the reduction interface, allowing users to specify an identity value even in situations where the implementation knows what the identity value should be.  The expected behavior when the wrong identity value is supplied is clarified.

Addresses internal Khronos issue https://gitlab.khronos.org/sycl/Specification/-/issues/525.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12302,14 +12302,12 @@ variable is accessed directly during the lifetime of a reduction (e.g. via an
 [code]#accessor# or USM pointer), the behavior is undefined.
 
 Certain implementations of reductions can be made more efficient if the
-identity value associated with the reduction operator is known.  For standard
-binary operators (e.g. [code]#plus# on arithmetic types), an implementation
-must determine the correct identity value automatically in order to avoid
-performance penalties.  For user-defined reduction operators, an implementation
-should issue a compile-time warning if the user does not provide an identity
-value and this is known to negatively impact performance.
+reduction operator has an identity value and this value is known to the
+implementation. For standard binary operators (e.g. [code]#plus# on arithmetic
+types), an implementation can determine the correct identity value
+automatically in order to avoid performance penalties.
 
-If an implementation can identify the identity value for a given combination of
+If an implementation can identify an identity value for a given combination of
 accumulator type and function object type, the value is defined as a member of
 the [code]#known_identity# trait class.  Whether this member value exists can
 be tested using the [code]#has_known_identity# trait class.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12349,9 +12349,10 @@ implementation-defined object of unspecified type, which is interpreted by
 [code]#parallel_for# to construct an appropriate [code]#reducer#
 type as detailed in <<reducer-class>>.
 
-If an identity value is supplied to a reduction, an implementation may use that
-value to initialize an unspecified number of temporary variables inside of any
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -12417,9 +12428,7 @@ const property_list &propList = {})
       of the variable(s) described by [code]#vars# using the combination
```
gmlueck, line 12428: GitHub won't let me add a comment to the line I really want to comment upon, my comment is really on this line:

> Available only when [code]#has_known_identity<BinaryOperation, BufferT::value_type>::value# is [code]#false#.

I think you want to remove this line, and the lines similar below.  These descriptions correspond to the comments you removed in the synopsis.

adoc/chapters/programming_interface.adoc
```diff
@@ -12349,6 +12349,17 @@ implementation-defined object of unspecified type, which is interpreted by
 [code]#parallel_for# to construct an appropriate [code]#reducer#
 type as detailed in <<reducer-class>>.
 
+If an identity value is supplied to a reduction, an implementation may use that
```
tomdeakin, line 12352:
```suggestion
If an identity value is supplied to a reduction, an implementation will use that
```

Think we need to say the identity you give won’t just be ignored.

adoc/chapters/programming_interface.adoc
```diff
@@ -12349,6 +12349,17 @@ implementation-defined object of unspecified type, which is interpreted by
...
+[NOTE]
+====
+Since the number of temporary variables is unspecified, supplying an identity
+value different to the identity value associated with the reduction operator
+may lead to unexpected results.
+====
+
 The initial value of the reduction variable is included
 in the reduction operation, unless the [code]#property::reduction::initialize_to_identity#
 property was specified when the [code]#reduction# interface was invoked.
```
tomdeakin, line 12365: Do we need a comment to say the identity will be the known one, unless you specify one and then it will be the one you specified?

adoc/chapters/programming_interface.adoc
Pennycook: Good catch, thanks.  Fixed in a9faf3e98b49cee456a92890093f6eeff8fc0121.

adoc/chapters/programming_interface.adoc
Pennycook: I've had a go at addressing this in b7fb2d6f3e59940078f0e5b8a7c94c08facee546.  I don't think it's as simple as replacing "may" with "will", because an implementation isn't required to use temporary accumulator variables at all -- but if an implementation does use temporary variables, they should be initialized to the identity.

adoc/chapters/programming_interface.adoc
Pennycook: I agree this needed to be clarified.  Rather than add something here, I've tried to clarify the expected behavior in the description of `initialize_to_identity` property itself.  How does 41cb15937975d2ccae97bcced63eb679c76ad054 look?

adoc/chapters/programming_interface.adoc
tomdeakin: That change LGTM.

adoc/chapters/programming_interface.adoc
tomdeakin: Yep, that looks great.

tomdeakin (approved):

keryell (approved):

keryell (approved):

fraggamuffin:
LGTM

PR #183, opened 2021-08-13
Add "std::byte" support to stream

gmlueck:
Since we encourage people to migrate from the old `sycl::byte` to
`std::byte`, I presume we want to support `std::byte` in the `stream`
class.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -18621,7 +18621,7 @@ include::{header_dir}/stream.h[lines=4..-1]
 a@
 [source]
 ----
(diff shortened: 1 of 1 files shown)


ProGTX:
Interesting, are bytes streamable? The example in https://en.cppreference.com/w/cpp/types/byte suggests you need to convert it to an integer first.

keryell:
#### 2021/09/02 SYCL Khronos committee meeting
Waiting for new wording from @gmlueck.
This PR is not ready for merging yet.

fraggamuffin:
awaiting draft for NEXT or KHR

TApplencourt:
@nliber  Std and sycl (this one seem to be a good candidate)

PR #185, opened 2021-08-18
Remove type restriction on "handler::fill()"

gmlueck:
The overload of `handler::fill()` that takes an accessor used to be
limited to only scalar and SYCL vector types, but it's not clear why
we need this restriction.  The underlying type of a buffer accessor
can be any "device copyable" type, and it seems like the fill operation
can be easily implemented for any of these types: the implementation
just needs to do a byte-wise copy.

There is no need to state that "T must be a device copyable type" in
this method (or any of the other explicit copy methods that take an
accessor argument) because all these operations are restricted to
buffer accessors, and the underlying type of a buffer accessor is
always device copyable.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13578,7 +13578,6 @@ void fill(accessor<T, dim, mode, tgt, isPlaceholder> dest,
 ----
 a@ Replicates the value of [code]#src# into the
 memory object accessed by [code]#dest#.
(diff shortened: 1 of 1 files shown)


ProGTX:
The restriction was originally introduced because of OpenCL limitations: https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/clEnqueueFillBuffer.html . I suppose it's possible to extend it to other types, but you might need some sort of emulation.

Issue #107, opened 2020-10-12
Allow 0-sized buffers


mfbalin:
When the user writes code, let's say the user has 100s of buffers and one of them has to have a 0 size. Since 0-sized buffers are not allowed in SYCL, existing SYCL implementations just throw an error. I don't think this is good at all. What is the harm of allowing 0-sized buffers?

ProGTX:
There has been some discussion about this. The first problem is that zeros are not allowed in a `range`. Codeplay proposed a default constructible buffer where the user doesn't have to specify a range: https://github.com/codeplaysoftware/standards-proposals/pull/97/files. However, the decision was made that since SYCL 2020 provisional targets C++17, a user can just use `std::optional<buffer>` instead.

fraggamuffin:
internal 542

fraggamuffin:
https://github.com/KhronosGroup/SYCL-Docs/issues/107

tomdeakin:
This was addressed in #192, so I think this issue can now be closed.

zjin-lcf:
Thanks.

PR #192, opened 2021-09-20
Clarify behaviour for range of zero

tomdeakin:
This PR will resolve internal Issue 542.

This clarifies the situation when a `range` or `nd_range` of zero is specified with a `parallel_for`.
In short, the kernel *function* does not execute, but all other aspects of kernel execution for dependencies are satisfied (i.e. accessor and event dependencies are still required to be met before scheduling, and queue submissions depending on this kernel submission will be satisfied). This follows the behaviour of OpenCL 2.1.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -581,6 +581,11 @@ work-item executes the same code but the specific execution pathway through the
 code and the data operated upon can vary by using the work-item global id to
 specialize the computation.
 
+An index space of size zero is allowed.  All aspects of kernel execution proceed
+as normal with the exception that the kernel function itself is not executed.
+Note this means the command queue will still schedule this kernel after satisfying
+the requirements and this satisfies requirements of any dependent enqueued kernels.
+
 ==== Basic kernels
 
 SYCL allows a simple execution model in which a kernel is invoked over an
@@ -1204,8 +1209,10 @@ work-group in <<local memory,local>> or <<global memory>> and to synchronize bet
 work-items in the same work-group by calling the
 [code]#group_barrier# function. All work-groups in a given
 [code]#parallel_for# will be the same size, and the global size
-defined in the nd-range must be a multiple of the work-group size in
-each dimension.
+defined in the nd-range must either be a multiple of the work-group size in
+each dimension, or the global size must be zero. In the case the global size
+is zero, the kernel function is not executed, the local size is ignored, and
+any dependencies are satisfied.
 
 Work-groups may be further subdivided into <<sub-group>>s.
 The size and number of sub-groups is implementation-defined and may differ for
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12939,8 +12939,8 @@ template <typename KernelName, int dimensions, typename... Rest>
       the callable.
 
 Throws an [code]#exception# with the [code]#errc::nd_range# error code if the
-global size defined in the associated [code]#executionRange# is not evenly
-divisible by the local size in each dimension.
+global size defined in the associated [code]#executionRange# defines a non-zero
+index space which is not evenly divisible by the local size in each dimension.
 
 a@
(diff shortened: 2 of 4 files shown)


gmlueck:
I like this clarification.  I think we also discussed that zero-sized buffers should be allowed.  Offhand, I don't see anything in the spec that disallows zero-sized buffers, though.  Do we think any further clarification is required for that?

tomdeakin:
I didn't address that deliberately in the first pass, as I had a couple of things that might need clarifying:

 * I'd expect a zero-sized buffer should behave like a `malloc(sizeof(T)*0)`, so therefore do we want to add default `buffer()` constructor? If so, what dimension should it be; equiv. can we have zero sized 1D, 2D and 3D buffers? I.e. are all of `buffer<dataT, 1, AllocatorT>`, ..., `buffer<dataT, 3, AllocatorT>` allowed?
 * Can see an argument for not adding a default constructor too - it avoids that question, and users get a compiler error - you have to opt-in for zero.
 * Is it allowed to make an accessor of a zero-range buffer? If so it should influence the dependency graph. Is this really desirable, as this could be abused to replicate pseudo `depends_on(event e)` functionality.
 * But what happens if you try to access it - runtime error? That's different to a zero-range which safely does nothing.

tomdeakin:
Thanks for the feedback - I agree with all of that. It seems the simplest too.
Leave it with me to add further clarifications to this PR for buffers.

tomdeakin:
This will also close #107

adoc/chapters/architecture.adoc
```diff
@@ -1204,8 +1209,10 @@ work-group in <<local memory,local>> or <<global memory>> and to synchronize bet
 work-items in the same work-group by calling the
 [code]#group_barrier# function. All work-groups in a given
 [code]#parallel_for# will be the same size, and the global size
-defined in the nd-range must be a multiple of the work-group size in
-each dimension.
+defined in the nd-range must either be a multiple of the work-group size in
+each dimension, or the global size must be zero. In the case the global size
```
gmlueck, line 1213:
```suggestion
each dimension, or the global size must be zero. When the global size
```

I think this is more concise.  If you prefer your original wording, I think there was some word missing.  Maybe you meant "In the case **where** the global size ...".

adoc/chapters/programming_interface.adoc
```diff
@@ -6287,6 +6291,12 @@ must take care to manually add the offset before dereferencing the pointer
 because accessing an element that is outside of the accessor's range results
 in undefined behavior.
 
+[NOTE]
+====
+There is no change in behavior for ranged accessors with a range of zero.
+It still creates a requisite for the entire underlying buffer, and 
+and attempt to access an element produces undefined behaviour.
```
gmlueck, line 6298:
```suggestion
an attempt to access an element produces undefined behaviour.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -6507,6 +6517,15 @@ the range starts at the beginning of the [code]#buffer#.  The optional
...
 [code]#accessRange# exceeds the range of [code]#bufferRef# in any dimension.
 
+If the range is zero sized, this constructs an empty accessor which fulfills the
+following post-conditions:
+--
+  * [code]#(empty() == true)#
+  * All size queries return [code]#0#.
+  * The only iterator that can be obtained is [code]#nullptr#.
+  * Trying to access the underlying memory is undefined behavior.
+--
```
gmlueck, line 6527: Do we need similar wording for all the **non**-ranged accessors for the case when the buffer size is `0`?  I presume all these statements are true for that case also, right?

Alternatively, maybe we don't need to say this at all.  I think the behavior is clear even without these statements.

tomdeakin: If the intent is clear, then I don't think we need to repeat this many, many time in the spec. I've removed this in 8fd20e133060b99cda654c45924b2253fe615036

PR #194, opened 2021-09-27
Add some shortcuts for querying kernels

gmlueck:
Add some shortcuts to `kernel_bundle` for the common case when the
application queries for a single kernel whose name is statically
known.

This addresses a review comment in KhronosGroup/SYCL-CTS#179.


keryell (approved):
Great!

ProGTX (approved):

PR #206, opened 2021-12-02
Clarify valid arguments for ND-range kernels

Pennycook:
Prior to this change, the basic parallel_for (accepting a range)
permitted generic arguments and types implicitly converted from
sycl::item, while the ND-range form of parallel_for permitted only
sycl::nd_item.

This change brings the ND-range form of parallel_for in line with the
basic parallel_for, permitting generic arguments and types implicitly
converted from sycl::nd_item.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13131,6 +13131,12 @@ implicitly converted from SYCL [code]#item#, representing the currently
 executing work-item within the range specified by the [code]#range#
 parameter.
 
+[NOTE]
+====
+Case 3) above includes user-defined types that can be constructed from
+[code]#item#, enabling users to layer their own abstractions on top of SYCL.
+====
+
 The execution of the kernel function is the same whether the parameter to
 the SYCL kernel function is a SYCL [code]#id# or a SYCL
 [code]#item#. What differs is the functionality that is available to
@@ -13239,6 +13245,12 @@ within the range specified by the [code]#nd_range# parameter. The
 its position in the range available, and provides access to functions
 enabling the use of a <<work-group-barrier>> to synchronize between the
 <<work-item>>s in the <<work-group>>.
+
+[NOTE]
+====
+Case 3) above includes user-defined types that can be constructed from
+[code]#nd_item#, enabling users to layer their own abstractions on top of SYCL.
+====
 
 The following example shows how sixty-four work-items may be launched
 in a three-dimensional grid with four in each dimension, and divided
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

tomdeakin:
What are the types implicitly converted from a `sycl::nd_item`? I had a quick scan through the constructors, but couldn't see anything.

Pennycook:
> What are the types implicitly converted from a `sycl::nd_item`? I had a quick scan through the constructors, but couldn't see anything.

I think you're right that there aren't any in the specification, but users can define their own classes that are implicitly convertible from `sycl::nd_item`.

If we look at the `sycl::item` functionality that existed before, you can see that the `item` class itself doesn't define a conversion to `sycl::id`; rather, the `id` class defines a conversion from `sycl::item`.  The specification already allows for the possibility of the conversion operator being defined outside of the class itself, and doesn't restrict itself to classes defined by the specification.

I think this may be an important use-case for embedded DSLs.  It allows developers to write things like:
```c++
q.parallel_for(N, [=](my_library::my_index_class idx) {
  // kernel
});
```
...and have the conversion from a SYCL index class to a user-defined index class happen automatically.

tomdeakin:
Cool. Do we want a non-normative note to that effect?

adoc/chapters/programming_interface.adoc
```diff
@@ -13229,10 +13229,16 @@ exposed in SYCL through [code]#+parallel_for (nd_range,...)+# and the
...
 parallel_for expects an [code]#nd_range#, specifying both local and
 global ranges, defining the global number of work-items and the number in
-each cooperating work-group. The resulting function object is passed an
-[code]#nd_item# instance making all the information available, as well
-as <<work-group-barrier>> to synchronize between the <<work-item>>s in
-the <<work-group>>.
+each cooperating work-group. The function object that represents the SYCL
+kernel function must take one of:
+1) a single SYCL [code]#nd_item# parameter, 2) a single generic parameter
+([code]#template# parameter or [code]#auto#), 3) any other type converted
```
gmlueck, line 13235: Should we state here someplace that the kernel argument's type will be `nd_item` if the formal parameter is declared `auto`?

Pennycook: Added in 993559f.

gmlueck (approved):

Pennycook:
> Cool. Do we want a non-normative note to that effect?

I've added notes in 00ed854 for both the `item` and `nd_item` cases.

Note that Section 4.9.4.2.2. "`parallel_for` invoke" is now very long (it was quite long before) and some of the information (e.g. about interpretation of `auto` arguments) is duplicated in these commits because basic and `nd_range` behaviors are discussed separately.  I think we could simplify this section significantly with some shuffling of content, but since that's a larger editorial change I'd like to keep it separate from these clarifications.

tomdeakin (approved):

keryell (approved):

fraggamuffin:
LGTM

PR #208, opened 2022-01-04
Fixed return type of group_broadcast(Group g, T x)

oleksandr-pavlyk:
Since `group_broadcast(g, x)` is equivalent to `group_broadcast(g, x, 0)`
return types of both these functions should be the type of `x`, i.e. `T`.

Spec stipulated `group_broadcast(g, x)` to have `bool` return type instead.

### DIFF
--- a/adoc/headers/groups/broadcast.h
+++ b/adoc/headers/groups/broadcast.h
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 template <typename Group, typename T>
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
@keryell Please review.

ProGTX (approved):

oleksandr-pavlyk:
@keryell Ping

keryell (approved):
Thanks!

fraggamuffin:
LGTM

PR #211, opened 2022-01-07
Fix definition of fp_config::correctly_rounded_divide_sqrt

npmiller:
The definition that was used before was taken from the OpenCL 1.2
specification, see section 4.2, table 4.3:

> CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT – divide and sqrt are correctly
> rounded as defined by the IEEE754 specification.

However this definition is misleading because what is really advertised
by this property is that the correctly rounded divide and sqrt are
supported, not that they are correctly rounded by default on the
given device.

In the OpenCL 1.2 specification this is made clear in section 5.6.4.2
which defines the build program flag
`-cl-fp32-correctly-rounded-divide-sqrt`.

Where it states the following:

> If this build option is not specified, the minimum numerical accuracy of
> single precision floating-point divide and sqrt are as defined in section 7.4
> of the OpenCL specification.

And:

> This build option can only be specified if the
> CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT is set in CL_DEVICE_SINGLE_FP_CONFIG (as
> defined in table 4.3) for devices that the program is being build.
> clBuildProgram or clCompileProgram will fail to compile the program for a
> device if the -cl-fp32-correctly-rounded-divide-sqrt option is specified and
> CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT is not set for the device.

But since SYCL doesn't specify build options, this entire section doesn't
appear in the SYCL specification and we're left with just the misleading
property definition that makes it sound like the correctly rounded divide and
sqrt will be used by default.

And so this patch alters the wording to only say that the correctly rounded
variants are supported, and not suggest that they are always used.

This new definition is still pretty vague, but it should allow implementations to support this device property as intended. And I'm unsure if further clarifications are feasible since the toggling mechanism (compiler flag) between regular precision or correctly rounded would likely remain implementation defined.

### DIFF
--- a/adoc/chapters/what_changed.adoc
+++ b/adoc/chapters/what_changed.adoc
@@ -4,6 +4,8 @@
 [[cha:what-changed-from]]
 = What has changed from previous versions
 
+The device query [code]#info::fp_config::correctly_rounded_divide_sqrt# has
+been deprecated.
 
 [[sec:what-changed-between]]
 == What has changed from SYCL 1.2.1 to SYCL 2020
--- a/adoc/chapters/what_changed.adoc
+++ b/adoc/chapters/what_changed.adoc
@@ -4,9 +4,6 @@
 [[cha:what-changed-from]]
 = What has changed from previous versions
 
-The device query [code]#info::fp_config::correctly_rounded_divide_sqrt# has
-been deprecated.
-
 [[sec:what-changed-between]]
 == What has changed from SYCL 1.2.1 to SYCL 2020
 
@@ -261,6 +258,9 @@ functions used as <<device-selector,device selectors>> described in
 A new device selector utility has been added to <<sec:device-selector>>,
 the [code]#aspect_selector#, which returns a selector object
 that only selects devices that have all the requested aspects.
+
+The device query [code]#info::fp_config::correctly_rounded_divide_sqrt# has
+been deprecated.
 
 A new reduction library consisting of the [code]#reduction# function and
 [code]#reducer# class was introduced to simplify the expression of variables
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2044,8 +2044,8 @@ info::device::half_fp_config
     negative infinity rounding modes are supported.
   * [code]#info::fp_config::fma:# IEEE754-2008 fused multiply add is
     supported.
-  * [code]#info::fp_config::correctly_rounded_divide_sqrt:# correctly rounded
-    divide and sqrt as defined by the IEEE754 specification are supported.
+  * [code]#info::fp_config::correctly_rounded_divide_sqrt:# divide and
+    sqrt are correctly rounded as defined by the IEEE754 specification.
   * [code]#info::fp_config::soft_float:# basic floating-point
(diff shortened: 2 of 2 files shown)


gmlueck:
With this new definition, it seems like the query isn't very useful.  An application can learn that there is some way to get correctly rounded sqrt and divide, but we don't tell the application how to get that behavior, and we don't even provide a way to specify the compilation flag if the programmer somehow knew it was necessary.

Here are a couple of alternatives that come to mind:

* If we think this is an obscure query that few people need, we could deprecate the query in SYCL 2020 and remove it in the next release.  Applications that do need the query can use OpenCL native APIs and the OpenCL backend interop.  Vendors could also provide a SYCL compiler command line option or environment variable that controls whether the rounding mode for sqrt and divide are correct.

* If we think that application developers will need some core SYCL AP to simply query whether sqrt and divide are correctly rounded, we could keep the existing wording (i.e. before this PR).  In that case, OpenCL backends would return `false` unless they supported the correctly rounded versions by default.  Applications could still use the OpenCL backend interop functions to get the correctly rounded behavior if they want.

npmiller:
You're absolutely right that the new query isn't very useful, at least without additional vendor specific compilation flag or something like that. And so I think your first suggested approach of just deprecating, and then removing the property, is probably the correct way to handle this.

As for your second suggestion, I'm not entirely sure what would be the use of a property essentially just reporting `sqrt` precision. The precision requirements are an upper bound, and many built-ins can be more precise than required, so if a device happens to use correctly rounded `sqrt` by default I'm not sure why that would warrant a specific property to report it, compared to any other math built-in.

npmiller:
I've updated the PR to mark the property as deprecated instead.

fraggamuffin:
Add to What's Changed. Approved workflow.

npmiller:
Added a mention of this in "What's changed".

I wasn't completely sure where to put it in the file so I simply placed it at the top outside of a specific section, I'm assuming these will be gathered into a new section for the next spec release, or should it just go in the SYCL 1.2.1 to SYCL 2020 section?

adoc/chapters/what_changed.adoc
```diff
@@ -4,6 +4,8 @@
 [[cha:what-changed-from]]
 = What has changed from previous versions
 
+The device query [code]#info::fp_config::correctly_rounded_divide_sqrt# has
+been deprecated.
```
gmlueck, line 8: I assume it is our intention that this change is a "fix" to SYCL 2020, right?  Or, is it our intention to make this change in SYCL-Next?

If it is a fix to SYCL 2020, then this paragraph should go in the section "What has changed from SYCL 1.2.1 to SYCL 2020".  In that case, move this paragraph to the end of that section.

npmiller: Oh I see, that makes sense, I moved the paragraph in the SYCL 2020 section next to other device aspects related changes.

gmlueck (approved):

keryell (approved):

fraggamuffin:
LGTM; npmiller please add yourself as contributor, congratulation on your first contribution.

PR #216, opened 2022-02-02
Add descriptions of the queue shortcut functions

Pennycook:
Clarify the relationship between each shortcut function in the queue
class and equivalent command-group functionality in the handler class.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3257,9 +3257,9 @@ include::{code_dir}/queueShortcuts.cpp[lines=4..-1]
 
 [[table.queue.shortcuts]]
 .Queue shortcut functions
-[width="100%",options="header",separator="@",cols="80%,20%"]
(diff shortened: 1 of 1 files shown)


Pennycook:
For some brief background on motivation:

- The table describing the queue shortcuts is quite large, and it's easy to land in the middle of it (e.g. Ctrl + F for "memcpy").  Knowing that these shortcuts create implicit command-group handlers is important for understanding how they operate; making this relationship more explicit reduces the probability that a reader will misunderstand how the shortcuts interact with other submissions to the same queue.

- Some shortcuts map directly to a function with the same name declared in the `handler` class, but many take different arguments (e.g. dependent events), or map to multiple function calls (e.g. additional `depends_on` and/or `requires` calls).  Although this information is in the paragraph above the table, it's easy to miss it.

tomdeakin (approved):
LGTM.

gmlueck (approved):

ProGTX (approved):

fraggamuffin:
LGTM

PR #219, opened 2022-02-07
Fix attribute position in example

gmlueck:
Fix the position of the C++ attribute in this example.  We changed the
attribute position in #136 (34732eea75a42865ea278f376a7d40568018e86e)
to adopt the standard location for attributes that appertain to a
function.  We should have changed this example in that PR also, but it
was missed.


tomdeakin (approved):

keryell (approved):
Thanks!

PR #220, opened 2022-02-07
Clarify relationship between scans and identities

Pennycook:
The descriptions of the exclusive_scan and inclusive_scan group
algorithms previously made reference to an identity value without
specifying how that value was determined.

For exclusive_scan, the identity should be the same as the one
identified by sycl::known_identity.

For inclusive_scan, the reference to an identity value was a copy-paste
error; unlike an exclusive scan, an inclusive scan is well-defined
without an initial value.


tomdeakin (approved):

gmlueck (approved):

keryell (approved):
Thanks!

PR #221, opened 2022-02-08
Added stream to the list of classes with common reference semantics in 4.5.2.

pkeir:
From 4.16. Stream class:

"The SYCL stream class provides the common reference semantics (see Section 4.5.2)."

But in Section 4.5.2 `stream` is not listed:


gmlueck (approved):

keryell (approved):
Good catch!

PR #224, opened 2022-02-17
Fix prototype of OCL make_event

gmlueck:
Fix the prototype of the OpenCL backend interop function `make_event()`.
The type of the first parameter must be
`const backend_input_t<backend::opencl, event> &`, but this type alias
is defined as `std::vector<cl_event>` in the table "List of native
types per SYCL object in the OpenCL backend".  We decided that the
table is correct, so change the prototype to match.

Closes internal issue 588.

### DIFF
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -491,10 +491,13 @@ context make_context(
 a@
 [source]
 ----
-event make_event(const cl_event &clEvent,
+event make_event(
+    const std::vector<cl_event> &clEvents,
     const context &syclContext)
 ----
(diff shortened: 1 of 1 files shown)


tomdeakin:
@gmlueck - do we need to do anything with this header too: https://github.com/KhronosGroup/SYCL-Docs/blob/06b2be12a81c1b89b7b553d309eef09b92c8e23a/adoc/headers/interop/templateFunctionMakeX.h#L22

tomdeakin (approved):

ProGTX (approved):

fraggamuffin:
LGTM

keryell (approved):
Good catch!
Thanks.

PR #225, opened 2022-02-21
Rephrase the confusing 'buffer scope' wording

u235axe:
This PR changes the following phrase:

"A SYCL application can use memory managed by a SYCL buffer within the buffer scope by using a host accessor as defined in <<subsec:accessors>>."

to:

"A SYCL application can access the contents of the memory managed by a SYCL buffer
by using a host accessor as defined in <<subsec:accessors>>.

The intent is to remove the confusing wording of 'buffer scope'.


tomdeakin (approved):

keryell (approved):
Thanks!
I am unsure about whether `content` or `contents` is more qualified here, as I am not a native English speaker.

gmlueck (approved):

gmlueck:
> Thanks! I am unsure about whether `content` or `contents` is more qualified here, as I am not a native English speaker.

I think "contents" is correct because the bytes in the memory range are "countable" (i.e. there are a finite number of them):

ProGTX (approved):

PR #226, opened 2022-02-21
Rephrase asynchronous error handling in case of secondary queue.

u235axe:
This PR fixes the following phrase and the paragraph around it:

"If there is no async_handler given on any of the queues, then no asynchronous error
reporting is done and no exceptions are thrown."

to be more in line with the intent of the previous paragraph on asynch handler fallback logic to use the context's or missing that, the default handler.

Please double check if the current phrasing is adequate and up to date with any other intentions, or further tweaks are required.


keryell (approved):
That makes sense.

gmlueck (approved):

ProGTX (approved):

Issue #222, opened 2022-02-09
Clarify what the device version property should report


npmiller:
The device property `info::device::version` is defined as:

> Returns the SYCL version as a std::string in the form: <major_version>.<minor_version>.

In section [4.6.4.2](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors), Table 25.

But the current SYCL version is `2020` so what should this property return?

In addition this property is also mentioned in the OpenCL backend spec with the following, confusing paragraph:

> The device version is an indication of the device’s capabilities, as represented by the device information returned by the sycl::device::get_info() member function. Examples of attributes associated with the device version are resource limits and information about functionality beyond the requirements in the [core SYCL specification](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#core-spec). The version returned corresponds to the highest version of the OpenCL specification for which the device is conformant, but is not higher than the version of the device’s platform which bounds the overall capabilities of the runtime operating the device.

In section [C.5.1](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_platform_mixed_version_support_2).

This paragraph seems to suggest that the device version informs on hardware capabilities, but then turns around to say it only represent the highest OpenCL version supported.

And this also conflicts with the definition in section [4.6.4.2](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors) that says it should be the SYCL version.

In summary it would be good to clarify:
* What is the meaning of value reported by `info::device::version`?
* What is the format of that value?
* Can a backend change the meaning of that value and report something different?

fraggamuffin:
definition is now year based +revision
Device or device driver version? Should platform version be also used?
use along with backend info descriptor query
Nicolas to draft based on today's discussion
not SYCL version anymore, but platform driver version is BE dependent

keryell:
I think this versioning was following the OpenCL model but this has no longer a meaning in the general case of SYCL 2020 with various back-ends.

PR #231, opened 2022-02-24
Update device and platform version descriptions

npmiller:
This PR is a draft in attempt to resolve https://github.com/KhronosGroup/SYCL-Docs/issues/222

Any further feedback on these is appreciated.

This patch updates:
* `device::version` description to use the new SYCL versioning scheme.
* Update wording of `platform::version`, a platform has multiple devices
  so referencing the device doesn't seem sensible.
* Remove mentions of OpenCL backend in descriptors table
* Update OpenCL backend doc to describe the meaning of the version
  numbers.

### DIFF
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -149,15 +149,17 @@ combination of the two.
 
 The SYCL system presents the user with a set of devices, grouped into some
 number of platforms.
-The device version is an indication of the device's
-capabilities, as represented by the device information returned by the
-[code]#sycl::device::get_info()# member function.  Examples of attributes
-associated with the device version are resource limits and information
-about functionality beyond the requirements in the <<core-spec>>.
-The version returned corresponds to the highest version of the OpenCL
-specification for which the device is conformant, but is not higher than
-the version of the device's platform which bounds the overall capabilities
-of the runtime operating the device.
+
+The version returned by [code]#sycl::platform::get_info()# and
+[code]#sycl::device::get_info()# for the attributes
+[code]#info::platform::driver_version# and [code]#info::device::driver_version#
+respectively, represents the OpenCL software driver version in the form:
+[code]#major_number.minor_number#.
+
+The version returned for the device corresponds to the highest version of the
+OpenCL specification for which the device is conformant, but is not higher than
+the version of the device's platform which bounds the overall capabilities of
+the runtime operating the device.
 
 
 === OpenCL memory model
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -156,6 +156,7 @@ described in the table below:
 |====
 | SYCL                                 | OpenCL
 | [code]#info::platform::version#      | [code]#CL_PLATFORM_VERSION#
+| [code]#info::device::version#        | [code]#CL_DEVICE_VERSION#
 | [code]#info::device::driver_version# | [code]#CL_DRIVER_VERSION#
 |====
 
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -157,7 +157,6 @@ described in the table below:
 | SYCL                                 | OpenCL
 | [code]#info::platform::version#      | [code]#CL_PLATFORM_VERSION#
 | [code]#info::device::version#        | [code]#CL_DEVICE_VERSION#
-| [code]#info::device::driver_version# | [code]#CL_DRIVER_VERSION#
 |====
 
 === OpenCL memory model
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -144,23 +144,20 @@ units (where each PE maintains its own program counter) or as some
 combination of the two.
 
 
-// From Architecture, Section 3.3.1 (Platform mixed version support)
-=== Platform mixed version support
+=== Backend specific information descriptors
 
-The SYCL system presents the user with a set of devices, grouped into some
-number of platforms.
+Some of the SYCL information descriptors are backend-defined. For the OpenCL
+backend these information descriptors map directly to OpenCL properties as
+described in the table below:
 
-The version returned by [code]#sycl::platform::get_info()# and
-[code]#sycl::device::get_info()# for the attributes
-[code]#info::platform::driver_version# and [code]#info::device::driver_version#
-respectively, represents the OpenCL software driver version in the form:
-[code]#major_number.minor_number#.
-
-The version returned for the device corresponds to the highest version of the
-OpenCL specification for which the device is conformant, but is not higher than
-the version of the device's platform which bounds the overall capabilities of
-the runtime operating the device.
-
+[[table.opencl.info]]
+.Mapping of SYCL information descriptors to OpenCL properties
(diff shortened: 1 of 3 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -2432,7 +2430,8 @@ info::device::version
 
     @ [.code]#std::string#
    a@ Returns the SYCL version as a [code]#std::string# in the form:
-      [code]#<major_version>.<minor_version>#.
+      [code]#<version>.<revision>#, with [code]#<version># being a year based
+      number, for example [code]#2020.4#.
```
gmlueck, line 2434: This query doesn't make sense to me.  The implementation already exposes the SYCL version via the `SYCL_LANGUAGE_VERSION` macro.  Since the entire implementation is conformant to that version, won't every device return the same information via `info::device::version`?

I'm also not sure we should return the minor version number here (or in `SYCL_LANGUAGE_VERSION`).  Do we really expect implementations to change each time we make a minor version release of the spec?  Will we add a conformance test to validate that the implementation returns the most up-to-date minor version?  Will vendors need to re-validate each time there is a new minor release of the spec?

Is it too late to mark this info descriptor as "deprecated"?  We still need to define the semantic even if it is deprecated.  We could either say that it returns "2020", or say something more generic like:

> Returns a vendor-defined string for the device.

fraggamuffin:
defined meaning to string is a concern

adoc/chapters/programming_interface.adoc
keryell: Perhaps it is a left-over of SYCL 2.2 when we had the idea of having SYCL 1.2 and 2.2 compatible devices in the same program.

npmiller: Updated the PR to mark `info::device::version` as deprecated and make it return just the year based SYCL version.

adoc/chapters/opencl_backend.adoc
```diff
@@ -149,15 +149,17 @@ combination of the two.
...
-The version returned corresponds to the highest version of the OpenCL
-specification for which the device is conformant, but is not higher than
-the version of the device's platform which bounds the overall capabilities
-of the runtime operating the device.
+
+The version returned by [code]#sycl::platform::get_info()# and
+[code]#sycl::device::get_info()# for the attributes
+[code]#info::platform::driver_version# and [code]#info::device::driver_version#
+respectively, represents the OpenCL software driver version in the form:
+[code]#major_number.minor_number#.
```
gmlueck, line 157: Is it the intent that this number corresponds to the OpenCL *specification* version?  For example, "1.2" would indicate that the backend conforms to OpenCL 1.2 and "3.0" would indicate that it conforms to OpenCL 3.0?

When I read this before, I though the intent was for each vendor to return some vendor-specific version number of the device driver.  This is what DPC++ does currently.  For example, we currently return this for one of our OpenCL devices:

`"2021.13.11.0.23_160000"`

adoc/chapters/programming_interface.adoc
gmlueck: > It would just be a little bit of a shame to "lose" a simple property like that, that could easily be used by backends to report more informative information on the device such as OpenCL device version, architecture, compute capabilities and such.

If you feel this way, then we should not add the wording you proposed in 2ecbb672466a08e267ab32c1ef0d0385119044dc.  Instead, we should say something like:

> Returns a vendor-defined string for the device

This gives vendor's the flexibility to return information like you suggest.

npmiller: Updated with what I suggested, although, note that I was thinking of backend-defined rather than vendor-defined, so what the value should be would still be defined by the SYCL spec but just on a per backend basis.

I feel like that gives us enough flexibility to report useful information for the specific backend types while still giving some consistency across vendors.

adoc/chapters/opencl_backend.adoc
npmiller: I've re-worked this paragraph into a table referring to the OpenCL properties which I think should be much clearer.

However I noticed that in the comments this section was referring to another one from the main spec: [Platform mixed version support](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_platform_mixed_version_support).

But I can't really make sense of what this section is trying to say or how this part of the OpenCL backend spec was addressing it, does it have something to do with the abandoned idea of mixing SYCL 1.2 and 2.2 devices as @keryell suggested?

adoc/chapters/programming_interface.adoc
```diff
@@ -2401,8 +2401,6 @@ info::device::driver_version
 
     @ [.code]#std::string#
    a@ Returns a backend-defined driver version as a [code]#std::string#.
-      If using the OpenCL backend, the returned value represents the
-      OpenCL software driver version in the form: major_number.minor_number.
```
gmlueck, line 2405: I see that the OpenCL backend interop spec says this corresponds to `CL_DRIVER_VERSION`.  OpenCL 3.0 defines that as:

> OpenCL software driver version string. Follows a vendor-specific format.

Since OpenCL doesn't mandate any specific format, should we just say this in the SYCL spec directly?  The SYCL definition would then be something like:

> A vendor-defined string describing the version of the underlying backend software driver.

gmlueck (commented):
I like the latest set of changes, but one more comment below.

adoc/chapters/programming_interface.adoc
npmiller: Oh it got updated in 3.0! That's a better definition for sure.

> A vendor-defined string describing the version of the underlying backend software driver.

I feel like this a bit in conflict with the idea of having it backend-defined, the way I see it we say it's backend-defined in the main spec, then in the backend spec we can define it further to vendor-defined if we want, for the OpenCL plugin saying it's `CL_DRIVER_VERSION` defers it to the OpenCL spec which ends up being vendor defined anyway.

But then for say a CUDA backend we could define it to the compute capability in the format `major.minor` for example rather than having it vendor defined.

But if we just say it's vendor defined in the main spec I feel like we can't really give more details on what it's supposed to be in the backend spec.

gmlueck: What is the motivation for specifying this query as backend-defined, though?  Most of our queries are not backend-defined.  If the intent of `driver_version` is to return the version string of the vendor's device driver, it's hard to see why a backend would mandate any specific format.  Wouldn't we want each vendor to have the freedom to decide it's own format?

npmiller: Oh right! I was still thinking about the `version` rather than the `driver_version`, since the `driver_version` was already backend specific before my patch.

But now that I think about it, I do think you're right that it makes more sense to have the `driver_version` simply be vendor specific, I'll change it as well.

npmiller: Updated the `driver_version` as suggested and removed the `CL_DRIVER_VERSION` part from the OpenCL backend spec.

adoc/chapters/opencl_backend.adoc
keryell: Yes, I think this is to suggest that while the top-level supported version has some SYCL versions but using some platform might supporting only an older SYCL version. But since we no longer have true feature inclusion, I am not sure it still makes sense.

keryell (approved):
Thanks.

PR #232, opened 2022-03-03
Clarify empty accessor

gmlueck:
Even before this change, we specified that the accessor default
constructor creates an empty accessor.  Previously, we said

> The only iterator that can be obtained is nullptr.

However, this makes no sense because iterators aren't (necessarily)
pointers.  I considered changing this to say that `begin() == end()`,
however, this is already implied because we say that `empty()` is
`true`.  I thought it was more important to say that `get_pointer()`
and `get_multi_ptr()` return unspecified values in this case.  I also
clarified the descriptions of those two functions for the empty
accessor case.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6501,7 +6501,8 @@ accessor()
 --
   * [code]#(empty() == true)#
   * All size queries return [code]#0#.
-  * The only iterator that can be obtained is [code]#nullptr#.
+  * The return values of [code]#get_pointer()# and [code]#get_multi_ptr()# are
+    unspecified.
   * Trying to access the underlying memory is undefined behavior.
 
 A default constructed accessor can be passed to a <<sycl-kernel-function>>
@@ -6803,7 +6804,8 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
 ----
    a@ Returns a [code]#multi_ptr# to the start of this accessor's underlying
       buffer, even if this is a <<ranged-accessor>> whose range does not start
-      at the beginning of the buffer.
+      at the beginning of the buffer.  The return value is unspecified if the
+      accessor is empty.
 
 This function may only be called from within a <<command>>.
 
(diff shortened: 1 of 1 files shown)


keryell (approved):
That is a good clarification.
I have just an English question between `The return value` and `The returned value`.

gmlueck:
> I have just an English question between The return value and The returned value.

The phrase "the return value" seems more natural to me, though I see we use both phrases currently in the SYCL spec.  According to [Text Ranch](https://textranch.com/33068/return-value/or/returned-value/) (which I have not heard of before), the phrase "the return value" is more common on the web.

adoc/chapters/programming_interface.adoc
```diff
@@ -6501,7 +6501,8 @@ accessor()
 --
   * [code]#(empty() == true)#
   * All size queries return [code]#0#.
-  * The only iterator that can be obtained is [code]#nullptr#.
+  * The return values of [code]#get_pointer()# and [code]#get_multi_ptr()# are
```
ProGTX, line 6504: Why unspecified instead of `nullptr`?

gmlueck: Remember, there are other ways to create an empty accessor.  For example, you could create a ranged accessor with an empty range.  I think we should make the same guarantee about `get_pointer()` for all empty accessors, regardless of how they were created.  Do want to guarantee that `get_pointer()` returns `nullptr` even when a ranged accessor is empty?  It wouldn't be too hard to implement this, but it seems like extra code in the implementation that doesn't really provide that much benefit.  Why is it important to guarantee that `get_pointer()` returns `nullptr` for an empty accessor?

Somewhat related is this description in [4.7.6.6. "Accessing elements of an accessor"](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_accessing_elements_of_an_accessor):

> Any attempt to access an element via an index that is outside of this range produces undefined behavior.

Since there is no valid index for an empty accessor, this means that any attempt to access an element of an empty accessor produces undefined behavior.  To me, that wording seemed consistent with saying that the value of `get_pointer()` is undefined for an empty accessor.

ProGTX: Sounds good to me, thanks for the clarification.

fraggamuffin:
LGTM

PR #238, opened 2022-03-22
Clarify SYCL_EXTERNAL

gmlueck:
Make a number of clarifications about the use of SYCL_EXTERNAL:

* Clarify the syntactic position of the macro.

* Clarify that it must be used on both the declaration and definition
  of the function.

* Clarify that it must be used on the first declaration of the function
  in the translation unit (similar to what we say about the C++
  attributes).

* Clarify the interaction with the attributes
  `[[sycl::reqd_work_group_size()]]` and
  `[[sycl::reqd_sub_group_size()]]`.

The spec did not previously state that a global variable that is used
in device code must be defined in the same translation unit, but I
think this was our intent.  I added that restriction.

Closes internal issue 444.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1837,15 +1837,6 @@ destructor for such a class on the device since the destructor must have no
 effect on the device.
 ====
 
-=== SYCL linker
-
-In SYCL, only offline linking is supported for SYCL programs and libraries,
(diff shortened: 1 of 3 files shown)


ProGTX (approved):

fraggamuffin:
LGTM

keryell (approved):
Good!

PR #239, opened 2022-03-24
Improve wording of host task capture restrictions

npmiller:
The original sentence was a little confusing at first glance because it
could be interpreted as "capturing with reference semantics" rather than
talking about the SYCL objects having reference semantics.

So this is very minor, but changing "with" to "that has" here clears up any
potential confusions.


gmlueck (approved):
Yes, that is better.  I agree that the old wording was confusing.

tomdeakin (approved):

fraggamuffin:
LGTM

keryell (approved):

PR #240, opened 2022-03-24
Clarify buffer and accessor lifetime

gmlueck:
Clarify that the application must not use an accessor after its
associated buffer / image is destroyed,  Also clarify the point at
which requisites are added for a command.  This happens after the
command group function object returns, not during the call to
`parallel_for` (etc.)

Closes internal issue 420

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6060,9 +6060,23 @@ to local memory on a <<device>>, and they define the *requirements* to memory
 objects which determine the scheduling of <<kernel,kernels>> (see
 <<sub.section.memmodel.app>>).
 
-A memory object requirement is created when an accessor is constructed, unless
-the accessor is a placeholder in which case the requirement is created when
-the accessor is bound to a <<command>> by calling [code]#handler::require()#.
+A memory object requirement is created by a <<command-group-function-object>>
+whenever a non-placeholder accessor is constructed during execution of that
+function and whenever a placeholder accessor is bound to a <<command>> during
+execution of that function by calling [code]#handler::require()#.  These
+requirements are created sometime after the <<command-group-function-object>>
+returns to the [code]#submit()# function and before the [code]#submit()#
+function itself returns.
+
+[NOTE]
+====
+Since the memory object requirements aren't created until after the command
+group function object returns, it is generally unwise to construct a buffer or
+image inside of command group scope.  The destructor for the buffer or image
+only blocks when there is a memory requirement, but variables defined in the
+command group function object go out of scope before those requirements are
+added.
+====
 
 There are several different {cpp} classes that implement accessors:
 
@@ -6437,6 +6451,15 @@ placeholder accessor is passed as an argument to a <<command>> without first
 being bound to a <<command-group>> with [code]#handler::require()#, the
 implementation throws a synchronous [code]#exception# with the
 [code]#errc::kernel_argument# error code when the <<command>> is submitted.
+
+An [code]#accessor# to a buffer is only a view over that buffer.  Therefore,
+it is the application's responsibility to ensure that the lifetime of the
+buffer is at least as long as the lifetime of the accessor.  Calling any
+member function on an accessor after its associated buffer has been destroyed
+results in undefined behavior.  In some cases the buffer destructor is defined
+to block until commands with requisites on that buffer have completed.  This
+results in well defined behavior so long as all invocations to member functions
+of accessors for that buffer occur before those commands complete.
 
 
 ===== Interface for buffer command accessors
@@ -7671,6 +7694,12 @@ The [code]#host_accessor# class supports the following access modes:
 [code]#access_mode::read#, [code]#access_mode::write# and
 [code]#access_mode::read_write#.
 
+A [code]#host_accessor# to a buffer is only a view over that buffer.
+Therefore, it is the application's responsibility to ensure that the lifetime
+of the buffer is at least as long as the lifetime of the host accessor.
+Calling any member function on a host accessor after its associated buffer has
+been destroyed results in undefined behavior.
+
 
 ===== Interface for buffer host accessors
 
@@ -8472,6 +8501,16 @@ template parameter as [code]#image_target::host_task# and then use the
 [code]#unsampled_image_accessor# from a <<sycl-kernel-function>> are ill
 formed.
 
+An unsampled image accessor is only a view over its associated image.
+Therefore, it is the application's responsibility to ensure that the lifetime
+of the image is at least as long as the lifetime of the accessor.  Calling any
+member function on an unsampled image accessor after its associated image has
+been destroyed results in undefined behavior.  In some cases the image
+destructor is defined to block until commands with requisites on that image
+have completed.  This results in well defined behavior so long as all
+invocations to member functions of accessors for that image occur before those
+commands complete.
+
 
 ===== Interface for unsampled image accessors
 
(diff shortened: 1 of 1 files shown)


tomdeakin (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -7671,6 +7694,12 @@ The [code]#host_accessor# class supports the following access modes:
 [code]#access_mode::read#, [code]#access_mode::write# and
 [code]#access_mode::read_write#.
 
+A [code]#host_accessor# to a buffer is only a view over that buffer.
+Therefore, it is the application's responsibility to ensure that the lifetime
+of the buffer is at least as long as the lifetime of the host accessor.
+Calling any member function on a host accessor after its associated buffer has
+been destroyed results in undefined behavior.
```
keryell, line 7701: I do not understand how it is an issue in a conforming implementation.
If you you have a live accessor, the buffer is still kept alive though for example an internal `shared_ptr`.

adoc/chapters/programming_interface.adoc
```diff
@@ -6060,9 +6060,23 @@ to local memory on a <<device>>, and they define the *requirements* to memory
...
+returns to the [code]#submit()# function and before the [code]#submit()#
+function itself returns.
+
+[NOTE]
+====
+Since the memory object requirements aren't created until after the command
+group function object returns, it is generally unwise to construct a buffer or
+image inside of command group scope.  The destructor for the buffer or image
+only blocks when there is a memory requirement, but variables defined in the
+command group function object go out of scope before those requirements are
```
keryell, line 6077: The problem I can see in my implementation is more like a deadlock.
But thinking more about it, I am unsure that not dealing with dependencies at accessor construction time instead of postponing after `submit` lambda execution can be a generic enough C++ implementation.

adoc/chapters/programming_interface.adoc
```diff
@@ -6438,6 +6452,15 @@ being bound to a <<command-group>> with [code]#handler::require()#, the
...
 [code]#errc::kernel_argument# error code when the <<command>> is submitted.
 
+An [code]#accessor# to a buffer is only a view over that buffer.  Therefore,
+it is the application's responsibility to ensure that the lifetime of the
+buffer is at least as long as the lifetime of the accessor.  Calling any
+member function on an accessor after its associated buffer has been destroyed
+results in undefined behavior.  In some cases the buffer destructor is defined
+to block until commands with requisites on that buffer have completed.  This
+results in well defined behavior so long as all invocations to member functions
+of accessors for that buffer occur before those commands complete.
```
keryell, line 6462: I have the feeling that the buffer constructor should block if there are accessors using it or at least being internally kept alive if there is no side effect on buffer use or buffer destruction.

adoc/chapters/programming_interface.adoc
```diff
@@ -8472,6 +8501,16 @@ template parameter as [code]#image_target::host_task# and then use the
...
 
+An unsampled image accessor is only a view over its associated image.
+Therefore, it is the application's responsibility to ensure that the lifetime
+of the image is at least as long as the lifetime of the accessor.  Calling any
+member function on an unsampled image accessor after its associated image has
+been destroyed results in undefined behavior.  In some cases the image
+destructor is defined to block until commands with requisites on that image
+have completed.  This results in well defined behavior so long as all
+invocations to member functions of accessors for that image occur before those
+commands complete.
```
keryell, line 8512: Should we have more content about `image` lifetime and synchronization?
I have the feeling we only addressed buffers up to now in the specification.

adoc/chapters/programming_interface.adoc
```diff
@@ -8744,6 +8783,15 @@ from a <<host-task>> are ill formed.  Likewise, programs which specify this
...
 [code]#sampled_image_accessor# from a <<sycl-kernel-function>> are ill formed.
 
+A sampled image accessor is only a view over its associated image.  Therefore,
+it is the application's responsibility to ensure that the lifetime of the image
+is at least as long as the lifetime of the accessor.  Calling any member
+function on a sampled image accessor after its associated image has been
+destroyed results in undefined behavior.  In some cases the image destructor is
+defined to block until commands with requisites on that image have completed.
+This results in well defined behavior so long as all invocations to member
+functions of accessors for that image occur before those commands complete.
```
keryell, line 8793: I have the feeling it depends only on construction and not member functions.

keryell (changes requested):
Diving more into it, I have the feeling it requires more discussion.
Selecting *Request changes* to avoid merging for now.

adoc/chapters/programming_interface.adoc
keryell: The case of the host accessor looks simpler, very synchronous without any lazy lambda evaluation and should keep a buffer internal state alive.

adoc/chapters/programming_interface.adoc
keryell: I can give up on this.

adoc/chapters/programming_interface.adoc
keryell: It looks more like implementation detail & trade-off, so I can give up my previous comment.

adoc/chapters/programming_interface.adoc
keryell: Your comment simplify the implementation, so I am fine with it.

keryell (commented):
I am still thinking that the `host_accessor` could keep a buffer alive internally.

gmlueck:
> I am still thinking that the `host_accessor` could keep a buffer alive internally.

What about a case like this:

```
host_accessor<int, 1> acc;
{
  int array[] = {1, 2, 3, 4};
  buffer b(array, range{4});
  acc = b.get_host_access(b);
}
bool test = (acc[0] == 1);  // Well defined?
```

Note that both the `buffer` and the underlying `array` go out of scope while there is still a live accessor.  Are you suggesting that the host accessor should make a copy of the `array` data, so that the last line still works even though the underlying `buffer` and `array` no longer exist?

To me, it seems better to adopt one of these two clarifications:

* Say that this is UB (the wording I have currently in this PR), or
* Say that the `buffer` destructor blocks while there are any outstanding accessors to it, which I think you proposed in another comment.  Thus, the code above would lead to a deadlock.

gmlueck:
I converted this to a draft because there is at least one thing that needs to be fixed before it can be merged.  Recent discussion in the internal issue 420 revealed that the current wording of this PR causes the following code snippet to be UB:

```
event e;
{
  buffer b{range{3}};
  e = q.submit([&](handler &cgh) {
    accessor a{b, cgh};
    cgh.single_task([=] {
      a[0] = 42;
    });
  });
  /* buffer goes out of scope */
}
/* command could run here */
e.wait();
```

The current PR makes this UB because the command could run even after the buffer is destroyed.

Code like this is currently legal according to the current wording of SYCL 2020, and I think we want to continue to allow code like this.  I see two options:

* Use the current wording in this PR, but make an exception for the case when a buffer is constructed *without* underlying host memory.  In such a case, we say that it *is* legal to use an accessor even after the underlying `buffer` is destroyed, or

* Say that the `buffer` destructor blocks while there are outstanding accessors to it.  In the code snippet above, the buffer destructor would block until the command completes.  Thus, the call to `e.wait()` is superfluous.
*

fraggamuffin:
use cases and examples needed

keryell:
Book 1 hour in advance to discuss this.

fraggamuffin:
a good F2F

keryell:
Thanks for the great presentation today explaining your proposal with all the corner cases about the life of `buffer` and `accessor`!
In triSYCL the `accessor` implementation extends the lifetime of the internal `detail::buffer` by keeping a reference on it:
https://github.com/triSYCL/triSYCL/blob/master/include/triSYCL/buffer/detail/accessor.hpp#L66
@illuhad I guess you might do something similar?

keryell:
@gmlueck can you send me your presentation so I can make a new version commenting your examples with a different hypothesis?

keryell:
I have update my presentation about this on the Khronos-internal https://members.khronos.org/wg/SYCL/document/31310

PR #241, opened 2022-03-24
Clarify that enumerator types are unspecified

gmlueck:
Clarify that the underlying type for all scoped enumerations is
implementation-defined.  Also clarify that the values of enumerators
within an enumeration are implementation-defined.

Closes internal issue 386


ProGTX (approved):

fraggamuffin:
LGTM

PR #243, opened 2022-03-29
Fix "!=" specification for NaN

gmlueck:
The `vec` and `marray` definitions of the `!=` operation were wrong
when one of the operands is NaN.  We previously said that this results
in `false`, but it should result in `true`.  This is in line with
IEEE 754 and with the OpenCL specification.  (The specification for all
the other relational operations was correct; they all result in `false`
when an operand is NaN.)

I also fixed an apparent typo.  The overloads for `vec` and `marray`
where the lhs is a scalar incorrectly said that they were *not*
available when `DataT` is a floating point type (for the `==`, `!=`,
etc. relational operations).  This, of course, makes no sense, so I
assume it was a copy/paste error.

Together with #234, this closes internal issue 600.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -16876,7 +16876,7 @@ a@
 ----
 vec<RET, NumElements> operatorOP(const vec& lhs, const vec &rhs)
 ----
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

fraggamuffin:
LGTM

keryell (approved):

PR #244, opened 2022-03-29
Fix "marray::operator!()"

gmlueck:
Fix an apparent copy/paste error with the specification of
`marray::operator!()`.  It appears that someone copy / pasted this
paragraph from the `vec` specification, but this is incorrect for
`marray`.


ProGTX (approved):

fraggamuffin:
LGTM

PR #245, opened 2022-03-30
Clarify note on simple parallel for arg types

gmlueck:
I found this section of the spec confusing when I was reading it
recently.  We say that the kernel argument must be `item` or be a type
that is implicitly convertible from `item`.  The next example shows a
kernel whose argument is `id`.  This makes no sense unless you remember
that `item` is implicitly convertible to `id`.  Make this a little
clearer by improving the non-normative note.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13233,8 +13233,13 @@ parameter.
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


tomdeakin (approved):
LGTM

ProGTX (approved):

keryell (approved):

PR #246, opened 2022-03-31
Clarify that exceptions in host_task are forwarded to async handler

### LINKED ISSUE #214: Exceptions inside `host_task`
The SYCL API prescribes scenarios where an exception must be thrown inside a `host_task`, for example calling `interop_handle::get_native_mem` on a placeholder accessor.  Is the intent here that users always catch these explicitly? What happens if an exception is not caught? Will it be reported through the async handler? Will it be swallowed?

Furthermore, what happens if user code throws an exception? It is not clear to me whether all asynchronous errors must also be derived from `sycl::exception`, which would make reporting user exceptions through the async handler problematic.

FWIW, ComputeCpp 2.8 seems to swallow exceptions, whereas DPC++ reports them through the async handler.

psalz:
I checked whether any of the descriptions for `wait` / `wait_and_throw` / `throw_asynchronous` or `exception_list` would need updating, but it does not seem so (although I find the description of the latter in section 4.13.2. to be worded somewhat strangely).

Fixes #214.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -14068,6 +14068,9 @@ The <<event>> returned by the submission of the associated <<command group>>
 enters the completed state (corresponding to a status of
 [code]#info::event_command_status::complete#) once the invocation of the
 provided {cpp} callable has returned.
+Any uncaught exception thrown during the execution of a <<host-task>> will be
+turned into an <<async-error>> that can be handled as described in
+<<subsubsec:exception.async>>.
 
 A <<host-task>> can optionally be used to interoperate with the
 <<native-backend-object,native backend objects>> associated with the <<queue>> executing the


gmlueck (approved):

fraggamuffin:
LGTM

psalz:
Point brought up by @keryell during WG meeting: Should user exceptions be forwarded to the async handler as-is, or somehow get wrapped into a `sycl::exception`?

psalz:
> Point brought up by @keryell during WG meeting: Should user exceptions be forwarded to the async handler as-is, or somehow get wrapped into a `sycl::exception`?

To expand a bit more on this, the canonical way of wrapping exceptions seems to be [`std::nested_exception`](https://en.cppreference.com/w/cpp/error/nested_exception). An implementation could thus do something like

```c++
try {
   run_host_task();
} catch(...) {
   std::throw_with_nested(sycl::exception(sycl::make_error_code(sycl::errc::host_task), "An exception was thrown inside host_task"));
}
```

which would then allow users to write more generic asynchronous handlers. Thoughts on this?

keryell:
Yes, this is `std::nested_exception` I was thinking too.
But in a way to KISS, I cannot see why we cannot rethrow directly the user exception without wrapping it in a nested exception.

keryell (approved):
Thanks for the clarification.

PR #247, opened 2022-03-31
Clarify what happens when submitting more than 1 command per CGF

### LINKED ISSUE #223: Clarify what happens when multiple commands are submitted in a single CGF
Section "4.9.3 Command group scope" states:

> A command group scope, as defined in Section 3.7.1, may execute a single command such as invoking a
kernel, copying memory, or executing a host task. It is legal for a command group scope to statically contain
more than one call to a command function, but any single execution of the command group function
object may execute no more than one command.

There is however no mention about what happens when a user *does* try to execute multiple commands in a single CGF. I also could not find any other location talking about this (although I may have missed something). In any case, I think this paragraph should be extended to mention that a synchronous exception will be thrown, either with `ercc:invalid` or even a new error code.

psalz:
Fixes #223.

OT: Any thoughts on switching to soft line breaks with a hard break after each sentence? Tends to make for much nicer diffs...

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12761,7 +12761,9 @@ may execute a single <<command>> such as invoking a kernel, copying memory,
 or executing a host task.  It is legal for a <<command-group-scope>> to
 statically contain more than one call to a <<command>> function, but any
 single execution of the <<command-group-function-object>> may execute no more
-than one <<command>>.  The statements that call <<command, commands>> together with
+than one <<command>>. If an application fails to do this, the implementation
+must throw a synchronous [code]#exception# with the [code]#errc::invalid# error
+code. The statements that call <<command, commands>> together with
 the statements that define the requirements for a kernel form the
 <<command-group-function-object>>.  The command group
 function object takes as a parameter an instance of the <<handler>> class which
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12761,7 +12761,8 @@ may execute a single <<command>> such as invoking a kernel, copying memory,
 or executing a host task.  It is legal for a <<command-group-scope>> to
 statically contain more than one call to a <<command>> function, but any
 single execution of the <<command-group-function-object>> may execute no more
-than one <<command>>. If an application fails to do this, the implementation
+than one <<command>>. If an application fails to do this, the function that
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -12761,7 +12761,9 @@ may execute a single <<command>> such as invoking a kernel, copying memory,
 or executing a host task.  It is legal for a <<command-group-scope>> to
 statically contain more than one call to a <<command>> function, but any
 single execution of the <<command-group-function-object>> may execute no more
-than one <<command>>.  The statements that call <<command, commands>> together with
+than one <<command>>. If an application fails to do this, the implementation
+must throw a synchronous [code]#exception# with the [code]#errc::invalid# error
+code. The statements that call <<command, commands>> together with
```
gmlueck, line 12766: Should we specify which function throws the exception?  With the current wording, it's not clear if it would be `handler::parallel_for()` or `queue::submit()`.  I think the exception should be thrown from `queue::submit()`, which is consistent with the point at which we create the requisites from accessors, and it is also consistent with the timestamp we return for `info::event_profiling::command_submit`.  If you agree, maybe word it like:

> If an application fails to do this, the function that submits the \<\<command\>\> (e.g. [code]#queue::submit#) must throw a synchronous [code]#exception# with the [code]#errc::invalid# error code.

psalz: Makes sense, however I'm not sure about the wording. Wouldn't `queue::submit` be submitting the corresponding *command group function object*, instead of the *command* itself? Also, is there any other way to erroneously submit multiple commands, other than `queue::submit`? This can't happen for the convenience shortcut functions after all.

fraggamuffin:
queue::submit is enough

psalz:
I've further clarified the wording to state that `queue::submit` must throw the exception.

adoc/chapters/programming_interface.adoc
keryell: > Wouldn't `queue::submit` be submitting the corresponding _command group function object_, instead of the _command_ itself?

I have the feeling this is already what you say and that sounds good.

keryell (approved):

gmlueck (approved):

adoc/chapters/programming_interface.adoc
gmlueck: Yes, this wording looks good to me too.

PR #249, opened 2022-04-08
Add missing prototype for get_sub_group()

gmlueck:
The synopsis for `nd_item` was mistakenly missing the declaration for
`get_sub_group()`.  There was already a description for this function
in the table of `nd_item` member functions; it was only missing from
the synopsis.

### DIFF
--- a/adoc/headers/nditem.h
+++ b/adoc/headers/nditem.h
@@ -23,6 +23,8 @@ public:
 
   group<Dimensions> get_group() const;
 
+  sub_group get_sub_group() const;
+
   size_t get_group(int dimension) const;
 
   size_t get_group_linear_id() const;


keryell (approved):
Perhaps you could put it later to be sorted?

gmlueck:
> Perhaps you could put it later to be sorted?

I did consider other locations when adding this function to the synopsis.  It seems like we generally order functions in the synopsis "thematically", not alphabetically.  I think we also generally use the same order in the synopsis as the table that defines each function.  Since the table already had this function, it seemed best to just add it to the synopsis in the same location.  Otherwise, we would need to reorder the table also.

keryell:
> I did consider other locations when adding this function to the synopsis. It seems like we generally order functions in the synopsis "thematically", not alphabetically.

OK, that makes sense.

ProGTX (approved):

PR #251, opened 2022-04-13
Remove & from accessor-like subscript operators

Pennycook:
Prior to this commit, the subscript operators of all accessors returned
a reference to an unspecified class (__unspecified__ &). The only
purpose of this class is to expose other subscript operators.

Forcing implementations to return a reference is unnecessary, and adds
implementation complexity because of lifetime considerations.

This commit removes the & from the subscript operators of all
accessor-like classes (accessor, local_accessor, host_accessor,
reducer). Implementations that wish to return a reference can still do
so, because the return type remains __unspecified__.


keryell (approved):
Thanks for making the unspecification more unspecified. :-)

ProGTX (approved):

PR #253, opened 2022-04-25
Fix bugs in USM API signatures

gmlueck:
Fix three bugs in the API signatures for USM:

* There is no reason to require the implementation of `usm_allocator`
  to use `default` versions of the copy / move constructors and
  assignment operators.  The only reason we might require this is to
  make `usm_allocator` a trivially copyable type.  However, it can not
  be trivially copyable because all implementations need to store a
  `context` object, which is not trivially copyable.

* The `usm_allocator` copy assignment operator should not be declared
  `delete`.  I think this was just a typo.

* The `context` and `queue` parameters to `sycl::free()` should be
  "const reference" instead of "reference".  This makes them consistent
  with all the other USM APIs that take `context` or `queue` parameters.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10065,10 +10065,10 @@ public:
                 const property_list &propList = {}) noexcept;
   usm_allocator(const queue &q,
                 const property_list &propList = {}) noexcept;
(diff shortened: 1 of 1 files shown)


keryell (approved):
Good catch.

PR #254, opened 2022-04-25
Clarify behavior of queue destructor

gmlueck:
The spec already said that the `queue` destructor does not block in
section 3.9.8.1 "Synchronization in the SYCL application":

> Note that the destructors of other SYCL objects (sycl::queue,
> sycl::context,...) do not block.

However, we never really said what happens to outstanding commands
that have been submitted to the queue.  This PR clarifies that these
commands just run normally, with the same behavior as though the
queue had not been destroyed.

This seems like a sensible behavior, and it also enables a use case
similar to `std::thread::detach`.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2922,6 +2922,14 @@ Each constructor takes as the last
 parameter an optional SYCL [code]#property_list# to provide properties to
 the SYCL [code]#queue#.
 
+A SYCL [code]#queue# may be destroyed even when there are uncompleted
+<<command, commands>> that have been submitted to the queue.  Doing so does not
+block.  Instead, any commands that have been submitted to the queue begin
+execution when their requisites are satisfied, just as they would had the queue
+not been destroyed.  Any event objects for those commands are signaled in the
+normal manner when the command completes.  Resources associated with the queue
+will be freed by the time the last command completes.
+
 The SYCL [code]#queue# class provides the common reference semantics
 (see <<sec:reference-semantics>>).
 


illuhad:
Can we guarantee that this actually works? E.g.
```c++
int main() {
  sycl::queue q;
  q.submit(/* something */);
}
```
Here it can well happen that things only get processed when the program is already shutting down, which might cause issues with lower layers like drivers etc.

AerialMantis:
I think this is a good clarification.

If the program ends (i.e. the main function returns, not considering static global variables) while there are still commands being executed then I think this needs to be UB. We could incorporate this into the wording so state that commands will continue to execute as they would if the queue hadn't been destroyed whilst the program is still executing, however, the program exiting whilst commands are still executing is undefined behaviour.

gmlueck:
I agree with @AerialMantis.  There is no guarantee that commands on the queue will run to completion if the program terminates.  Note that this issue exists even before this PR.  Consider an application like this:

```
#include <sycl/sycl.hpp>
#include <cstdlib>

int main() {
  sycl::queue q;
  q.submit(/* something */);
  std::exit(0);
}
```

This issue is also not unique to SYCL.  With plain C++ you have a similar issue with threads.  What if the application starts a thread and then exits before that thread completes?  The C++ spec does not provide any guarantee that the application waits for the thread to terminate before exiting.  Instead, the application just exits and the thread terminates abruptly.

tomdeakin:
What happens in the following, where the deconstructor order matters? The buffer deconstructor should block, so the kernel will run, but the queue will have already been destroyed before the buffer deconstructor is called.

```
#include <sycl/sycl.hpp>
#include <cstdlib>

int main() {
  sycl::buffer<int, 1> buf(1);
  {
    sycl::queue q;
    q.submit([&](sycl::handler cgh) {
      sycl::accessor foo {cgh, buf};
      /* more */
    });
  }
}
```

AerialMantis:
So this is something we were discussing recently as well, in regards to interop and ownership. So the `queue` does not block on destruction, but it also doesn't need to exist for the commands to execute. So in effect the `queue` enqueues commands to the runtime, but the `buffer` destructor doesn't wait on the `queue` it waits on the commands that were submitted, so the `queue` only needs to exist to enqueue the commands.

tomdeakin:
OK, makes sense @gmlueck, thanks. We should  be careful in how we word the clarification about UB to not contradict that: i.e. restrict the UB to only when the program exits. I'm not sure there is wording for that in the change yet though?

gmlueck:
I'm a little unsure what to say for the case when the application exits with uncompleted commands.  I don't think we want to say this is UB, do we?  Many real-world application do call `std::exit()` when a fatal error happens, even if there are pending threads or pending SYCL commands.  If we say it is UB, then the implementation could crash or hang in this case, and I don't think we want to allow that.

I tried looking in cppreference for some similar language about calling `std::exit()` when there are uncompleted threads.  However, I couldn't find any description of this case.  @nliber, do you know if the C++ specification says anything about this case?

AerialMantis:
It would be good if we could say that the state of the commands is UB the program should still shutdown successfully, but that might be tricky as an `std::exit()` could be called at any arbitrary point, and it might be difficult to guarantee that a backend API will handle this gracefully.

keryell (approved):
That sounds good.

PR #255, opened 2022-04-29
Clarify wording for "device selector" in the Glossary

aelovikov-intel:
[4.6.1.1] states

   If the highest value is negative no device is selected.

As such, be more precise about the behavior for "0" 
in the glossary wording.


keryell (approved):
Good clarification.

PR #256, opened 2022-05-05
Loosen restrictions with sub-devices and context

gmlueck:
After gaining some implementation experience, we've learned that
most SYCL APIs that take both a `device` and a `context` should have
a looser requirement on the relationship between that device and
context.  In most cases, the spec previously said something like
"the context must encapsulate the device".  This is a little ambiguous
because we don't define what "encapsulate" means.

If we assume a strict definition like "the device object must be one of
the devices used to construct the context", this leads to inconvenient
code for applications that use sub-devices.  It is very common for
applications to create a single context that contains all of the root
devices in a platform and then use that context throughout the
application.  However, the strict definition does not work well for
applications like this:

```
void foo() {
  // Get the default device and create a context that contains all
  // of the devices in the same platform.
  sycl::device d;
  sycl::context c{d.get_platform().get_devices()};

  // Sometime later the application creates a sub-device.
  sycl::device sd =
    d.create_sub_devices(sycl::info::partition_affinity_domain::numa)[0];

  // This fails because "sd" is not contained by context "c".
  sycl::queue q{sd, c};

  // This fails for the same reason.
  void *p = sycl::malloc_device(N, sd, c);
}
```

Of course, the application could avoid the problem by creating a new
context that contains the sub-device `sd`.  However, this is
inconvenient if the application creates sub-devices "on the fly".  It
also requires the application to create a new context each time it
creates a sub-device, which could be inefficient.

We think a better solution is to have a looser definition like "the
device object must either be contained by the context or be a sub-device
of some device contained by the context".  By adopting this
interpretation, the example code above works without any changes.

We think this is a better interpretation because it makes SYCL easier
to use and improves efficiency by reducing the number of contexts that
an application needs to create.  We also think this interpretation can
be implemented by backends.

This PR changes the wording of all APIs that take both a `device` and
a `context` to clarify that the device may also be a sub-device (or
sub-sub-device, etc.) of the context.

As a side-effect of this change, the `noexcept` keyword was removed
from two of the `usm_allocator` constructors because we want to throw
an exception when the device parameter is not contained by the context.
I think it was a mistake to have `noexcept` here in the first place
because these constructors also take a `property_list` parameter, and I
presume we would want to diagnose errors with incorrect properties by
throwing an exception.

Closes internal issue 598


keryell (changes requested):
Great idea!
Can you add a "what changed" entry?

fraggamuffin:
LGTM with the change

gmlueck:
> Great idea! Can you add a "what changed" entry?

Done in 59249822b6550a07556ad9b4c522b0e0034a0dcc.  I only noted the change in the `queue` constructor overloads because the "What's changed" section only describes changes since 1.2.1 and the USM and `kernel_bundle` APIs (which also changed in this PR) did not exist in SYCL 1.2.1.

keryell (approved):
Great!

PR #257, opened 2022-05-05
Fix name of error code

gmlueck:
There is no error code named `invalid_object`.  This got renamed a
while ago, and we must have missed this one.


ProGTX (approved):

keryell (approved):
Nice catch!

PR #260, opened 2022-05-09
Improve and clarify host task interop description

npmiller:
This patch doesn't introduce any changes in how host task interop works,
but improves its description and fixes some contradictions within the
spec, it does the following:

* Add host-task interoperability as a third type of interop along with
  sycl-application interoperability and kernel function interoperability
  in the main interoperability section.
* Fix wording that said that capturing a `target::device` accessor in a
  host task is ill formed. It is allowed and used for interop.
* Change the wording that says that capturing any SYCL class that has
  reference semantics is undefined behavior to make an exception for
  accessors.
* Clarify that The `get_native_mem` functions return the native handle
  for the underlying memory objects rather than of the accessor.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -177,13 +177,16 @@ interface for interoperating between the <<sycl-runtime>> object and the
 <<native-backend-object>> in order to support interoperability within an
 application between SYCL and the associated <<backend-api>>.
 
-There are two forms of interoperability with <<sycl-runtime>> classes:
-interoperability on the <<sycl-application>> with the <<backend-api>>
-and interoperability within a <<sycl-kernel-function>> with the equivalent
-kernel language types of the <<backend>>. <<sycl-application,SYCL application>>
-interoperability and <<sycl-kernel-function>> interoperability are
-provided via different interfaces and may have different
-<<native-backend-object>> types.
+There are three forms of interoperability with <<sycl-runtime>> classes:
+interoperability on the <<sycl-application>> with the <<backend-api>>,
+interoperability within a <<sycl-kernel-function>> with the equivalent kernel
+language types of the <<backend>>, and interoperability within a <<host-task>>
+with the [code]#interop_handle#.
+
+<<sycl-application,SYCL application>> interoperability,
+<<sycl-kernel-function>> interoperability and <<host-task>> interoperability
+are provided via different interfaces and may have different behavior for the
+same SYCL object.
 
 <<sycl-application,SYCL application>> interoperability may be provided for
 [code]#buffer#,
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -6475,10 +6487,12 @@ following values for the [code]#AccessMode# template parameter:
     | Access a buffer from a <<host-task>>.
 |====
 
-Programs which specify the access target as [code]#target::device# and then use
-the [code]#accessor# from a <<host-task>> are ill formed.  Likewise, programs
-which specify the access target as [code]#target::host_task# and then use the
-[code]#accessor# from a <<sycl-kernel-function>> are ill formed.
+Programs which specify the access target as [code]#target::device# and then
+capture the [code]#accessor# in a <<host-task>> can only use the accessor for
+interoperability through the [code]#interop_handle#.
```
gmlueck, line 6492:
```suggestion
capture the [code]#accessor# in a <<host-task>> can use the accessor for
interoperability through the [code]#interop_handle#.  Any other use of of such an
accessor in a <<host-task>> results in undefined behavior.
```

I think we should state explicitly that you get UB if you use a `device` accessor in any other way from a host task.

Please also close internal issue 609 when this is merged.

adoc/chapters/programming_interface.adoc
```diff
@@ -6475,10 +6487,12 @@ following values for the [code]#AccessMode# template parameter:
...
+
+Programs which specify the access target as [code]#target::host_task# and then
+use the [code]#accessor# from a <<sycl-kernel-function>> are ill formed.
```
gmlueck, line 6495:
```suggestion
use the [code]#accessor# from a <<sycl-kernel-function>> result in undefined behavior.
```

It's probably better to  be consistent with the previous paragraph and say this  is UB vs. ill formed.  Does the term "ill formed" mean that the compiler is required to issue a diagnostic?

adoc/chapters/programming_interface.adoc
```diff
@@ -14233,8 +14247,11 @@ defined in <<sub.section.memmodel.app>>.
 Since a <<host-task>> is invoked directly by the <<sycl-runtime>> rather
 than being compiled as a <<sycl-kernel-function>>, it does not have the same
 restrictions as a <<sycl-kernel-function>>, and can therefore contain any
-arbitrary {cpp} code. However, capturing or using any SYCL class that has reference
-semantics (see <<sec:reference-semantics>>) is undefined behavior.
+arbitrary {cpp} code.
+
+Capturing <<accessor,accessors>> in a <<host-task>> is allowed, however,
+capturing or using any other SYCL class that has reference semantics (see
+<<sec:reference-semantics>>) is undefined behavior.
```
gmlueck, line 14254: Thanks.  Please also  close internal issue 605 when this is merged.

adoc/chapters/programming_interface.adoc
npmiller: Updated as suggested to specify that it is undefined behavior.

gmlueck (approved):

tomdeakin (approved):
LGTM

keryell (approved):

PR #261, opened 2022-05-10
Stable order of platform and device objects

gmlueck:
Clarify that the order of the `platform` and `device` objects is the
same each time `get_platforms()` or `get_devices()` is called.  This
allows an application to store the index of a platform or device and
use that index later to identify the same platform or device.

Also clarify that the `get_devices()` function returns only the root
devices, not any sub-devices that have been created.

Closes internal issue 603.

### DIFF
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -416,6 +416,10 @@ object. For the full description please refer to <<subsec:buffers>>.
     non-associative or non-commutative, the behavior of a reduction may be
     non-deterministic.
 
+[[root-device]]root device::
+    A device that is not a sub-device.  The function
+    [code]#device::get_devices()# returns a vector of all the root devices.
+
 [[rule-of-five]]rule of five::
     For a given class, if at least one of the copy constructor, move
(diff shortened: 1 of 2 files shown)


tomdeakin:
I think we need to define "root device", given that we use the words parent and descendant when referring to devices and sub-devices. Also, it's possible there is an interpretation for root device conflating with a [PCIe root complex device](https://en.wikipedia.org/wiki/Root_complex).

Note the SYCL platform model section doesn't talk about sub-devices at all, which might complicate the issue.

It might be simplest to just add a glossary entry for "root device" and add cross-references/update the phrasing in any place we mean a parent device and not a sub-device.

What do you think?

gmlueck:
> I think we need to define "root device"

Yes, that's a good point.  I added a glossary entry for "root device" and changed the words "root device" to be a link to this glossary entry.  This was done in 5e101eeef8482b97ed00dba1ef3d56b36717d145

tomdeakin (approved):
LGTM now

ProGTX (approved):

fraggamuffin:
LGTM

PR #262, opened 2022-05-13
some math definition corrections/clarifications

JackAKirk:
logb / ilogb had incorrect/unclear definitions: The "exponent of x" is clearly not "the integral
      part of logr (latexmath:[|x|])": we can't find an exponent of x from a function of x, the exponent `n` of `x` needs to be defined, i.e. `x^n`.

For fmod I am still confused and would appreciate any clarification anyone has: If we are multiplying `x mod y` by `trunc (x/y)` then why is the function called fmod? DPC++ implementation of fmod is just doing `x mod y`. So which is wrong, definition or implementation?

**Further Note:** atan2pi is also not rendering correctly in the spec PDF, although the latex looks fine so I'm unsure why this is or how to fix it.

Signed-off-by: JackAKirk <jack.kirk@codeplay.com>

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20898,7 +20898,7 @@ a@
 ----
 genint ilogb (genfloat x)
 ----
-   a@ Compute the integral part of logr (latexmath:[|x|]) and return the result as an integer, where r is the value returned by std::numeric_limits<genfloat>::radix.
+   a@ Compute the integral part of logr (latexmath:[|x|]) and return the result as an integer, where r is the value returned by [code]#std::numeric_limits<genfloat>::radix#.
 
 a@
 [source]
@@ -20960,7 +20960,7 @@ a@
 ----
 genfloat logb (genfloat x)
 ----
-   a@ Compute the integral part of logr (latexmath:[|x|]), where r is the value returned by std::numeric_limits<genfloat>::radix.
+   a@ Compute the integral part of logr (latexmath:[|x|]), where r is the value returned by [code]#std::numeric_limits<genfloat>::radix#.
 
 a@
 [source]
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20865,7 +20865,7 @@ a@
 ----
 genfloat fmod (genfloat x, genfloat y)
 ----
-   a@ Modulus. Returns latexmath:[x \bmod y].
+   a@ Modulus. Returns latexmath:[x - y \cdot \mathrm{trunc}(x/y)].
 
 a@
 [source]
@@ -20898,7 +20898,7 @@ a@
 ----
 genint ilogb (genfloat x)
 ----
-   a@ Compute the integral part of the logarithm of latexmath:[x]. The base of the logarithm is the radix (integer base) used by the representation of floating point numbers.
+   a@ Compute the integral part of logr (latexmath:[|x|]) and return the result as an integer, where r is the value returned by std::numeric_limits<genfloat>::radix.
 
 a@
 [source]
@@ -20960,7 +20960,7 @@ a@
 ----
 genfloat logb (genfloat x)
 ----
-   a@ Compute the integral part of the logarithm of latexmath:[|x|]. The base of the logarithm is the radix (integer base) used by the representation of floating point numbers.
+   a@ Compute the integral part of logr (latexmath:[|x|]), where r is the value returned by std::numeric_limits<genfloat>::radix.
 
 a@
 [source]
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -20865,7 +20865,7 @@ a@
 ----
 genfloat fmod (genfloat x, genfloat y)
 ----
-   a@ Modulus. Returns latexmath:[x \bmod y \cdot \mathrm{trunc}(x/y)].
+   a@ Modulus. Returns latexmath:[x \bmod y].
```
gmlueck, line 20868: These functions all come from OpenCL.  [OpenCL 3.0](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#math-functions) defines it as:

> Modulus. Returns x - y * trunc(x/y).

where the `trunc` function is defined later (in both OpenCL and in SYCL).

This might be a better way to define the semantics, rather than defining it as  `x mod y`, because there is no formal definition in the spec for `mod`.

adoc/chapters/programming_interface.adoc
```diff
@@ -20960,8 +20960,7 @@ a@
 ----
 genfloat logb (genfloat x)
 ----
-   a@ Compute the exponent of x, which is the integral
-      part of logr (latexmath:[|x|]).
+   a@ Compute the integral part of the logarithm of latexmath:[|x|]. The base of the logarithm is the radix (integer base) used by the representation of floating point numbers.
```
gmlueck, line 20963: FWIW,  OpenCL 3.0 has exactly the same wording as SYCL (prior to your PR).  I _think_ this is correct, except that the base `r` isn't defined.  The same function exists in ISO C++, and [cppreference](https://en.cppreference.com/w/cpp/numeric/math/logb) defines it this way:

> Extracts the value of the unbiased radix-independent exponent from the floating-point argument arg, and returns it as a floating-point value
> [...]
> Formally, the unbiased exponent is the signed integral part of `log
 r|arg|` (returned by this function as a floating-point value), for non-zero `arg`, where `r` is `std::numeric_limits<T>::radix` and `T` is the floating-point type of `arg`. If `arg` is subnormal, it is treated as though it was normalized.

For any sane implementation, I think `r` is `2`, but I guess IEEE 754 allows other  radices for floating-point representation?

I think your description is equivalent, except it might be better to replace "the radix (integer base) used by the representation of floating point numbers" with "`std::numeric_limits<T>::radix`, where `T` is the type of `x`".

adoc/chapters/programming_interface.adoc
JackAKirk: Ah OK, yeah the OpenCL definition makes sense: looks like it was just a typo where '\bmod' repaced `-`. I can switch it to OpenCL def.

adoc/chapters/programming_interface.adoc
gmlueck: Yes, I think it is OK to explain (in words) what `r` is, but we should also say that it is the value returned by `std::numeric_limits<T>::radix` since that is a more formal definition.

adoc/chapters/programming_interface.adoc
JackAKirk: Done

> Ah OK, yeah the OpenCL definition makes sense: looks like it was just a typo where '\bmod' repaced `-`. I can switch it to OpenCL def.

Done

adoc/chapters/programming_interface.adoc
JackAKirk: > std::numeric_limits<T>::radix

I've just referred to it as std::numeric_limits<T>::radix . I think this is fine. Happy to add a wordy description of you have a suggestion.  Thanks

adoc/chapters/programming_interface.adoc
```diff
@@ -20898,7 +20898,7 @@ a@
 ----
 genint ilogb (genfloat x)
 ----
-   a@ Return the exponent as an integer value.
+   a@ Compute the integral part of logr (latexmath:[|x|]) and return the result as an integer, where r is the value returned by std::numeric_limits<genfloat>::radix.
```
gmlueck, line 20901:
```suggestion
   a@ Compute the integral part of logr (latexmath:[|x|]) and return the result as an integer, where r is the value returned by [code]#std::numeric_limits<genfloat>::radix#.
```

Probably, this should be formatted as code.  Same for the description of `logb`.

adoc/chapters/programming_interface.adoc
gmlueck: I think this looks good.  I just had a small comment on formatting.

adoc/chapters/programming_interface.adoc
JackAKirk: Thanks for the input. I've used the code formatting now.

TApplencourt (approved):

keryell (approved):

PR #263, opened 2022-05-18
Clarify get_pointer_{type,device}

gmlueck:
Clarify several things about the USM query functions `get_pointer_type`
and `get_pointer_device`:

* The return value of `get_pointer_device` is the same `device` that
  was used to allocate the USM memory (for "device" and "shared"
  allocations).  Some readers were previously confused about the
  behavior when the allocation was made against a sub-device.

* For `get_pointer_device`, the pointer need not be at the start of the
  USM memory region.  It can be any pointer within the USM memory
  region.  We said this already for `get_pointer_type`, but this
  wording was missing from `get_pointer_device`.

* For both queries, the USM memory region must correspond to the
  `context` that is passed.  We didn't say this before, but it was
  clearly our intent.  Otherwise, there would be no need for these
  functions to take a `context` parameter.

Closes internal issue 599.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10914,20 +10914,25 @@ a@
 usm::alloc get_pointer_type(const void *ptr,
                             const context &ctxt)
 ----
-a@ Returns the USM allocation type for [code]#ptr# if [code]#ptr#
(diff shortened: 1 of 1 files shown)


keryell (approved):
Sounds good!

ProGTX (approved):

fraggamuffin:
LGTM

Issue #97, opened 2020-05-12
User adding symbols to sycl namespace


rolandschulz:
Currently the spec only specifies how the implementer should use the sycl namespace. Section 4.1  states that extension should go into vendor namespace and implementation details into `detail`. It doesn't put any restriction on user code. Is the user allowed to put anything in sycl namespace? I assume the intention is that this isn't allowed. It should be clarified.

keryell:
Yes, we should prevent this, since it can break implementations and have various undefined behaviors.

fraggamuffin:
already disallowed
done and close

keryell:
Closed by https://github.com/KhronosGroup/SYCL-Docs/pull/268

PR #268, opened 2022-07-29
Disallow declaring anything in sycl:: namespace or sub-namespaces

keryell:
This clarifies https://github.com/KhronosGroup/SYCL-Docs/issues/97

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -105,6 +105,12 @@ where [code]#<backend_name># is the name of the <<backend>> as defined in the
 namespaces for any extensions that a vendor may provide, including any
 <<backend>> that the vendor may define outside of the Khronos SYCL group.
 
+An end-user should not declare anything in the [code]#::sycl#
+namespace or its nested namespaces, except if allowed for specific
+cases by the specification itself (like customization points in some
+extensions for example).
+
+
 == Class availability
 
 In SYCL some <<sycl-runtime>> classes are available to the SYCL application,
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -105,10 +105,9 @@ where [code]#<backend_name># is the name of the <<backend>> as defined in the
 namespaces for any extensions that a vendor may provide, including any
 <<backend>> that the vendor may define outside of the Khronos SYCL group.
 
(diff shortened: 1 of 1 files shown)


keryell:
I have not dived into Ruby and AsciiDoctor for a while and I see that there are now some compilation errors.
@oddhack Is there something obvious that changed in https://github.com/KhronosGroup/SYCL-Docs/runs/7571208025?check_suite_focus=true since last commit in https://github.com/KhronosGroup/SYCL-Docs/runs/6510988109?check_suite_focus=true ?

oddhack:
I did update the asciidoctor-spec Docker image recently and that is likely the source of the errors. Let me experiment a bit tonight; at worst we can back off to an older version of the tools while sorting out what changed.

Because of persistent difficulty getting GH Actions and gitlab CI to flush their container image cache when versions are updated, I've started moving to specifying images by their SHA256 hash instead of just the name. Probably the right answer for SYCL spec as well.

ProGTX (approved):

fraggamuffin:
http://eel.is/c++draft/namespace.std

adoc/chapters/programming_interface.adoc
```diff
@@ -105,6 +105,12 @@ where [code]#<backend_name># is the name of the <<backend>> as defined in the
 namespaces for any extensions that a vendor may provide, including any
 <<backend>> that the vendor may define outside of the Khronos SYCL group.
 
+An end-user should not declare anything in the [code]#::sycl#
+namespace or its nested namespaces, except if allowed for specific
+cases by the specification itself (like customization points in some
+extensions for example).
+
+
```
gmlueck, line 113: The C++ spec uses this wording:

> Unless otherwise specified, the behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std[.](http://eel.is/c++draft/namespace.std#1.sentence-1)

fraggamuffin:
https://isocpp.org/std/standing-documents/sd-8-standard-library-compatibility
thanks Nevin
use any entity, and unless otherwise specified

fraggamuffin:
Unless otherwise specified, the behavior of a SYCL program is undefined if it adds any entity to namespace sycl  or to a namespace within namespace sycl

gmlueck (approved):

fraggamuffin:
LGTM

Issue #153, opened 2021-06-29
Normalized the position of * and & in source code examples and API


keryell:
There are some `T *p`, `T* p`, `T * p`,...
I suggest `T& r` and `T* p` since it is part of the type and less confusing.
Single line multiple declarations like
```c
T a, *p, b, *q;
```
is a confusing notation inherited from C I would not take as a rationale towards the other choice. :-)

keryell:
This has been solved.

PR #269, opened 2022-07-29
Normalize position of & and * in the source code parts of the specification

keryell:
Put the * and & close to the type.

The changes in the specification itself is manual work because of the intermixing of AsciiDoctor and C++ snippets but clang-format was used for changing the headers and code samples.
Since clang-format has been used, the coding style has been homogenized.

This should solve https://github.com/KhronosGroup/SYCL-Docs/issues/153

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,8 @@
+Language: Cpp
+BasedOnStyle: LLVM
+AccessModifierOffset: -1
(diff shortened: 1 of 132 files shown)


fraggamuffin:
need one more approval.

tomdeakin (approved):

fraggamuffin:
LGTM

PR #271, opened 2022-08-02
Fix for PDF build problem vs. asciidoctor-pdf

oddhack:
Per
https://github.com/KhronosGroup/SYCL-Docs/pull/268#issuecomment-1198790889

A recent update to the asciidoctor-spec Docker image moved it from
asciidoctor-pdf 1.6.1 to 2.2.0, and deprecated the `pdf-stylesdir`
attribute, causing the PDF build to fail. This branch updates the
deprecated attribute and adds a note to the repository README. It also updates the custom PDF theme to match changes in the current asciidoctor-pdf default theme, and switches from 'optimize-pdf' to 'hexapdf' to reduce PDF output size. 'hexapdf' is found in the latest asciidoctor-spec Docker image.

I believe this fixes all the issues noted on !268. There is a small visual difference making the PDF spec about 5 pages shorter. I think this may be due to an updated table layout configuration that operates more compactly - hopefully it's not an issue as finding the root cause could take a while.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,17 @@ Dockerfile at
 
 https://github.com/KhronosGroup/DockerContainers/blob/master/asciidoctor-spec.dockerfile
 
-Note that the Khronos Docker image layers on the official Ruby 2.7 Docker
+Note that the Khronos Docker image layers on the official Ruby 3.1 Docker
 image, so you must install Ruby first.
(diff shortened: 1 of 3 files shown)


keryell (approved):
Thanks for fixing this!

PR #274, opened 2022-08-15
Small clarification to SYCL_EXTERNAL

gmlueck:
State that `SYCL_EXTERNAL` can only be applied to functions with
external linkage, which clarifies that it does not "override" the C++
linkage.


keryell (approved):
Yes, that sounds good not to add too much magic in SYCL.

TApplencourt (approved):

PR #275, opened 2022-08-18
Clarify device selector

gmlueck:
Make two clarifications to the specification of device selector:

* Clarify that the implementation passes only the root devices to the
  selector, not any sub-devices.  I think this was our intent, but we
  didn't actually say this.

* Clarify that the word "callable" means the C++ named requirement
  "Callable".  (This was a point of confusion recently, in our team.)
  In other places where we refer to a C++ named requirement, we
  capitalize the name and use code font, so I did the same here.


keryell (approved):
Thanks.

fraggamuffin:
LGTM

PR #276, opened 2022-08-22
Fix specialization id example

gmlueck:
Some of the comments in this example got accidentally deleted when clang-format was run.  Add those comments back.

Also make a small tweak to the example to avoid a C++ syntax error, from this:

```
namespace { /*...*/ }

inline namespace {
}
```

to:

```
namespace { /*...*/ }

inline namespace other2 {
}
```


keryell (approved):
Thanks.

TApplencourt (approved):

fraggamuffin:
LGTM

PR #278, opened 2022-08-25
Add operator= for the 0-dimensional accessors

aelovikov-intel:
That enables uses like this:

```
  q.submit([&](sycl::handler &cgh) {
    sycl::local_accessor<int, 0> A{cgh};
    cgh.single_task<class Kernel>([=] { A = 42; });
  });
```

This PR also fixes incosistency in the description of "operator
reference" for buffer accessor - it is only available for non-atomic
access-mode as mentioned in some places but not in its synopsis.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6393,7 +6393,8 @@ class in the sections that follow.
 Accessors of type [code]#accessor#, [code]#host_accessor#, and
 [code]#local_accessor# can have zero, one, two, or three Dimensions.  A zero
 dimension accessor provides access to a single scalar element via an implicit
-conversion operator to the underlying type of that element.
+conversion operator to the underlying type of that element and via an overloaded
+assignment operator from the underlying type of the element.
 
 One, two, or three dimensional specializations of these accessors provide
 access to the elements they contain in two ways.  The first way is through a
@@ -8337,6 +8338,22 @@ For [code]#host_accessor# and [code]#local_accessor# available only when
 [code]#(Dimensions == 0)#.
 
 Returns a reference to the single element that is accessed by this accessor.
+
+For [code]#accessor# and [code]#local_accessor#, this function may only be
+called from within a <<command>>.
+
+a@
+[source]
+----
+reference operator=(const value_type &&other) const
+----
+   a@ For [code]#accessor# available only when
+      [code]#(AccessMode != access_mode::atomic && Dimensions == 0)#.
+
+For [code]#host_accessor# and [code]#local_accessor# available only when
+[code]#(Dimensions == 0)#.
+
+Assignment to the single element that is accessed by this accessor.
 
 For [code]#accessor# and [code]#local_accessor#, this function may only be
 called from within a <<command>>.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6899,7 +6899,7 @@ This function may only be called from within a <<command>>.
 a@
 [source]
 ----
-accessor &operator=(const value_type &other) const
+const accessor &operator=(const value_type &other) const
 ----
    a@ Available only when
       [code]#(AccessMode != access_mode::atomic && Dimensions == 0)#.
@@ -6911,7 +6911,7 @@ This function may only be called from within a <<command>>.
 a@
 [source]
 ----
-accessor &operator=(const value_type &&other) const
+const accessor &operator=(const value_type &&other) const
 ----
    a@ Available only when
       [code]#(AccessMode != access_mode::atomic && Dimensions == 0)#.
@@ -7987,7 +7987,7 @@ accessor was constructed.  For other accessors, returns the default constructed
 a@
 [source]
 ----
-host_accessor &operator=(const value_type &other) const
+const host_accessor &operator=(const value_type &other) const
 ----
    a@ Available only when [code]#(Dimensions == 0)#.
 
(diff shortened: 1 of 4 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -6895,6 +6896,30 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
 
 This function may only be called from within a <<command>>.
 
+a@
+[source]
+----
+accessor &operator=(const value_type &other) const
```
gmlueck, line 6902: I think this needs to return a `const accessor &` because the member function is `const`.

tomdeakin:
LGTM, except this is adding a new API. Can we do that into SYCL 2020?

gmlueck:
> LGTM, except this is adding a new API. Can we do that into SYCL 2020?

I would put this in the category of "oversight" rather than "new functionality".  In it's current form, there is no way to write a value into a 0-dimension accessor.  Surely, that was not our intent!

gmlueck:
> Assume operator+=, etc follows from the operator= returning a reference?

We do not currently define any other operators like this.  I was also wondering about adding other operators while reviewing this, but I was concerned about making this PR too big as a "bug fix" against the spec.  It seems like `operator=` is the minimum you need to use a 0-dimensional accessor.  You can always do operations on the raw `T` value and then assign to the accessor.

If we consider adding `+=`, then we need to consider all the other `op=` operators as well as the non-assigning versions.  And then you need to consider the case when `T` doesn't support mathematical operators like `+`.

tomdeakin (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -6895,6 +6896,30 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
...
+      [code]#(AccessMode != access_mode::atomic && Dimensions == 0)#.
+
+Assignment to the single element that is accessed by this accessor.
+
+This function may only be called from within a <<command>>.
+
+a@
+[source]
+----
+const accessor &operator=(const value_type &&other) const
```
keryell, line 6914: Should we have instead
```suggestion
const accessor& operator=(value_type&& other) const
```
?

adoc/chapters/programming_interface.adoc
```diff
@@ -6895,6 +6896,30 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
 
 This function may only be called from within a <<command>>.
 
+a@
+[source]
+----
+const accessor &operator=(const value_type &other) const
```
keryell, line 6902:
```suggestion
const accessor& operator=(const value_type& other) const
```

adoc/chapters/programming_interface.adoc
```diff
@@ -7959,6 +7984,24 @@ If this is a <<ranged-accessor>>, returns the offset that was specified when the
...
+const host_accessor &operator=(const value_type &other) const
+----
+   a@ Available only when [code]#(Dimensions == 0)#.
+
+Assignment to the single element that is accessed by this accessor.
+
+a@
+[source]
+----
+const host_accessor &operator=(const value_type &&other) const
```
keryell, line 7999:
```suggestion
const host_accessor& operator=(value_type&& other) const
```
?

adoc/headers/accessorBuffer.h
```diff
@@ -153,9 +153,15 @@ class accessor {
   /* Available only when: (Dimensions > 0) */
   id<Dimensions> get_offset() const;
 
-  /* Available only when: (Dimensions == 0) */
+  /* Available only when: (AccessMode != access_mode::atomic && Dimensions == 0) */
   operator reference() const;
 
+  /* Available only when: (AccessMode != access_mode::atomic && Dimensions == 0) */
+  const accessor &operator=(const value_type &other) const;
```
keryell, line 160:
```suggestion
  const accessor& operator=(const value_type& other) const;
```

keryell (changes requested):
Coding style and too much constness.

fraggamuffin:
what to do with the type from +=? so prefer minimal change for now
adding api is ok

keryell (approved):
Thanks.

adoc/chapters/programming_interface.adoc
```diff
@@ -7959,6 +7984,24 @@ If this is a <<ranged-accessor>>, returns the offset that was specified when the
 accessor was constructed.  For other accessors, returns the default constructed
 [code]#id<Dimensions>{}#.
 
+a@
+[source]
+----
+const host_accessor &operator=(const value_type &other) const
```
keryell, line 7990:
```suggestion
const host_accessor& operator=(const value_type& other) const
```

adoc/chapters/programming_interface.adoc
```diff
@@ -8156,6 +8199,28 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
 
 This function may only be called from within a <<sycl-kernel-function>>.
 
+a@
+[source]
+----
+const local_accessor &operator=(const value_type &other) const
```
keryell, line 8205:
```suggestion
const local_accessor& operator=(const value_type& other) const
```

adoc/chapters/programming_interface.adoc
```diff
@@ -8156,6 +8199,28 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
...
+   a@ Available only when [code]#(Dimensions == 0)#.
+
+Assignment to the single element that is accessed by this accessor.
+
+This function may only be called from within a <<command>>.
+
+a@
+[source]
+----
+const local_accessor &operator=(const value_type &&other) const
```
keryell, line 8216:
```suggestion
const local_accessor& operator=(const value_type&& other) const
```

keryell (changes requested):
Almost there!

keryell (approved):
Thanks!

Issue #277, opened 2022-08-22
`Host Prefetch`


TApplencourt:
When using USM (`sycl::malloc_shared`), one may want to use `prefetch` to potentially transfer data from the host to the device. 

Are we missing the reverse operation (sending the data from the device to the host)? 

It exists in CUDA via [cudaMemPrefetchAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge8dc9199943d421bc8bc7f473df12e42):
> Passing in cudaCpuDeviceId for dstDevice will prefetch the data to host memory

And in OpenCL (kind of, it's stronger than a prefetch) via [clEnqueueMigrateMemObjects](https://registry.khronos.org/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueMigrateMemObjects.html) / clEnqueueSVMMigrateMem(https://registry.khronos.org/OpenCL/sdk/2.2/docs/man/html/clEnqueueSVMMigrateMem.html)

>CL_MIGRATE_MEM_OBJECT_HOST:  This flag indicates that the specified set of memory objects are to be migrated to the host, regardless of the target command queue.

TApplencourt:
It's good that you have an extension. That means it's important to support the feature :D

We may want to be future-proof to do USM from Device -> Device.  This is why I didn't go through the `prefetch_host()` strategy.
And also `sycl::host_selector{}` is now optional in SYCL2020 no?

illuhad:
Can we not express device->device in exactly the same way, by just having a prefetch operation submitted to a queue bound to the device to which you want to prefetch?
`prefetch_host()` is probably a more common scenario so it maybe makes sense to have a shortcut there.

> And also sycl::host_selector{} is now optional in SYCL2020 no?

Maybe, but it also works with `cpu_selector`, that's the same thing in hipSYCL ;)

TApplencourt:
>Can we not express device->device in exactly the same way, by just having a prefetch operation submitted to a queue bound to the device to which you want to prefetch?

![image](https://user-images.githubusercontent.com/6524907/186212267-f02303db-e8c0-4dfd-9dbc-50c2aa83b1f4.png)

You are right `prefetch_host` seems a natural way. I agree!

PR #280, opened 2022-08-29
Add prefetch_host

TApplencourt:
This is PR who follows up on #277,  and adds some `prefetch_host` API.

[ I don't think the addition of `prefetch_host` API was discussed by the committee, so sorry for opening the PR now. It's just that I got some free time and didn't want to forget about this ]

We have `prefetch` that allows the runtime to transfer data from the host to the device, but we are missing the reverse, transferring data from the device to the host. 
Cuda and OpenCL have the API. As usual ( :) ) HipSYCL have already an extension for this: https://github.com/illuhad/hipSYCL/blob/develop/doc/extensions.md#hipsycl_ext_prefetch_host #9

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3555,6 +3555,33 @@ a@ Equivalent to submitting a command-group containing
 a@
 [source]
 ----
+event prefetch_host(void* ptr, size_t numBytes)
+----
+a@ <<sec:usm, USM>>
+a@ Equivalent to submitting a command-group containing
+[code]#handler::prefetch_host(ptr, numBytes)#.
+a@
+[source]
+----
+event prefetch_host(void* ptr, size_t numBytes,
+               event depEvent)
+----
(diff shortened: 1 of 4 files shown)


gmlueck:
Should this be a proposed PR for SYCL-Next instead?  This repo is for fixes to SYCL 2020.

TApplencourt:
I can do that. This adds a new API call so that it may be better in SYCL2020 indeed.

gmlueck:
> Should I make the PR to a new SYCLnext branch that I create?

There is already a SYCL-Next branch on the internal GitLab repo.  It should be an MR there.

TApplencourt:
Closing it, will move to internal

TApplencourt:
I think it's still a good idea (just got beaten by that today). Need to check if I did follow up and opened a issue on the internal

TApplencourt:
Opened here https://gitlab.khronos.org/sycl/Specification/-/merge_requests/729

PR #283, opened 2022-09-02
Add table of known identity values

### LINKED ISSUE #163: Clarify for which types (has_)known_identity should be implemented
Which overloads of `sycl::reduction()` a client is allowed to call depends on whether `sycl::has_known_identity_v` is true for the argument types:

* `reduction(buffer<T>, handler &, Op, sycl::property::reduction::initialize_to_identity)` requires `has_known_identity_v<Op, T>`
* `reduction(buffer<T>, handler &, T identity, Op)` requires `!has_known_identity_v<Op, T>`

Therefore, to allow portability, the spec must state precisely for which operations and types `has_known_identity` should be provided out-of-the-box. The 2021-07-08 build talks about _standard binary operators_, but it is not clear if that is meant to apply to the function objects from section 4.17.2. Even if it does, the argument types are still not specified.

For reference, DPC++ currently implements the following pairs:

* `sycl::plus`, `sycl::multiplies` for integral and floating-point types 
* `sycl::bit_or`, `sycl::bit_and`, `sycl::bit_xor` for integral types
* `sycl::minimum`, `sycl::minimum` for integral and floating-point types, where identities are the largest/smallest representable number

IMO, the following should also be included:

* `sycl::logical_or` and `sycl::logical_and` for bool
* `sycl::plus`, `sycl::bit_{and,or,xor}`, `sycl::minimum` and `sycl::maximum` for built-in vector types

It's up for discussion whether the following should be added as well:

* `sycl::multiplies` for built-in vector types (decide if this should be a component-wise product?)

The implemented operation and argument types should either be listed exhaustively or described in terms of C++ type traits (i.e. `std::is_arithmetic_v`). For all types for which SYCL requires a specific representation, the identity values should be listed explicitly.

Pennycook:
Clarifies which combinations of BinaryOperator and AccumulatorT are
guaranteed to have known identities, and the values of those identities.

Closes #163.


adoc/chapters/programming_interface.adoc
```diff
@@ -12525,6 +12525,182 @@ be tested using the [code]#has_known_identity# trait class.
...
+sycl::minimum
+----
+a@
+[source]
+----
+std::is_floating_point_v<AccumulatorT> || std::is_same_v<std::remove_cv_t<AccumulatorT>, sycl::half>
+----
+a@
+----
+std::numeric_limits<AccumulatorT>::infinity()
```
psalz, line 12669: As pointed out by @fknorr in #163, the spec currently does not mandate that `std::numeric_limits` be specialized for `sycl::half`.

adoc/chapters/programming_interface.adoc
```diff
@@ -12525,6 +12525,182 @@ be tested using the [code]#has_known_identity# trait class.
 include::{header_dir}/identity.h[lines=4..-1]
```
gmlueck, line 12525: I wanted to  make this comment on the previous paragraph that starts "Certain implementations of reductions can be made more efficient ...", but GitHub won't let me comment on an unchanged paragraph.

I wonder if the intent would be more clear if this paragraph started out like:

> Some of the overloads for the `reduction` function take an identity value and some do not.  An implementation is required to compute a correct reduction even when the application does not specify an identity value.  However, the implementation may be more efficient when the identity value is either provided by the application or is known by the implementation.  [...]

Pennycook: I like this.  I've made the change in 954933ce3a9411fa7f21170803ff31d3da2b6622.  Since I was updating this paragraph, I also folded in @psalz's feedback from https://github.com/KhronosGroup/SYCL-Docs/issues/163#issuecomment-1240678038 to clarify that the identity depends on more than the binary operator alone.

adoc/chapters/programming_interface.adoc
keryell: > As pointed out by @fknorr in #163, the spec currently does not mandate that `std::numeric_limits` be specialized for `sycl::half`.

Perhaps we should?

keryell (approved):
Good clarification.

fraggamuffin:
should we fix numeric_limits for half now? (will require impl to change)or do it for SYCL NEXT?
Seems ok now.

fraggamuffin:
Can we also say all std traits work?

fraggamuffin:
http://eel.is/c++draft/numeric.limits.general#6

fraggamuffin:
Non-arithmetic standard types, such as [complex<T>](http://eel.is/c++draft/complex), shall not have specializations[.](http://eel.is/c++draft/numeric.limits.general#6.sentence-1)

Pennycook:
I've added the wording about `std::numeric_limits<sycl::half>`, but in a different place to where #163 suggested.

#163 suggested that it should be added to the [Half precision floating-point](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:extension-fp16) section,  but this is a subsection of the specification related to OpenCL.  The `std::numeric_limits` requirement should apply to all backends, so I've added the requirement to the table that first defines `sycl::half`.

Pennycook:
> Can we also say all std traits work?

I now think we should defer this to a later PR, so we can spend more time getting the language right.

The specification already makes reference to `std::numeric_limits<sycl::half>` in device code for some math functions (via `std::numeric_limits<genfloat>`, so this PR isn't introducing any new requirements, just clarifying them.

gmlueck (approved):

psalz (approved):

ProGTX (approved):

PR #285, opened 2022-09-09
Fix definitions of sycl::minimum and sycl::maximum

Pennycook:
The original definitions of sycl::minimum and sycl::maximum used std::lesser and std::greater, respectively.

These definitions unintentionally introduce differences from ISO C++:
- std::max() is defined in terms of operator <
- std::max_element() is defined in terms of operator <
- sycl::maximum requires users to define an additional operator (>)

This commit redefines both sycl::minimum and sycl::maximum in terms of operator <, to align with ISO C++ as intended.

Partially addresses internal issue 538.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19832,8 +19832,7 @@ a@
 ----
 T operator()(const T& x, const T& y) const
 ----
-   a@ Applies [code]#std::less# to its arguments, in the same order, then
-      returns the lesser argument unchanged.
+   a@ Returns [code]#y# if [code]#y < x# and #x# otherwise.
 
 |====
 
@@ -19847,8 +19846,7 @@ a@
 ----
 T operator()(const T& x, const T& y) const
 ----
-   a@ Applies [code]#std::greater# to its arguments, in the same order, then
-      returns the greater argument unchanged.
+   a@ Returns [code]#y# if [code]#x < y# and #x# otherwise.
 
 |====
 


TApplencourt (approved):

TApplencourt:
Should we also rename them, or make some alias, to `max` and `min` to mimic C++? 
But maybe confusing as we don't support all the `std` overload.

Pennycook:
> Should we also rename them, or make some alias, to `max` and `min` to mimic C++? But maybe confusing as we don't support all the `std` overload.

We already have `sycl::max()` and `sycl::min()`, which are functions that mimic `std::max()` and `std::min()`. We need different names to avoid this conflict, and to help people differentiate between function objects (e.g. `maximum`) and functions (e.g. `max`).

We decided to provide a function object because it allows us to express min/max reductions (and other algorithms) the same way as we would any other operator.  This isn't possible in ISO C++ without writing your own functor -- you can't pass `std::max` to a function like `std::accumulate`, because `std::max` is a function and the algorithms expect function objects.

FWIW: The names `maximum` and `minimum` are also aligned with frameworks like Thrust and oneDPL.

I'm not trying to argue that the names are perfect, just giving some background on why they are what they are.

BenBrock:
This is off topic, but it would be nice if there were `<functional>`-style versions of min/max that handled signed/unsigned comparison properly (i.e. with a specialization to use `std::cmp_less`/`std::cmp_greater` for `std::integral` types, along the lines of [Walter Brown's proposal](https://cppcon.digital-medium.co.uk/wp-content/uploads/2021/10/min-max-and-More.pdf)).

As for `sycl::minimum`, I agree that I don't really see the purpose of defining it to use `std::less` over `operator<`, since `std::less` is defined to use `operator<` except for some pointer types (or  potentially user-defined specializations).  (In fact, if it's the same as `std::less`, why not call is `sycl::less`?)

Pennycook:
> This is off topic, but it would be nice if there were <functional>-style versions of min/max that handled signed/unsigned comparison properly (i.e. with a specialization to use std::cmp_less/std::cmp_greater for std::integral types, along the lines of [Walter Brown's proposal](https://cppcon.digital-medium.co.uk/wp-content/uploads/2021/10/min-max-and-More.pdf)).

Thanks for the link to that presentation -- very interesting.

I think this is something we should explore as an extension before looking to change the SYCL specification.  I don't think we have many use cases yet (in SYCL) for computing the minimum and maximum across different types: functions like `sycl::reduce_over_group` typically see every work-item contribute the same data type.

> As for sycl::minimum, I agree that I don't really see the purpose of defining it to use std::less over operator<, since std::less is defined to use operator< except for some pointer types (or potentially user-defined specializations). (In fact, if it's the same as std::less, why not call is sycl::less?)

There's a subtle difference: `std::less` returns a `bool` representing the result of the comparison, whereas `sycl::minimum` returns the lesser of the two arguments.

If we later decide for some reason that functors calling `std::less` and `std::greater` instead of `operator <` would also be useful, I'd suggest adopting names like `least` and `greatest` to maintain the link while still differentiating the behavior.

fraggamuffin:
Check if this is consistent with ISO C++, min max

Pennycook:
> Check if this is consistent with ISO C++, min max

The definitions of `min` and `max` in http://eel.is/c++draft/alg.min.max say:
> Returns: The smaller value. Returns the first argument when the arguments are equivalent.
> Returns: The larger value. Returns the first argument when the arguments are equivalent.

@keryell, @nliber:  Should we use the wording above instead?  I think the mapping of the words "smaller" and "larger" to `<` is covered by http://eel.is/c++draft/alg.sorting, but it's not obvious to me when to use `(a < b)` vs `!(b < a)`.

keryell:
@Pennycook I have the feeling we should have the same semantics (and wording as an easy action) as in C++.
For the definition of _equivalent_ there is http://eel.is/c++draft/alg.sorting#general-8
But that means that in some case it requires 2 comparisons instead of 1.

Pennycook:
@keryell: Thanks for the link to _equivalent_.  I think that gives us clear direction here, and I've convinced myself that although C++ requires 2 comparisons to _prove_ equivalence, we can still get away with 1 here.  Here's my logic for `min` (the logic for `max` is the same, but flipped) -- please double check it!

- If `(b < a)` is `true`, then:
  - `!(a < b) && !(b < a)` is `unknown && false`; therefore, `a` and `b` are not equivalent
  - `b` is smaller
  - `min(a, b)` should return `b`
- If `(b < a)` is `false`, then:
  - `!(a < b) && !(b < a)` is `unknown && true`; therefore, `a` and `b` _may_ be equivalent (depending on `!(a < b)`)
  - `a` is smaller or `a` and `b` are equivalent
  - `min(a, b)` should return `a`

Implementing `min(a, b)` as `return (b < a) ? b : a` therefore meets the description "Returns: The smaller value. Returns the first argument when the arguments are equivalent."

Do you agree?  If so, I'll change the descriptions to match http://eel.is/c++draft/alg.min.max.

keryell (approved):
Astute!

ProGTX (approved):

PR #286, opened 2022-09-14
Fix buffer reinterpret for buffer_allocator with const

AidanBeltonS:
This PR fixes a mistake in the buffer reinterpret member function. Currently if the user reinterprets the buffer from type `T` to type `const T` then the buffer_allocator is updated to type `const T` this disagrees with the `buffer_allocators` description which states, "A buffer of data type const T uses buffer_allocator<T> by default". 
I am proposing that the buffer reinterpret function removes the `const` keyword for the `buffer_allocator` to keep it consistent with the classes description and the buffers definition which also removes any `const` keyword for the `buffer_allocator`.


keryell (approved):
Good catch.

fraggamuffin:
LGTM

gmlueck (approved):

PR #289, opened 2022-09-28
Add access::decorated::no to multi_ptr deduction guides

steffenlarsen:
In the current version of the SYCL 2020 specification the deduction guides for `multi_ptr` are missing the `access::decorated` template argument in the deduced type. This commit adds `access::decorated::no` to the deduced types.

### DIFF
--- a/adoc/headers/multipointer.h
+++ b/adoc/headers/multipointer.h
@@ -331,10 +331,10 @@ multi_ptr<ElementType, Space, DecorateAddress> address_space_cast(ElementType*);
 template <int Dimensions, access_mode Mode, access::placeholder IsPlaceholder,
           class T>
 multi_ptr(accessor<T, Dimensions, Mode, target::device, IsPlaceholder>)
-    -> multi_ptr<T, access::address_space::global_space>;
+    -> multi_ptr<T, access::address_space::global_space, access::decorated::no>;
 template <int Dimensions, access_mode Mode, access::placeholder IsPlaceholder,
           class T>
 multi_ptr(local_accessor<T, Dimensions>)
-    -> multi_ptr<T, access::address_space::local_space>;
+    -> multi_ptr<T, access::address_space::local_space, access::decorated::no>;
 
 } // namespace sycl


Naghasan:
I think it should be `no` for a default. It is not safe to expose an implementation define type by default, users should opt-in to this.

The rational is if we go forward with your proposal, then the following won't work with ComputeCpp and DPC++ when compiling in device mode (but would for the host mode)

```cpp
local_accessor<int, 2> Acc;
multi_ptr p{Acc};
static_assert(std::is_same_v<std::remove_pointer_t<decltype(p.get())>, int>);
````

Setting to `access::decorated::no`, this would be fine as `get` must return `int*`

steffenlarsen:
I am fine with either. We have implicit conversions between the decorations anyway, so if the user is explicit about the type, they should be able to control the decoration, so having it undecorated by default may be the more conservative solution indeed.

Naghasan (approved):

Naghasan:
The user `access::decorated` control the user facing API. Internally an implementation should store the pointer in the class as a decorated one regardless, that allows the compiler to carry the information as much as possible without exposing any C++ extension.

keryell (approved):
Thanks!

ProGTX:
This introduces some inconsistency because the base template doesn't specify a default (which is technically a breaking change compared to SYCL 1.2.1). Should we add a default there as well, and be the same as for the deduction guides? I'm also not sure that there is one correct answer for the default, maybe the default needs to be implementation defined.

steffenlarsen:
> This introduces some inconsistency because the base template doesn't specify a default (which is technically a breaking change compared to SYCL 1.2.1). Should we add a default there as well, and be the same as for the deduction guides? I'm also not sure that there is one correct answer for the default, maybe the default needs to be implementation defined.

I think if we want to avoid breaking 1.2.1 we would need to make the default decoration in `multi_ptr` to be `legacy` as otherwise user-code may be using members that have changed or are unavailable in non-legacy `multi_ptr`. However, I don't think having `legacy` in the deduction guides is a good idea.

Naghasan:
> Should we add a default there as well, and be the same as for the deduction guides?

This was discussed back in end of 2020, the consensus was to not provide a default.

PR #291, opened 2022-10-04
Small clarification to buffer sync rules

gmlueck:
The section on buffer synchronization rules define the scenarios when the buffer destructor must block vs. the scenarios where it does not block.  The old wording made it seem like a buffer created from no host pointer is only non-blocking if it uses the default memory allocator. I assume this was not our intent, and such a buffer should have a non-blocking destructor regardless of the memory allocator.

The main change in this PR is to remove the phrase "and using the default buffer allocator".  However, I also reworded a bit to clarify which type of buffers we are talking about.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4162,6 +4162,19 @@ host. Note that in this case the runtime will raise an error if it requires
 host memory but it is not available (e.g when moving data across <<backend>>
 contexts).
 
+In some cases, the implementation may retain a copy of the allocator object
+even after the buffer is destroyed.  For example, this can happen when the
+buffer object is destroyed before commands using accessors to the buffer have
+completed.  Therefore, the application must be prepared for calls to the
+allocator even after the buffer is destroyed.
+
+[NOTE]
+====
+If the application needs to know when the implementation has destroyed all
+copies of the allocator, it can maintain a reference count within the
+allocator.
+====
+
 The definition of allocators extends the current functionality of SYCL,
 ensuring that users can define allocator functions for specific hardware or
 certain complex shared memory mechanisms (e.g. NUMA), and improves
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4986,10 +4986,10 @@ More precisely:
   . A buffer can be constructed from a [code]#hostData# pointer.  The buffer
     will use this host memory for its full lifetime, but the contents of this
     host memory are unspecified for the lifetime of the buffer. If the host
-    memory is modified by the host, or mapped to another buffer or image during
-    the lifetime of this buffer, then the results are undefined. The initial
-    contents of the buffer will be the contents of the host memory at the time
-    of construction.
+    memory is modified by the host or if it is used to construct another
+    buffer or image during the lifetime of this buffer, then the results are
+    undefined.  The initial contents of the buffer will be the contents of the
+    host memory at the time of construction.
 +
 --
 When the buffer is destroyed, the destructor will block until all
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4976,20 +4976,20 @@ with attached host memory and is still in use.
 
 More precisely:
 
-  . A buffer can be constructed with just a size and using the default
-    buffer allocator. The memory management for this type of buffer is
-    entirely handled by the SYCL system. The destructor for this type of
+  . A buffer can be constructed from a [code]#range# (and without a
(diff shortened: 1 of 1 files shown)


gmlueck:
We do not have any issue about this, so we haven't discussed whether this is the correct interpretation of the spec.  Our implementation team raised a question about how to implement buffer destructors, though, and I think this PR reflects what we probably intended.  Feel free to object if you think this is not what we meant.

gmlueck:
Our implementation team raised a counter argument: What if the allocator is user-supplied and stateful.  For example, consider an allocator that returns blocks of memory from a pool.  If the buffer destructor doesn't block, how does the application know when there are no more buffers that are using the pool?

If we wanted to preserve the "does not block" semantic, I think we could add some guarantee that limits the lifetime of the runtime's use of the allocator.  For example, we could say that any memory allocated for a buffer is guaranteed to be deallocated by the time all commands with accessors to that buffer have completed.

There is also another option, which is a somewhat larger change.  I've always thought that the rules about when the buffer blocks are pretty complex.  We could simplify this and say that the buffer destructor always blocks until all commands with accessors to that buffer have completed.  This seems like a much simpler rule, which is easy to understand.  It also addresses the outstanding issue in #240.

keryell (approved):
Thanks for the clarification.

fraggamuffin:
add a stateful allocator to track the state? add wording

adoc/chapters/programming_interface.adoc
```diff
@@ -4976,20 +4976,20 @@ with attached host memory and is still in use.
...
-    buffer allocator. The buffer will use this host memory for its full
-    lifetime, but the contents of this host memory are unspecified for the
-    lifetime of the buffer. If the host memory is modified by the host, or
-    mapped to another buffer or image during the lifetime of this buffer,
-    then the results are undefined. The initial contents of the buffer will
-    be the contents of the host memory at the time of construction.
+  . A buffer can be constructed from a [code]#hostData# pointer.  The buffer
+    will use this host memory for its full lifetime, but the contents of this
+    host memory are unspecified for the lifetime of the buffer. If the host
+    memory is modified by the host, or mapped to another buffer or image during
```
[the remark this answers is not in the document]
gmlueck: I think that statement is just saying that you cannot construct another buffer from the same `hostData` pointer.  I agree that the word "mapped" is a bit confusing, though.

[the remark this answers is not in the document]
gmlueck: I clarified the wording here in 87a71b125e396b8f55a349ee0ddae53f902eed11

gmlueck:
> add a stateful allocator to track the state? add wording

I added some information about the use of buffer allocators in f4f70c77f0b238d42b9b339cfd1c60a1b4f2b8e9.

I think this PR is ready now, so please re-review.

adoc/chapters/programming_interface.adoc
```diff
@@ -4976,20 +4989,20 @@ with attached host memory and is still in use.
...
     contents of the buffer are unspecified.
-  . A buffer can be constructed with associated host memory and a default
-    buffer allocator. The buffer will use this host memory for its full
-    lifetime, but the contents of this host memory are unspecified for the
-    lifetime of the buffer. If the host memory is modified by the host, or
-    mapped to another buffer or image during the lifetime of this buffer,
-    then the results are undefined. The initial contents of the buffer will
-    be the contents of the host memory at the time of construction.
+  . A buffer can be constructed from a [code]#hostData# pointer.  The buffer
+    will use this host memory for its full lifetime, but the contents of this
```
keryell, line 5000: Just curious about plural of `contents`. In French I would use singular, so just checking. :-)

adoc/chapters/programming_interface.adoc
```diff
@@ -4976,20 +4989,20 @@ with attached host memory and is still in use.
...
+    host memory are unspecified for the lifetime of the buffer. If the host
+    memory is modified by the host or if it is used to construct another
```
keryell, line 5002:
```suggestion
    memory is modified on the host or if it is used to construct another
```
to avoid weird case where an RDMA from the outer-space is changing the value.

keryell (approved):
Good clarification.

adoc/chapters/programming_interface.adoc
gmlueck: This is complicated in English!  We use "content" for things that are uncountable and "contents" for things that are countable.  It's not entirely clear whether it should be "content of memory" or "contents of memory".  When I do a Google search of each, I see there are about the same number of occurrences of each.  It seems like the SYCL spec use "contents of memory" in many places, so I think we should use that consistently here.

ProGTX (approved):

TApplencourt (approved):

PR #292, opened 2022-10-05
Run Clang-format on AsciiDoctor source blocks

TApplencourt:
Follow-up on #269.

>The changes in the specification itself is manual work because of the intermixing of AsciiDoctor and C++ snippets but clang-format was used for changing the headers and code samples.

I developed a little script that allowed me to automatically run clang-format on the AsciiDoctor source block (see below). Current limitations are:
- Alignment of enumerations is not conserved (see `has_kernel`, for example). 
- Some functions declarations are not valid C++, so we ignore them for now (see the multiples `genfloatf` for example)

If you don't think the PR is useful, please feel free to close it.

Code:
```ruby
#!/usr/bin/env ruby
require 'open3'
require 'digest'

def fix_file(file)
  str = File.open(file) { |f| f.read }
  str_fixed = str.gsub(/(\[source\]\n----$)(.*?)(^----$)/m)  { |s|
    # Some functions are declared using a non-valid C++ syntax. Such as:
      #[source]
      #----
      #genfloatf nan (ugenint nancode)
      #genfloatd nan (ugenlonginteger nancode)
      #----
    # So we avoid them for now.
    next s unless $2.include?(";")
    o, s = Open3.capture2("clang-format", :stdin_data=>$2)
    "#{$1}#{o}#{$3}"
  }
  File.open(file, "w") { |f| f.write str_fixed } if Digest::MD5.digest(str) != Digest::MD5.digest(str_fixed)
end

ARGV.each { |f|
  p f
  fix_file(f)
}
```
To run:
```bash
ruby fix_code_inline.rb adoc/chapters/*
```

### DIFF
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -353,8 +353,7 @@ kernel
 a@
 [source]
 ----
-template <bundle_state State>
-kernel_bundle<State>
+template <bundle_state State> kernel_bundle<State>
 ----
    a@ [code]#cl_program#
    a@ [code]#std::vector<cl_program>#
@@ -482,17 +481,15 @@ The OpenCL backend provides the following specializations of the
 a@
 [source]
 ----
-context make_context(
-    const cl_context& clContext,
-    const async_handler& asyncHandler = {})
+context make_context(const cl_context& clContext,
+                     const async_handler& asyncHandler = {})
 ----
    a@ Constructs a SYCL [code]#context# instance from an OpenCL [code]#cl_context# in accordance with the requirements described in <<sec:backend-interoperability>>.
 a@
 [source]
 ----
-event make_event(
-    const std::vector<cl_event>& clEvents,
-    const context& syclContext)
+event make_event(const std::vector<cl_event>& clEvents,
+                 const context& syclContext)
 ----
    a@ Constructs a SYCL [code]#event# instance from a vector of OpenCL
       [code]#cl_event# objects in accordance with the requirements described in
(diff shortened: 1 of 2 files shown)


keryell:
Can you remove the `[draft]`?

TApplencourt:
Thanks for the nice feedback! 

I first will enhance the script to handle the one-line code block without ";", and then remove [draft]. 
I'm in meeting all day long, this is the most productive work I can do...

@ProGTX  I will rewrite it in Python. AsciiDoc is written in ruby, but Python will be more maintainable for the committee indeed.

keryell:
> I will rewrite it in Python. AsciiDoc is written in ruby, but Python will indeed be more maintainable for the committee.

I had to learn Ruby for the publication of SYCL 2020 to adapt the `rouge` source-code highlighter to SYCL. I was not fluent and I have forgotten most of it since... :-(

TApplencourt:
I improved the script to be able to handle more cases. `2 files changed, 814 insertions(+), 1104 deletions(-)`.
As usual, it was more tricky than anticipated. But it seems to work (famous last word).

The Python script is bellow. I will make a PR when you think the script is ready:

```python
#!/usr/bin/env python3

def is_a_valid_block(body):
    # It's easier to detect code that is NOT an enumeration
    conditions = (
        ";" in body,  # `;` mean a valid C++ code
        body.count("(") == 1,  # If we have one and only one `foo()` it's not an enumeration by construction
        "template" in body,  # We don't enumerate templates (luck)
    )
    return any(conditions)


def run_clang_format(matchobj):
    import subprocess

    header, body, footer = matchobj.groups()

    # Some functions are enumerated using a non-valid C++ syntax. Such as:
    # [source]
    # ----
    # genfloatf nan (ugenint nancode)
    # genfloatd nan (ugenlonginteger nancode)
    # ----
    # or
    # [source]
    # ----
    # info::device::preferred_vector_width_char
    # info::device::preferred_vector_width_short
    # info::device::preferred_vector_width_int
    # ----

    # We cannot parse the block directly, as clang-format will produce ugly code.
    # We work-around it to clang-format each line individually

    # Identifying those Particular/Problematic blocks is a little challenging
    # See "is_a_valid_block"

    bodies = [body] if is_a_valid_block(body) else body.splitlines()
    # Encoding require Python 3.6
    bodies_formatted = (
        subprocess.check_output("clang-format", input=b, encoding="ascii") for b in bodies
    )
    body_formatted = "\n".join(bodies_formatted)
    return f"{header}{body_formatted}{footer}"


def parse_file(file):
    import re

    regex = r"""(^\[source\]\n----\n) # Header
                (.*?)                 # Body
                (\n----$)             # Footer"""

    with open(file) as f:
        data = f.read()

    data_formated = re.sub(
        regex, run_clang_format, data,
        flags=re.MULTILINE | re.DOTALL | re.VERBOSE
    )

    if data == data_formated:
        return

    with open(file, "w") as f:
        f.write(data_formated)

if __name__ == "__main__":
    import sys
    for file in sys.argv[1:]:
        print(f"Parsing {file}")
        parse_file(file)
```

TApplencourt:
I fixed the merge conflict and rerun the script. It should be good to go (famous last word).

PR #296, opened 2022-10-12
Buffer from container always blocks in destructor

gmlueck:
Clarify the behavior of the buffer destructor for a buffer that is created from a container.  The previous wording said that the destructor blocks if the container is non-const, but it didn't say what happens for a const container.  The new wording clarifies that it blocks in this case also.

This is consistent with the behavior of buffers that are constructed from a host pointer.  In that case, the destructor always blocks, even when host pointer is const.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5052,11 +5052,10 @@ lifetime, and the contents of this memory are unspecified for the
 lifetime of the buffer. If the container memory is modified by the host
 during the lifetime of this buffer, then the results are undefined.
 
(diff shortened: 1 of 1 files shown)


gmlueck:
As with my other recent PRs, there is no issue associated with this, so we haven't discussed it yet.  Feel free to object if you don't agree.

tomdeakin (approved):
This looks reasonable - I think an implementation would *need* to create a copy of the container data just for this one case without this change, which it wouldn't be doing for any other case.

keryell (approved):

ProGTX (approved):

PR #298, opened 2022-10-13
Move description of shared_ptr buffer destructor

gmlueck:
The buffer destructor has special rules about blocking when the buffer is constructed from a `shared_ptr`.  Previously, these rules were described in section 4.7.4.3 "Shared SYCL ownership of the host memory". However, the behavior of the buffer destructor in other cases is described in section 4.7.2.3 "Buffer synchronization rules".  This PR moves the description to 4.7.2.3 in order to keep the descriptions of the buffer destructor all in one place.

After doing this, I realized that section 4.7.4.3 did not contain any new information that wasn't already described in 4.7.2.3.  Therefore, I removed that section.

I also tightened the specification a bit for the case when a buffer is constructed from a `shared_ptr`:

* If the underlying type of the `shared_ptr` is `const`, then I assume we want the buffer to be read-only, just as in the raw pointer case.

* I also assume that the contents of the host memory referenced by the `shared_ptr` is undefined during the lifetime of the buffer, just as in the raw pointer case.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5029,20 +5029,30 @@ copy of data back to host.
 When the buffer is destroyed, the destructor will block until all work
 in queues on the buffer have completed.
 --
-  . A buffer can be constructed using a [code]#shared_ptr# to host
-    data. This pointer is shared between the SYCL application and the
-    runtime. In order to allow synchronization between the application and
-    the runtime a [code]#mutex# is used which will be locked by the
-    runtime whenever the data is in use, and unlocked when it is no longer
-    needed.
+  . A buffer can be constructed using a [code]#std::shared_ptr# to host
+    data.  This is similar to the case when the buffer is constructed using a
+    raw [code]#hostData# pointer, except that the SYCL runtime holds a
+    reference count to the [code]#shared_ptr# for the lifetime of the buffer.
+    Thus, the application can release its [code]#shared_ptr# even before the
+    buffer is destroyed.  The buffer will use this host memory for its full
+    lifetime, but the contents of this host memory are unspecified for the
+    lifetime of the buffer.  If the host memory is modified on the host or if
+    it is used to construct another buffer or image during the lifetime of this
+    buffer, then the results are undefined.  The initial contents of the buffer
+    will be the contents of the host memory at the time of construction.
 +
 --
-The [code]#shared_ptr# reference counting is used in order to prevent
-destroying the buffer host data prematurely. If the [code]#shared_ptr#
-is deleted from the user application before buffer destruction, the buffer
-can continue securely because the pointer hasn't been destroyed yet. It will
-not copy data back to the host before destruction, however, as the
-application side has already deleted its copy.
+When the buffer is destroyed, the destructor is guaranteed to block if the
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -6057,57 +6067,6 @@ buffer construction.
...
-  {
-    buffer<int, 1> b { ptr, range<2>{ 10, 10 } };
-    // update the data
-    [...]
-    ptr.reset();
-  } // Data is not copied back, there is no user side shared_ptr.
-}
-----
-
-
```
keryell, line 6110: Should we keep these examples?

keryell (commented):
Globally that sounds like a good simplification and clarification.

adoc/chapters/programming_interface.adoc
```diff
@@ -6057,57 +6067,6 @@ buffer construction.
...
-[source,,linenums]
-----
-{
-  std::shared_ptr<int> ptr { data };
-  {
-    buffer<int, 1> b { ptr, range<2>{ 10, 10 } };
-    // update the data
-    [...]
-    ptr.reset();
```
keryell, line 6105: This line require a comment, for casual readers not using routinely `std::shared_ptr`.

adoc/chapters/programming_interface.adoc
keryell: I have checked again : these were the only 2 code examples showing `sycl::buffer` and `std::shared_ptr`. Why removing them? Are they wrong now?

gmlueck: I agree that the examples are good to keep.  I was hoping to move all of the normative description of the buffer APIs to [section 4.7.2 Buffers](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:buffers), and in particular the normative description of the blocking behavior and copy-back behavior to [section 4.7.2.3 Buffer synchronization rules](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:buf-sync-rules).

How do you feel about the following ... I'll still augment section 4.7.2.3 to contain a complete description of the blocking behavior and copy-back behavior, so all of that information is together in one place.  We can still keep section 4.7.4, though, as a set of examples.  In this case, I'd probably retitle the section to something like "Buffer examples", and the introduction of that section will read something like:

> This section provides some examples showing typical use cases for buffer. These examples are intended to clarify the definition of the buffer interfaces, but the content of this section is nonnormative.

Over time, we can add more examples to this section to clarify other behavior about buffers.

gmlueck: I restored the examples, and converted section 4.7.4 to an examples section as I describe in the comment above.  See commit 15e8eab47d8198b1bb263da991bf5a7e80990505.

Note that this PR is incomplete at this point because I need answers to the questions in (https://gitlab.khronos.org/sycl/Specification/-/issues/478)

keryell (commented):
Great clarification!

keryell:
Dedicate some time like 1 hour to solve this after some preparing time.

fraggamuffin:
A good F2F

PR #300, opened 2022-10-18
Align discussion of forward progress with C++

Pennycook:
The commits here implement the clarifications and bug fixes that were discussed during the F2F.

A quick summary of the changes below:

- Align with terminology like "concurrent forward progress guarantees"
- Clarify the minimum requirements for SYCL implementations regarding work-item forward progress
- Clarify when commands are guaranteed to start and make progress

Addresses internal issues 311 and 627.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1114,15 +1114,6 @@ arrives at a group barrier acting on group _G_, implementations must eventually
 select and potentially strengthen another work-item in group _G_ that has not
 yet arrived at the barrier.
 
-Synchronizing work-items via memory operations is unsafe in general, but is
-supported if and only if the following conditions are true:
-
-  * acquire-release or sequentially consistent memory ordering is supported
-    at the scope containing the set of work-items being synchronized;
-
-  * the work-items being synchronized both provide concurrent forward progress
-    guarantees.
-
 When a host thread blocks on the completion of a command previously submitted
 to a SYCL queue (for example, via the [code]#sycl::queue::wait# function), it
 blocks with forward progress guarantee delegation.
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1099,7 +1099,7 @@ to the buffer or USM allocation from device code.
 ==== Forward progress
 
 This section, and any subsequent section referring to progress guarantees, uses
-the following terms as defined in the {cpp}17 specification: thread of
+the following terms as defined in the {cpp} core language: thread of
 execution; weakly parallel forward progress guarantees; parallel forward
 progress guarantees; concurrent forward progress guarantees; and block with
 forward progress guarantee delegation.
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1098,6 +1098,12 @@ to the buffer or USM allocation from device code.
 
 ==== Forward progress
 
+This section, and any subsequent section referring to progress guarantees, uses
+the following terms as defined in the {cpp}17 specification: thread of
+execution; weakly parallel forward progress guarantees; parallel forward
+progress guarantees; concurrent forward progress guarantees; and block with
+forward progress guarantee delegation.
+
 Each work-item in SYCL is a separate thread of execution, providing at least
 weakly parallel forward progress guarantees.  Whether work-items provide
 stronger forward progress guarantees is implementation-defined.  All
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1123,6 +1123,14 @@ When a host thread blocks on the completion of a command previously submitted
 to a SYCL queue (for example, via the [code]#sycl::queue::wait# function), it
 blocks with forward progress guarantee delegation.
 
+[NOTE]
+====
+SYCL commands submitted to a queue are not guaranteed to begin executing until
+a host thread blocks on their completion.  In the absence of multiple host
+threads, there is no guarantee that host and device code will execute
+concurrently.
+====
+
 // Later, this label will move onto a new subsection - see below
 [[sec:progmodel.cpp]]
 == The SYCL programming model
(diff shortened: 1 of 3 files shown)


adoc/chapters/architecture.adoc
```diff
@@ -1099,23 +1098,42 @@ to the buffer or USM allocation from device code.
...
 Synchronizing work-items via memory operations is unsafe in general, but is
 supported if and only if the following conditions are true:
 
   * acquire-release or sequentially consistent memory ordering is supported
     at the scope containing the set of work-items being synchronized;
 
-  * the work-items being synchronized are guaranteed to make forward progress
-    with respect to one another.
+  * the work-items being synchronized both provide concurrent forward progress
+    guarantees.
```
mkinsner, line 1124: Some interpretations of "synchronize" probably don't require concurrent forward progress guarantees.  Should this be strengthened to "synchronizes with", and the term imported from C++ above in [L1101-1105]?(https://github.com/KhronosGroup/SYCL-Docs/pull/300/files#diff-3211f50334dd66e4638c43c11bb1b9dc16f0c26a8b7e5af104ea4e9f7b1744ecR1101-R1105)

Pennycook: You're right, this section is unclear.  Reading this again, I think we might even be conflating two concepts.

The first bullet point definitely means "synchronizes with". A device needs to support acquire-release ordering or sequential consistency in order to have atomics from different work-items synchronize with one another.

The second bullet point is a more colloquial usage of "synchronized", and I think it's actually talking about the situation where one work-item attempts to synchronize with another using an atomic operation that is not "lock-free".  It's the blocking nature of the atomic operation that requires strong forward progress guarantees.

I'm not sure what to do here.  If we import "synchronizes with" and "lock-free" we potentially have to rewrite large parts of the specification (e.g. the section currently titled "Synchronization"!).  We might want to do that eventually, but it seems like something we could tackle separately.

Another option would be to drop these bullet points from the specification temporarily and reintroduce them later. Since they're only offering guidance to developers, it may make more sense to have them as non-normative notes in a different section.

Pennycook: I've drafted a change in https://github.com/KhronosGroup/SYCL-Docs/pull/300/commits/3a5e79070b21d6a6d4a215838bf2aebbdd9be386 to show what the last option would look like.  I think this fixes the immediate issue of the paragraph you pointed to being confusing.

I'll open an internal issue to discuss "synchronizes-with" further.

adoc/chapters/architecture.adoc
```diff
@@ -1099,23 +1098,33 @@ to the buffer or USM allocation from device code.
 
 ==== Forward progress
 
-Memory consistency guarantees are independent of any forward progress
-guarantees.  A SYCL implementation must execute work-items concurrently
-and must ensure that the work-items in a group obey the semantics of
-<<group-barrier,group barriers>>, but are not required to provide any
-additional forward progress guarantees.
+This section, and any subsequent section referring to progress guarantees, uses
+the following terms as defined in the {cpp}17 specification: thread of
```
gmlueck, line 1102:
```suggestion
the following terms as defined in the {cpp} core language: thread of
```

We state in section 3.3 "Normative references" that the SYCL specification refers to C++17 as the "C++ core language".  Doing it this way avoid the need to update lots of references in our spec when we move to a newer version of C++.

Pennycook: Thanks, Greg.  Fixed!

TApplencourt (approved):

adoc/chapters/architecture.adoc
```diff
@@ -1099,23 +1098,33 @@ to the buffer or USM allocation from device code.
...
+Each work-item in SYCL is a separate thread of execution, providing at least
+weakly parallel forward progress guarantees.  Whether work-items provide
+stronger forward progress guarantees is implementation-defined.
 
-  * acquire-release or sequentially consistent memory ordering is supported
-    at the scope containing the set of work-items being synchronized;
+All implementations must additionally ensure that the work-items in a group
+obey the semantics of <<group-barrier,group barriers>>: when a work-item
+arrives at a group barrier acting on group _G_, implementations must eventually
+select and potentially strengthen another work-item in group _G_ that has not
```
tomdeakin, line 1114:
```suggestion
select and potentially strengthen the progress guarantees of another work-item in group _G_ that has not
```

tomdeakin: My motivation for this comment is SYCL programmers might not be experts in the C++ forward progress, but we should try to give as many helpful hints here to make it easy to read the spec, write SYCL, and dive into C++ later if they want.

tomdeakin (changes requested):
Minor point, but otherwise it think this is a good clarification.

adoc/chapters/architecture.adoc
Pennycook: I agree with the comment about readability, but I'm less sure about the fix.

I modelled this on the language used by the C++ draft, which always talks about strengthening a thread of execution, rather than strengthening a thread of execution's progress guarantees.  I'd like to try and avoid introducing minor differences in wording if we can, just to be safe.

If we rewrote this paragraph to explain the semantics of group barriers in words, would it be clear enough?  Something like:

> All implementations must additionally ensure that a work-item arriving at a
> <<group-barrier, group barrier>> does not prevent other work-items in the same
> group from making progress. When a work-item arrives at a group barrier acting
> on group _G_, implementations must eventually select and potentially strengthen
> another work-item in group _G_ that has not yet arrived at the barrier.

tomdeakin: Fair enough - better to follow the C++ wording. The new paragraph looks fine to me too.

Pennycook: Sorry, it's just I've been burnt enough times by unintentionally misusing words like "synchronize" that it's made me paranoid. 😆 

I've put the new paragraph in 386749ce9ce4b372d7e574647c78955a1dd01d10.  Thanks again for pointing out the readability/accessibility issue here.

tomdeakin (approved):

keryell (approved):
Great clarification!

PR #301, opened 2022-10-18
Clarify that command groups enqueue commands

Pennycook:
Previous wording suggested that only device kernels were enqueued, and that other commands (e.g. data transfers, host tasks) were not scheduled according to their data dependencies.


gmlueck (approved):

tomdeakin (approved):
After some soul searching in the Glossary, I think this is correct use of the word "command".

TApplencourt (approved):

keryell (approved):
Thanks.

PR #305, opened 2022-10-26
Fix description of `shift_group_right`

Melirius:


TApplencourt:
Nice find!  Maybe a naive question but can't we just say `can be lower than 0`?
I was under the impression that `min(get_linear_id()) == 0`

Melirius:
Yes, it is even simpler.

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -20154,8 +20154,8 @@ group.
 
 _Returns:_ the value of [code]#x# from the work-item whose group local id
 ([code]#id#) is [code]#delta# smaller than that of the calling work-item.
```
keryell, line 20156: Just looking at the sentence, is this concept of `delta larger` or `delta smaller` a well-known mathematical concept or could we make this clearer by using an equality instead?

tomdeakin: Resolve this consistently in another PR

keryell (approved):
Looks good for now.

ProGTX (approved):

PR #306, opened 2022-10-26
Fix description of the half precision math functions

Melirius:
AFAIK "ULP value" has a different meaning. :)

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -21218,7 +21218,7 @@ In SYCL the half precision math functions are defined in
 [code]#sycl::half_precision#. The functions that are
 available within this namespace are specified in
 <<table.half.math.functions>>. These functions are
-implemented with a minimum of 10-bits of accuracy i.e. an ULP value is
+implemented with a minimum of 10-bits of accuracy i.e. the maximum error is
 less than or equal to 8192 ulp.
 
 


TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -21218,7 +21218,7 @@ In SYCL the half precision math functions are defined in
 [code]#sycl::half_precision#. The functions that are
 available within this namespace are specified in
 <<table.half.math.functions>>. These functions are
-implemented with a minimum of 10-bits of accuracy i.e. an ULP value is
+implemented with a minimum of 10-bits of accuracy i.e. the maximum error is
```
keryell, line 21221: Curious sentence indeed!
I wonder where it does come from.
Assuming we stick on this 10 bits with rounding to nearest, I guess it would be `(1 << 10)/2` ulp, which is 512 ulp.

Melirius: The point, if I understand it correctly, is that they return `float`, not `half`. And half precision is something like (1 << 13) ulp of `float`.

keryell (approved):
Thanks!
Once a year on average I rediscover we have this weird maths functions that looks like `float` but have the taste of `half`...

tomdeakin:
Ready to merge

tomdeakin:
Need to open similar issue against OpenCL @Melirius please!

Melirius:
Made [OpenCL PR](https://github.com/KhronosGroup/OpenCL-Docs/pull/860)

Issue #312, opened 2022-10-31
`sycl::cl_int` and/or `sycl::opencl::cl_int`?


TApplencourt:
In [C6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl) we define `sycl::cl_int` and `sycl::cl_size_t`.

> Some types in SYCL vary according to pointer size or vary on the host according to the host ABI, such as size_t or long. In order for the SYCL device compiler to ensure that the sizes of these types match the sizes on the host and to enable data of these types to be shared between host and device, the OpenCL interoperability types are defined, sycl::cl_int and sycl::cl_size_t.

But in [C.7.8.](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_data_types_2) we said that `cl_int` is also part of `sycl::opencl`.

>For the purpose of interoperability and portability, SYCL defines a set of aliases to C++ types within the sycl::opencl namespace using the cl_ prefix. These aliases are described in [Table 192](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.types.aliases)

And in Table 192, we have `cl_int` but not `cl_size_t`. 

- Is that expected? 
- Or should we move all the `cl_` type into the `sycl::opencl` namespace? ( and declare `cl_size_t` in this namespace)

Naghasan:
> I don't see why cl_int and cl_size_t are special. 

`cl_int` is guarantee to be 32 bits, `cl_long` 64 etc. using `std::int32_t` would achieve a similar result but at least it is not optional. Legacy of the opencl times :)

`cl_size_t` is a bit more tricky, it is tied to the bitness used by the device. Can be useful for some (crazy) platform that uses a 64 bits host and a 32 bits device or 32/64. So for a 64/32 configuration, you would typically have `size_t` being 64 bits but `cl_size_t` 32 bits.

jbrodman:
Nobody should ever use the cl_ types.

If you want a std::int32_t - use a std::int32_t.  For std::size_t - why can't we use std::size_t?

TApplencourt:
To recap, the "consensus " seems to be:

- Deprecate `sycl::cl_int` and `sycl::cl_size_t` and move then into the `sycl::opencl` namespace.
[...]

PR #314, opened 2022-11-01
Simplify opencl::cl_ type and sizeof(T)

TApplencourt:
This addresses concerns raised in #312.

1. It removes a sentence in [C.6. Interoperability with the OpenCL API](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl). 

   - This sentence contained a typo (it should have been `sycl::opencl::cl_int`)
   - It's the only place on the spec where `sycl::cl_size_t` is defined. The OpenCL spec doesn't even define an  `OpenCL size_t`. We reached the conclusion that such a type was not needed.
   - This sentence is paraphrasing [C.7.8. Data types](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_data_types_2) 

> The OpenCL C language standard [Section 6.11](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#opencl12) defines its own built-in scalar data types, and these have additional requirements in terms of size and signedness on top of what is guaranteed by ISO C++. For the purpose of interoperability and portability, SYCL defines a set of aliases to C++ types within the sycl::opencl namespace using the cl_ prefix. These aliases are described in [Table 192](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.types.aliases)

2. It removes a paragraph in [5.5. Built-in scalar data types](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:scalartypes).  Indeed this paragraph is more confusing than helpful.  The type size should be identical between the host and the device.

### DIFF
--- a/adoc/chapters/device_compiler.adoc
+++ b/adoc/chapters/device_compiler.adoc
@@ -219,24 +219,6 @@ The standard {cpp} fixed width types, e.g. [code]#int8_t#,
 should have the same size as defined by the {cpp} standard for host and
 device.
 
-ifdef::showtodos[]
-[NOTE]
-.Note
-====
-To ensure that pointer types and [code]#size_t# use the same amount of
-storage on host and device when inside structures, SYCL also requires that
-device compilers use the host sizes for pointers or [code]#size_t#. Devices
-may be using a smaller size internally for pointers and [code]#size_t#,
-but this should not impact the programming model for users. In the case where
-devices use larger pointer size internally than the host, then
-[code]#size_t# will match the <<host-pointer>> size. For structures shared
-between host and device it is recommended to use fixed width types for pointers.
-Scalar data types for a SYCL device compiler are described in
-<<table.types.fundamental>>.
-====
-endif::showtodos[]
-
-
 
 [[table.types.fundamental]]
 .Fundamental data types supported by SYCL
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -443,13 +443,6 @@ whether the object has actively been used on these devices yet or not.
 
 // From 3.10 Language restrictions in kernels
 
-Some types in SYCL vary according to pointer size or vary on the host
-according to the host ABI, such as [code]#size_t# or [code]#long#. In order
-for the SYCL device compiler to ensure that the sizes of
-these types match the sizes on the host and to enable data of these types
-to be shared between host and device, the OpenCL interoperability types
-are defined, [code]#sycl::cl_int# and [code]#sycl::cl_size_t#.
-
 The OpenCL C function qualifier [code]#+__kernel+# and the access
 qualifiers: [code]#+__read_only+#, [code]#+__write_only+# and [code]#+__read_write+#
 are not exposed in SYCL via keywords, but are instead encapsulated in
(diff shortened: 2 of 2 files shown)


gmlueck (approved):
LGTM!

adoc/chapters/device_compiler.adoc
```diff
@@ -219,24 +219,6 @@ The standard {cpp} fixed width types, e.g. [code]#int8_t#,
...
-device compilers use the host sizes for pointers or [code]#size_t#. Devices
-may be using a smaller size internally for pointers and [code]#size_t#,
-but this should not impact the programming model for users. In the case where
-devices use larger pointer size internally than the host, then
-[code]#size_t# will match the <<host-pointer>> size. For structures shared
-between host and device it is recommended to use fixed width types for pointers.
-Scalar data types for a SYCL device compiler are described in
-<<table.types.fundamental>>.
-====
-endif::showtodos[]
```
keryell, line 237: I forgot that we had this description.
It looks like more as an implementation note.
Do you think this has to be removed?

adoc/chapters/opencl_backend.adoc
```diff
@@ -443,13 +443,6 @@ whether the object has actively been used on these devices yet or not.
 
 // From 3.10 Language restrictions in kernels
 
-Some types in SYCL vary according to pointer size or vary on the host
-according to the host ABI, such as [code]#size_t# or [code]#long#. In order
-for the SYCL device compiler to ensure that the sizes of
-these types match the sizes on the host and to enable data of these types
-to be shared between host and device, the OpenCL interoperability types
-are defined, [code]#sycl::cl_int# and [code]#sycl::cl_size_t#.
```
keryell, line 451: This section does not seems wrong as an introductory matter, besides the `size_t` and the name space.
So why not just replacing `size_t` by `short` for example and replace the `sycl::cl_` by `sycl::opencl::cl_`?

adoc/chapters/device_compiler.adoc
gmlueck: I suggested removing this note in #312.  The first part of this note reads like a suggestion for implementers, but the second to last sentence sounds like a requirement for applications:

> For structures shared between host and device it is recommended to use fixed width types for pointers.

That is definitely not true.  We do not want to force applications to use the fixed-width types for all structures shared between host and device.

Personally, I don't think this note is helpful to implementers.  We already say that fundamental types have the same size on host and device, so this statement about `size_t` is redundant.  The statement about devices using a larger pointer size also seems confusing to me.  I was just worried that this note added more confusion than clarity.

adoc/chapters/opencl_backend.adoc
gmlueck: To me, it seems like this whole paragraph is talking about cases where a type might have a different size in host vs. device.  I think the first sentence uses `long` as an example because that could be 32-bits on some systems and 64-bits on others.  The implication is that the host and device might have different interpretations about the size of `long`.  However, this is not possible in SYCL because we mandate that all fundamental types have the same size on host and device.

Naghasan (approved):

tomdeakin:
Ready to merge please. Thanks.

PR #322, opened 2022-11-28
Fix description of max_work_item_sizes

### LINKED ISSUE #310: Unclear descriptions of max_work_item_sizes
In Table 25. Device information descriptors, there are three `max_work_item_sizes` descriptors described like this:

> Returns the maximum number of work-items that are permitted of the work-group of the nd_range. The minimum value is (1) for devices that are not of device type info::device_type::custom.

What does "permitted of the work-group of the nd_range" mean?  It's unclear to me whether this is intended to return:

1. the maximum number of work-items that are permitted in each dimension of a work-group
2. the maximum number of work-items that are permitted in each dimension of an ND-range
3. the maximum number of work-groups that are permitted in each dimension of an ND-range (probably not this one)

Does anybody know what the intent was here?

### LINKED ISSUE #330: Return value of max_work_item_sizes
`info::device::max_work_item_sizes<1>` returns an `id<1>`.  Since it describes a number of work-items, should it return a `range<1>` instead?

As written, the output of `max_work_item_sizes` cannot be used to construct an `nd_range`, which I found surprising.

Pennycook:
New description is aligned with the behavior of OpenCL's CL_DEVICE_MAX_WORK_ITEM_SIZES.

Also replaced an instance of "both" with "each" to clarify that implementations are not required to return the same maximum value for both dimensions of a two-dimensional work-group.

Closes #310, closes #330.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1808,7 +1808,7 @@ a@
 info::device::max_work_item_sizes\<1>
 ----
 
-    @ [.code]#id<1>#
+    @ [.code]#range<1>#
    a@ Returns the maximum number of work-items that are permitted in a
       work-group for a kernel running in a one-dimensional index space. The
       minimum value is latexmath:[(1)] for [code]#devices# that are not of
@@ -1820,7 +1820,7 @@ a@
 info::device::max_work_item_sizes\<2>
 ----
 
-    @ [.code]#id<2>#
+    @ [.code]#range<2>#
    a@ Returns the maximum number of work-items that are permitted in each
       dimension of a work-group for a kernel running in a two-dimensional index
       space. The minimum value is latexmath:[(1,1)] for [code]#devices# that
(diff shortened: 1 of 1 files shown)


tomdeakin (approved):
LGTM

adoc/chapters/programming_interface.adoc
```diff
@@ -1809,8 +1809,8 @@ info::device::max_work_item_sizes\<1>
 ----
 
     @ [.code]#id<1>#
-   a@ Returns the maximum number of work-items that are permitted of the
-      work-group of the [code]#nd_range#. The minimum value is latexmath:[(1)]
+   a@ Returns the maximum number of work-items that are permitted in a
+      work-group. The minimum value is latexmath:[(1)]
```
gmlueck, line 1813: Is the idea that this query is only meaningful for a 1-D parallel_for?  If so, should we say something like:

> Returns the maximum number of work-items that are permitted in a work-group for a kernel running in a one-dimensional index space.

Etc. for the other queries.

Pennycook: I like that.  Applied your suggestion in 31e120f5.

gmlueck (approved):
LGTM (though my review doesn't count since I'm in the same org as John).

tomdeakin (approved):
LG(er)TM

keryell (approved):
Thanks for the clarification.

keryell:
Waiting for a new commit for https://github.com/KhronosGroup/SYCL-Docs/issues/330

PR #323, opened 2022-12-01
Clarify "group" meaning in algorithm descriptions

Pennycook:
The specification sometimes uses "group" to mean "work-group" (as in sycl::group) and at other times to mean "group of work-items" (a concept describing both sycl::group and sycl::sub_group).

This commit clarifies several usages of "group" in the descriptions of group algorithms, by explicitly referring to the specific instance of a group object (g).

---

Aside: I consider this a short-term fix to clarify the existing specification.  But the fact we have this ambiguity is leading me to question whether `sycl::group` is a good name now that "group" is the name of a higher level concept. But addressing that would be more than a clarification, and I don't think it can be tackled as part of this PR.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19904,7 +19904,7 @@ include::{header_dir}/groups/broadcast.h[lines=4..-1]
 +
 --
 _Returns:_ The value of [code]#x# from the work-item with the smallest linear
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks for the editorial clarification.

tomdeakin (approved):
LGTM

PR #336, opened 2023-01-03
Add atomic_ref::fetch_xor to synopsis

gmlueck:
This function is described in table 154 "Additional member functions available on an object of type `atomic_ref<T>` for integral `T`, but we forgot to add it to the synopsis.  Do that now.


keryell (approved):
Thanks!

TApplencourt (approved):

PR #337, opened 2023-01-05
Clarifications to sub-group

gmlueck:
Clarify some of the wording about sub-groups:

* In the architecture chapter, clarify that the work-items that compose a sub-group are chosen in an implementation defined way.  Previously, we said only that the size and number of sub-groups is implementation defined.

* Also in the architecture chapter, strengthen the wording about what device code must do w.r.t. sub-groups in order to be portable.

* In the programming interface chapter, expand the specification of `group::get_group_id` and `sub_group::get_group_id` to clarify that the group id can be used to identify the work-item in the global nd-range while the sub-group id cannot.

While I was there, clean up some minor things:

* Change some glossary references for plural words so that the "s" is included in the text of the glossary link.

* Use the wording "Returns the same value as" for some `group` and `sub_group` functions to follow our existing precedent.

Partially addresses internal issue 638.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1204,14 +1204,14 @@ kernels using the features described in <<sec:reduction>>.
 
 === Work-group data parallel kernels
 
-Data parallel <<kernel>>s can also execute in a mode where the set of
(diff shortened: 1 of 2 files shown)


tomdeakin (approved):
LGTM

keryell (approved):
Thanks.

PR #338, opened 2023-01-09
Mention kernel_handler in kernel function definition

npmiller:
The `operator()` may also take a `kernel_handler`.


gmlueck (approved):

TApplencourt (approved):

keryell (approved):
Good catch.

PR #343, opened 2023-01-13
Relax requirement on backend traits being available

npmiller:
The current wording mandates that `backend_traits` specializations must be provided for every backend in the backend enum class.

This seems a bit overly restrictive and means that implementations would have to change the backend enum class depending on which backends are being supported on a specific installation.

In addition backend traits are defined based on underlying vendor types, which may rely on specific vendor headers that are not always available.

So this patch allows implementations to use the backend macro to also cover the case where the backend is in the enum class but the backend traits and, or backend are still not available. This should work fine with existing applications since using the backend traits requires the enum value and so backend traits use should already be guarded by the backend macro.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -52,6 +52,8 @@ _backend_name_ is the associated identifier from [code]#backend# in
 all upper-case.  See <<chapter.extensions>> for the name of the macro
 if the vendor defines the <<backend>> outside of the Khronos SYCL group.
 
+If a backend listed in [code]#enum class backend# is not available at compile
+time, the associated macro will be left undefined.
 
 [[sec:generic-vs-non-generic]]
 == Generic vs non-generic SYCL
@@ -242,7 +244,8 @@ A series of type traits are provided for <<backend>> interoperability,
 defined in the [code]#backend_traits# class.
 
 A specialization of [code]#backend_traits# must be provided for each named
-<<backend>> enumerated in the enum class [code]#backend#.
+<<backend>> enumerated in the enum class [code]#backend# that is available at
+compile time.
 
   * For each <<sycl-runtime>> class [code]#T# which supports
     <<sycl-application>> interoperability with the <<backend>>, a
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -22,7 +22,7 @@ specifically indicates a values for the enumerator.
 [[sec:backends]]
 == Backends
 
-The <<backend, SYCL backends>> that are available to a SYCL implementation can be identified
+The <<backend, SYCL backends>> that can be supported by a SYCL implementation are identified
 using the [code]#enum class backend#.
 
 [source,,linenums]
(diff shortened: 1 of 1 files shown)


gmlueck:
I'm not opposed to this change, but I think an implementation could conform to the original wording by doing this:

```
enum class backend {
#ifdef SYCL_BACKEND_OPENCL
  opencl = 1,
#endif
#ifdef SYCL_BACKEND_FOO
  foo = 2,
#endif
};
```

This way, the backend is only enumerated in the enum class if it is available at compile time.

adoc/chapters/programming_interface.adoc
```diff
@@ -52,6 +52,8 @@ _backend_name_ is the associated identifier from [code]#backend# in
 all upper-case.  See <<chapter.extensions>> for the name of the macro
 if the vendor defines the <<backend>> outside of the Khronos SYCL group.
 
+If a backend listed in [code]#enum class backend# is not available at compile
+time, the associated macro will be left undefined.
```
keryell, line 56: This seems to contradict lines 33—35:

> The [code]#enum class backend# is implementation-defined and must be
> populated with a unique identifier for each <<backend>> that the SYCL
> implementation supports.

If the back-end is not available, it is not supported and cannot be in the `enum`.

npmiller: Good point, I've attempted to update the other sections to reflect the changes as well.

npmiller:
About @gmlueck 's suggestion, I was initially a bit worried about potential ABI issues with having different subsets of the enum defined in different cases, for example a SYCL library may be built with all of them defined, and a SYCL application only have some of them, but thinking about it a bit further I do think it would be a suitable solution if implemented properly.

So maybe this patch is not actually as required as I initially thought.

However I do wonder if it still wouldn't be nicer to have it as described in this patch so that the enum doesn't change and is able to list backends that *can* be supported and not just the ones that are.

For a bit more background, maybe it's a bit implementation specific, but in DPC++ we have a plugin architecture, so we want to be able to have a "generic" DPC++ installation that can support many backends, but then be able to add the backends individually by installing the corresponding plugins, and with the current spec that means adding the backend macro and also adding the enum value.

gmlueck:
> However I do wonder if it still wouldn't be nicer to have it as described in this patch so that the enum doesn't change and is able to list backends that can be supported and not just the ones that are.

I'm not sure we really need a spec change for this.  Is there any way for an application to enumerate all of the enumerators in `enum class backend`?  I think there is no way to do that (using standard C++).  Therefore, it is not observable whether an implementation has "extra" enumerators in `enum class backend`.

On the other hand, I don't see any harm with accepting this change either, so I am not opposed.

keryell:
> I'm not sure we really need a spec change for this. Is there any way for an application to enumerate all of the enumerators in `enum class backend`? I think there is no way to do that (using standard C++). Therefore, it is not observable whether an implementation has "extra" enumerators in `enum class backend`.

At least in triSYCL I use the great https://github.com/Neargye/magic_enum. Hackisch for now, but just works™. :-)

keryell:
Gordon to follow up on this.

AerialMantis:
I've discussed this with @npmiller and we concluded that we've been able to work around this in DPC++, so we don't strictly need this anymore, however, we think there is still some value to the change as it would allow for simpler implementations.

Effectively, the current wording requires that the macros and enum members be defined in tandem, however, this means that the enum class would need to be wrapped in #ifdef macros as @gmlueck described above, so you could potentially have different definitions of the enum class depending on how the library is built. We established this is not necessarily a problem for ABI as long as the enum class has a consistent type and explicit values for each member as there shouldn't be any observable effect since users need to check for the macro before using enum member anyway, but it does require the implementation to also #ifdef the macros anytime it's used.

This change would allow the enum definition to remain static regardless of which macros are defined, this allows the implementation to assume the full enum class is always defined, simplifying the implementation, but should not impact user code as the spec still asserts that usage of the enum members is backend-specific and therefore must be wrapped with an #ifdef.

We'd like to still consider this PR, however, we don't feel too strongly about this.

gmlueck (approved):

keryell (approved):
Thanks.

fraggamuffin:
LGTM

PR #344, opened 2023-01-13
Minor fix to "item::operator size_t()"

gmlueck:
I happened to notice this when reviewing the CTS test plan.


TApplencourt (approved):

keryell (approved):
Good catch on this SYCL 1.2 left-over.

PR #345, opened 2023-01-13
Clarify the "reducer" member types and constants

gmlueck:
This was motivated by some confusion that arose while reviewing the CTS test plan for reduction variables.  It wasn't clear to some reviewers what the dimensionality of the reduction variable was in all cases.  This commit adds a table which clarifies the value of `dimensions` as well as the type aliases that are defined for the reducer class.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12798,7 +12798,8 @@ implementation is free to re-use reducer variables (e.g. across work-groups
 scheduled to the same compute unit) if it can guarantee that it is safe to do
 so.
 
-The member functions of the [code]#reducer# class are listed in
+The type aliases and constant static members of the [code]#reducer# class are
+listed in <<table.types.reducer>> and its member functions are listed in
 <<table.members.reducer>>.  Additional shorthand operators may be made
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks for the clarification

adoc/chapters/programming_interface.adoc
```diff
@@ -12808,6 +12809,41 @@ operation, as described in <<table.operators.reducer>>.
...
+----
+   a@ The type of the combiner operator [code]#BinaryOperation# that was
+      passed to the reduction function that created this reducer object.
+
+a@
+[source]
+----
+static constexpr int dimensions
+----
+   a@ The number of dimensions of the reduction variable.  If this reducer
```
TApplencourt, line 12841: Sorry, Side question. If the reduction variable was created using a `nd_item` it will take the dimension of the `nd_item`?

gmlueck: No, I don't think so.  I believe the number of dimensions of the reduction variable is unrelated to the number of dimensions of the parallel_for.

TApplencourt (approved):

fraggamuffin:
LGTM

PR #346, opened 2023-01-17
Clarify native_specialization_constant when empty

gmlueck:
Clarify the behavior of `native_specialization_constant` when the kernel bundle contains no specialization constants.  I think we intended this to return `false` in this case, not `true`.


TApplencourt (approved):

keryell (approved):
Thanks.

AerialMantis (approved):

fraggamuffin:
LGTM

PR #352, opened 2023-01-26
Fix misprints in marray and reduction

nolmoonen:
Reduction:
- Changed `remove_v_t` to `remove_cv_t` in identity conditions.

Math array:
- Consistently named argument `v` instead of `lhs`/`rhs` for unary operators, in the table for hidden friend functions and in the code snippet.
- Removed non-floating-point requirement for one logical operator overload (the other overloads already are correct).
- Changed return type `RET` to actual type `bool` in one instance.
- Fixed indentation in code snippet.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12491,7 +12491,7 @@ sycl::logical_and
 a@
 [source]
 ----
-std::is_same_v<std::remove_v_t<AccumulatorT>, bool>
+std::is_same_v<std::remove_cv_t<AccumulatorT>, bool>
 ----
 a@
 ----
@@ -12506,7 +12506,7 @@ sycl::logical_or
 a@
 [source]
 ----
-std::is_same_v<std::remove_v_t<AccumulatorT>, bool>
+std::is_same_v<std::remove_cv_t<AccumulatorT>, bool>
 ----
 a@
 ----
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -17736,11 +17736,9 @@ Where [code]#OP# is: [code]#&#, [code]#|#, [code]#^#.
...
 [source]
 ----
-marray<RET, NumElements> operatorOP(const DataT& lhs, const marray& rhs)
+marray<bool, NumElements> operatorOP(const DataT& lhs, const marray& rhs)
 ----
-   a@ Available only when: [code]#DataT != float && DataT != double && DataT != half#.
-
-Construct a new instance of the [code]#marray# class template with [code]#DataT = bool# and
+   a@ Construct a new instance of the [code]#marray# class template with [code]#DataT = bool# and
 same NumElements as [code]#lhs# [code]#marray# with each element of the new [code]#marray# instance
```
gmlueck, line 17742:
```suggestion
same NumElements as [code]#rhs# [code]#marray# with each element of the new [code]#marray# instance
```

This also looks like it was wrong in the description.

nolmoonen: Thanks, fixed it and also another occurrence of the same problem.

keryell (approved):
Thanks for the cleaning!

PR #354, opened 2023-02-02
Clarify blocking behavior of `queue::submit`

### LINKED ISSUE #307: Clarify blocking behavior of `queue::submit`
There seems to be a typo in the current wording regarding the behavior of `queue::submit` in section 4.6.5 (emphasis mine):

> The submit member function schedules ` command groups` **asynchronously**, so any errors in the submission of a `command group` are handled by throwing **synchronous** SYCL exceptions.

I assume the intent here was to say ~command groups~ (edit: command group function objects) are scheduled *synchronously*. Furthermore, I couldn't find any information on whether the CGF is guaranteed to be executed in the thread that calls `queue::submit`. While fixing the wording above would somewhat imply this, the word "schedule" could also mean "may run in a different thread but completes before the call to `queue::submit` returns".

I did a quick check and it looks like ComputeCpp, DPC++ and hipSYCL currently all run the CGF in the calling thread. This is the less surprising (and dangerous) behavior and should be codified in the spec, in my opinion.

psalz:
Clarify that the CGF provided to `queue::submit` is invoked in the calling thread.

Resolves #307.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2906,9 +2906,12 @@ function.
 
 All member functions of the [code]#queue# class are synchronous and errors
 are handled by throwing synchronous SYCL exceptions. The [code]#submit#
-member function schedules <<command-group,command groups>> asynchronously, so any errors
-in the submission of a <<command-group>> are handled by throwing synchronous
-SYCL exceptions. Any exceptions from the <<command-group>> after it has
+member function synchronously invokes the provided
+<<command-group-function-object>> (as described in
+<<sec:command-groups-exec-order>>) in the calling thread, thereby scheduling a
+<<command-group>> for asynchronous execution. Any error in the submission of a
+<<command-group>> is handled by throwing a synchronous SYCL exception.
+Any exceptions from the <<command-group>> after it has
 been submitted are handled by passing <<async-error,asynchronous errors>> at specific times to an
 <<async-handler>>, as described in <<error-handling>>.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2911,8 +2911,8 @@ member function synchronously invokes the provided
 <<sec:command-groups-exec-order>>) in the calling thread, thereby scheduling a
 <<command-group>> for asynchronous execution. Any error in the submission of a
 <<command-group>> is handled by throwing a synchronous SYCL exception.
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -2906,9 +2906,12 @@ function.
...
 All member functions of the [code]#queue# class are synchronous and errors
 are handled by throwing synchronous SYCL exceptions. The [code]#submit#
-member function schedules <<command-group,command groups>> asynchronously, so any errors
-in the submission of a <<command-group>> are handled by throwing synchronous
-SYCL exceptions. Any exceptions from the <<command-group>> after it has
+member function synchronously invokes the provided
+<<command-group-function-object>> (as described in
+<<sec:command-groups-exec-order>>) in the calling thread, thereby scheduling a
+<<command-group>> for asynchronous execution. Any error in the submission of a
+<<command-group>> is handled by throwing a synchronous SYCL exceptions.
```
gmlueck, line 2913:
```suggestion
<<command-group>> is handled by throwing a synchronous SYCL exception.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -2906,9 +2906,12 @@ function.
...
+Any exceptions from the <<command-group>> after it has
```
gmlueck, line 2914:
```suggestion
Any errors from the <<command-group>> after it has
```

I think we mean errors more generally here, right?  For example, an error from the underlying backend could cause the asynchronous error handler to be called.  An exception from the application's host task could also cause the asynchronous error handler to be called, but we say that in the definition of host tasks.

Notably, exceptions in other commands are **not** guaranteed to call the asynchronous error handler (because exception are not allowed in device code).

psalz: Right. Not really related to this PR but I guess we can do a drive-by clarification :)

adoc/chapters/programming_interface.adoc
psalz: Oops, thanks.

gmlueck (approved):

keryell (approved):

AerialMantis (approved):

illuhad (approved):

fraggamuffin:
will wait, need more time
need to access same TLS vars in outside scope
submit calls fn object (factory fn) is synchronous to control scope of how command group is constructed
LGTM

PR #356, opened 2023-02-07
Clarify USM allocation of zero size

### LINKED ISSUE #355: USM allocation behaviour when requesting memory of zero size
Hi,

I could not find any information in the [SYCL spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm) about how the USM allocations functions (including the usm_allocator) should behave when requested to allocate a memory of zero size.

For buffers it is explicitly stated that zero-sized buffers are permitted. However, this information seems missing for USM.

Testing with DPC++, `usm_allocator::allocate` throws a `sycl::_V1::memory_allocation_error` exception (for both host and shared specializations), while `malloc_shared` performs fine and returns a nullptr.

gmlueck:
Clarify the behavior of the USM malloc-like functions when the requested size is zero bytes, making them consistent with the behavior of `std::malloc`.

Closes #355.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10184,9 +10184,10 @@ leak.  If there are not enough resources to allocate the requested memory,
 these functions return [code]#nullptr#.
 
 When the allocation size is zero bytes ([code]#numBytes# or [code]#count# is
-zero) the value returned by these functions is unspecified, but the returned
-pointer may not be used to access storage.  If the pointer is not null, it
-must be passed to [code]#sycl::free# to avoid a memory leak.
+zero), these functions behave in a manor consistent with {cpp}
+[code]#std::malloc#.  The value returned is unspecified in this case, and the
+returned pointer may not be used to access storage.  If this pointer is not
+null, it must be passed to [code]#sycl::free# to avoid a memory leak.
 
 [[table.usm.device.allocs]]
 .Device USM Allocation Functions
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10183,6 +10183,11 @@ eventually be deallocated with [code]#sycl::free# in order to avoid a memory
 leak.  If there are not enough resources to allocate the requested memory,
 these functions return [code]#nullptr#.
 
+When the allocation size is zero bytes ([code]#numBytes# or [code]#count# is
+zero) the value returned by these functions is unspecified, but the returned
+pointer may not be used to access storage.  If the pointer is not null, it
+must be passed to [code]#sycl::free# to avoid a memory leak.
+
 [[table.usm.device.allocs]]
 .Device USM Allocation Functions
 [width="100%",options="header",separator="@",cols="65%,35%"]


TApplencourt (approved):

TApplencourt:
Minus comment, should we add a sentence to say that we follow `std::maloc` behavior? May help people realize that we are not doing something "weird".

gmlueck:
> should we add a sentence to say that we follow std::maloc behavior? May clarify that we are not doing something "weird" / special in sycl.

Yes, good idea.  See f1b84006b9280ee8247fa15230abb109f63949f0

keryell:
> LGTM. Minus comment, should we add a sentence to say that we follow `std::maloc` behavior? May clarify that we are not doing something "weird" / special in sycl.

Do you really think this is some behavior we want to add instead of returning a null pointer?
What is the kind implementation you are thinking to?
Or is this is for an horror movie?

TApplencourt:
haha. cpp committee peoples are strange... But cpp library writers are even stranger! (locking really carefully in the direction of boost)

nliber:
IIRC, historically `malloc` is like this because different vendors had different behavior, and this was the compromise for the original C standardization effort (which had no guarantee of success, especially if they broke existing code).

We could do better.  Since there are no valid use cases for `malloc(0)`, we could throw an exception or even abort.  We could also do this in the future if we made it a precondition that the size be non-zero.

keryell:
@TApplencourt And then we will need to specify that, if the pointer is not null, it has to be different every time in the context of https://cplusplus.github.io/LWG/issue9? All this seems to add some technical debt to SYCL inherited from specific `malloc` CPU implementation, which probably does not make sense in the context of heterogeneous memory or SYCL for Safety Critical Systems.

gmlueck:
Testing with both g++ and clang++, it seems that both `std::malloc(0)` and `std::allocator<T>::allocate(0)` return a non-null pointer.  Successive calls seem to return unique non-null pointer values.  Therefore, I think we should not mandate that the SYCL malloc-like functions return a null pointer when the size is zero.  This might make it hard for a vendor to reuse existing `std` code for their USM allocation functions.

> Since there are no valid use cases for malloc(0), we could throw an exception or even abort.

I think we should not do this.  I have written code in the past that assumes `malloc(0)` succeeds because this is a natural way to allow an algorithm to handle the "no input" case without any special case code.  I'm sure other people have written code with similar assumptions.

I thought our goal was to match standard C/C++ behavior in SYCL when possible.  Why wouldn't we follow that tenant here too?

TApplencourt:
@keryell  `Opened: 1998-01-04 `! I was not even born!

I see your point now. My naive first idea is that everybody obviously knows everything about C++ and its quirk, or at least pretends to know and is used to the quirk.  So my idea, similar to Greg, was that as far as possible, we should follow C++ semantics to follow the principle of astonishment.

Saying that the reality seems to be that C++ is far more horrible than I expected (and that gcc and clang don't even follow the standard)... So I guess we need to find a balance between fixing some horrible legacy and not breaking too many people's implicit assumption that we are a preview of C++56.

For example, I totally agree with greg that we should allow `malloc(0) / new(0)` for generality purposes. I also have written code that is really on that. But maybe we don't need to stipulate that each pointer needs to be unique...

nliber:
@TApplencourt when you say you have code relying on `malloc(0)`, are you relying on what it returns, or that you can just call `free(p)` on it?

The `operator new(0)` case is a corner case (good find @keryell!), as you have to call it with that exact syntax in order to get a zero sized allocation.  Like C++, I'm not particularly concerned about that case showing up in real code.

From a correctness point of view, if we give it semantics that are hard to reason about, people will write code dependent on those semantics.  IMO, I don't want to give it any more semantics than being able to `free(p)` it, because anything else makes the resulting code harder to reason about.

The safetly point of view tends to want everything have well-defined semantics, even at the cost of correctness.

TApplencourt:
> I just rely on the fact that I  can call free on it. Typical code that I wrote:
```
foo(N)
   int *a = malloc(N);
   s = 0;
   for  (int i=0; i < N; i++) 
      s+=compute()
   free (a);
   return s
} 
```

> IMO, I don't want to give it any more semantics than being able to free(p) it, because anything else makes the resulting code harder to reason about.

Work for me!

gmlueck:
The code pattern I have used in the past is similar to what @TApplencourt shows above.  It just assumes that the result of `malloc(0)` can be passed to `free()`, and this is guaranteed by the C++ and C specifications.

Regarding the `operator new(0)` case ... I think there is no need for the SYCL specification to say anything about this.  SYCL inherits the semantics of `new` from the C++ specification, so there is no need for us to say anything more.  This PR is only clarifying the malloc-like functions that are specific to SYCL.

keryell:
My point is that `sycl::anymalloc(0)` should return a null pointer which you can obviously `sycl::anyfree()` (or not) with love and will be a no-op.
I do not like the idea that `sycl::anymalloc(0)` could return a non-null pointer that *you absolutely need to `sycl::anyfree()`* because some Jurassic C implementation used by Thomas' grand-grand-mother did. Not even thinking about what a curious mind could do with such a tempting non-null pointer to some writable memory...
Not even thinking to the USM allocator allocating with the GPU device-driver a 2GiB page of shared memory by default for this.
I guess on a CPU allocator they just optimize the hot path because it is heavy used. So why not allocating a slab even for people asking for a `malloc(0)`, because any sensible high-performance application will not do it any way? I guess the CPU branch predictor or an `[[unlikely]]` on this test in a modern implementation will make the test for free.
There was a great ISO C++ SG23 meeting of safety and security this afternoon by the way. :-)
I have even heard about SYCL for Safety Critical Systems somewhere... :-)
That said, if people insist to have pointers and `malloc`, why not providing such a confusing feature after all...

gmlueck:
> I do not like the idea that `sycl::anymalloc(0)` could return a non-null pointer that _you absolutely need to `sycl::anyfree()`_ because some Jurassic C implementation used by Thomas' grand-grand-mother did.

I don't know why the C/C++ committees decided to specify it this way, but this is what they did.  It also seems that *modern* compilers do in fact return a non-null pointer in this case.  Again, I'm not certain why they decided to do this, but this seems to be the way it works.  You can see in this godbolt test.  I verified that both clang 15.0.0 and gcc 12.2 on x86_64 return a non-null pointer for `malloc(0)`.

https://godbolt.org/z/casn789dr

nliber:
@keryell @gmlueck Looking at [n4917](https://wg21.link/N4917) [[basic.stc.dynamic.allocation]p2 footnote 27](http://eel.is/c++draft/basic.stc.dynamic.allocation#footnote-27), we have "C++ differs from C in requiring a zero request to return a non-null pointer.".

We should require the same.

nliber:
@keryell I only provide the history for context.  It doesn't actually matter why they did what they did for C back in the '80s.  Consistency, however, does matter, even if you don't want to weigh it highly.

> Not even thinking about what a curious mind could do with such a tempting non-null pointer to some writable memory...

I don't understand this.  Surely that line of reasoning applies to *any* non-null pointer, doesn't it?

> Not even thinking to the USM allocator allocating with the GPU device-driver a 2GiB page of shared memory by default for this.

This is the edge case where they are deliberately allocating `0` bytes.  I have a hard time seeing how that is ever going to matter.

> That said, if people insist to have pointers and malloc, why not providing such a confusing feature after all...

Because, IMO, it is harder to reason about if a null pointer return doesn't just indicate a failure to allocate.  Of course, because some implementations back in the '80s wanted it to indicate that no allocation happened while others wanted it to indicate a failure to allocate is why we are in this situation in the first place... :-)

keryell:
First the C++ draft says that `malloc()`/`free()` works like in C (sorry for the bad formatting from the PDF):
```
20.2.12
1
C library memory allocation
[c.malloc]
[Note 1 : The header <cstdlib> (17.2.2) declares the functions described in this subclause. — end note]
void* aligned_alloc(size_t alignment, size_t size);
void* calloc(size_t nmemb, size_t size);
void* malloc(size_t size);
void* realloc(void* ptr, size_t size);
2Effects: These functions have the semantics specified in the C standard library.
3Remarks: These functions do not attempt to allocate storage by calling ::operator new() (17.6.3).
4These functions implicitly create objects (6.7.2) in the returned region of storage and return a pointer
to a suitable created object. In the case of calloc and realloc, the objects are created before the
storage is zeroed or copied, respectively.
§ 20.2.12
608Dxxxx
© ISO/IEC
void free(void* ptr);
5Effects: This function has the semantics specified in the C standard library.
6Remarks: This function does not attempt to deallocate storage by calling ::operator delete().
See also: ISO C 7.22.3
```
Then the latest C23 draft says in https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3047.pdf says in *7.24.3 Memory management functions*

> If the size of the space requested is zero, the behavior is implementation-defined: either a null pointer is returned to indicate an error, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

So it is perfectly possible to have USM SYCL malloc implemented as
```c++
if (!size) [[unlikely]]
  return nullptr;
else
  return some_real_alloc(size);
```

And we can call `free()` later on the `nullptr` too if we insist as a perfectly 1978 K&R C API, but we do not care.
My point is to make this choice because it simplifies the understanding by removing the possibility of UB access of non-null pointer and even simplifies the implementation in the case of real device paged allocation.

Answering to @nliber:
>>  Not even thinking about what a curious mind could do with such a tempting non-null pointer to some writable memory...

> I don't understand this. Surely that line of reasoning applies to any non-null pointer, doesn't it?

Except this is not a normal non-null pointer. We remove the corner case where `malloc(0)` returns every time a different non-null pointer. But it is UB to dereference this nice pointer, which make this API even more troubling.

nliber:
It's a non-dererenceable pointer, which no different than, say, what `vector::data()` can return for an empty `vector` (which may or may not be a null pointer).  Under typical implementations:

```c++
vector<int> v;
assert(nullptr == v.data());
v.reserve(1);
assert(nullptr != v.data());
```

And `malloc(0)` is fairly contrived, because real code is more like `static_cast<T*>(malloc(n * sizeof(T)))` (with `sizeof` guaranteed by the language to be strictly `> 0`) followed by looping over `n`.  (Okay, real code is probably a C-style cast, but not really relevant to this point.).  I don't see how a dereferencing bug sneaks in here.  If they forget to multiply by `sizeof(T)` it still accidentally just works for the `n == 0` case.

keryell:
> @keryell It's a non-dererenceable pointer, which no different than, say, what `vector::data()` can return for an empty `vector` (which may or may not be a null pointer). Under typical implementations:
> 
> ```c++
> vector<int> v;
> assert(nullptr == v.data());
> v.reserve(1);
> assert(nullptr != v.data());
> ```

Of course there are a lot of arcane points in C++ :-( but by standardizing SYCL we can reduce the cognitive burden while still being compatible with the plain old malloc()/free() coding style. :-) Think to the tooling people who will have to implement UBsan-like tools for SYCL in the case USM alloc does not return 0 for an empty allocation...

> And `malloc(0)` is fairly contrived, because real code is more like `static_cast<T*>(malloc(n * sizeof(T)))` (with `sizeof` guaranteed by the language to be strictly `> 0`) followed by looping over `n`. (Okay, real code is probably a C-style cast, but not really relevant to this point.). I don't see how a dereferencing bug sneaks in here. If they forget to multiply by `sizeof(T)` it still accidentally just works for the `n == 0` case.

I was thinking more to https://godbolt.org/z/8Kro9djYM
```c++
#include <cstdlib>
#include <iostream>

template <typename T>
T* my_alloc(std::size_t O) {
  return static_cast<T*>(std::malloc(sizeof(T)*0)); // haha, a 0 instead of O
}

int main() {
    auto p = my_alloc<int>(1);
    // entering the the rainbow unicorn world where, with UB, everything is possible
    *p = 42;
    std::cout << *p << std::endl;
}
```
works perfectly an display 42. :-)
Is this the kind of bugs Khronos wants to promote?
With my solution you get just a blunt segmentation violation.
I invite everybody to look (again) at the presentation from JF Bastien last week ISO C++ SG23 of safety and security.

nliber:
@keryell

1. This seems awfully contrived.
2. It doesn't actually fix the problem, because `*p = 42;` is UB in both cases (null pointer or not).  You are just hoping it crashes.
3. A UBSan that can only detect dereferencing on null pointers and no other kinds of bad dereferences isn't terribly useful.
4. Special cases increase the cognitive burden, because user code can then rely on it returning a null pointer, so all code readers have to take that into account when looking at code.
5. JF has actual evidence that his proposal will address real issues (and not just one anecdote).  Have you ever run into a bug caused by `malloc(0)`?

If you think using `malloc` is really this error-prone, then propose to remove it.

keryell:
I am depressed...
Today this black-box can implement A or B according to the C spec. If we pick A, a user can still use it as B and it is trivial to implement detection of UB. My choice is easy, :-)

gmlueck:
Perhaps we could resolve this by adding functions in SYCL-Next like `new_shared`, `new_host`, and `new_device` that have C++ `new` semantics while allocating objects in USM.  This way, applications that like the C-style malloc functions can use `malloc_shared` and friends with the normal C/C++ semantics (including the C/C++ semantic for `malloc(0)`).  Applications that prefer C++ `new` behavior can use `new_shared` and friends, which would follow normal C++ semantics for `new`.

However, I'm not sure if this will address @keryell's concerns either because it appears that C++ rules for `new[]` do not require any sort of error when the array dimension is `0`.  In fact, modern compilers seem to return a non-null pointer in this case also:

https://godbolt.org/z/9r3Mraxq4

fraggamuffin:
need one more

keryell:
@gmlueck It looks like there is a misunderstanding. My proposal have exactly the requested normal C/C++ semantics.

keryell:
At least my proposal simplify the writing of the CTS. :-)

TApplencourt:
So I just learned that OpenMP returns `null` in this case... So It's the final blow; I'm on ronan side now :(

fraggamuffin:
Vote:
1. null ptr
2. not null ptr: 
3. unspecified between 1 and 2: C++ std malloc does this

keryell:
Whatever the conclusion of this discussion is, I guess what it will be in SYCL for Safety Critical Systems and this is is a good motivation to have a SYCL for Safety Critical Systems standard. :-)
Just removing all these corner cases that nobody understand like this one so simple that it led to the creation of https://cplusplus.github.io/LWG/issue9.
At least if we "implement" https://cplusplus.github.io/LWG/issue9 this gives us a test plan for the CTS. :-)

keryell:
> At least if we "implement" [cplusplus.github.io/LWG/issue9](https://cplusplus.github.io/LWG/issue9) this gives us a test plan for the CTS. :-)

Actually this issue explains the problem but is not directly related to what we are discussing here since in C++ `operator new` does more things than the `std::malloc`. So let's focus on just `std::malloc`/`std:free`-like operations for USM allocation functions.

adoc/chapters/programming_interface.adoc
```diff
@@ -10183,6 +10183,12 @@ eventually be deallocated with [code]#sycl::free# in order to avoid a memory
 leak.  If there are not enough resources to allocate the requested memory,
 these functions return [code]#nullptr#.
 
+When the allocation size is zero bytes ([code]#numBytes# or [code]#count# is
+zero), these functions behave in a manor consistent with {cpp}
+[code]#std::malloc#.  The value returned is unspecified in this case, and the
+returned pointer may not be used to access storage.  If this pointer is not
+null, it must be passed to [code]#sycl::free# to avoid a memory leak.
```
keryell, line 10190:
```suggestion
When the allocation size is zero bytes ([code]#numBytes# or [code]#count# is
zero), these functions returns a null pointer.

Note that this behavior is still compatible with the {cpp}
[code]#std::malloc# as [code]#sycl::free# can be used without any effect
on the returned (null) pointer.
```

keryell (changes requested):

adoc/chapters/programming_interface.adoc
nliber: As mentioned a few hours ago at the meeting you and I were at, that behavior is NOT specification compatible with the C or C++ versions of malloc.

Code that assumes that `malloc(0)` always returns a null pointer value (and, say, doesn't call `free` on it) cannot be backported to regular C or C++.

I REJECT that characterization in the wording, and I REJECT any changes to the wording until the vote takes place next week.

nliber:
```c++
template <typename T>
T* my_alloc(std::size_t O) {
  return static_cast<T*>(std::malloc(sizeof(T)*0)); // haha, a 0 instead of O
}
```

If this is your actual concern, the problem isn't with `malloc`; rather, the problem is that `O` can easily be mixed up with `0` (and presumably `l` with `1`).

> Is this the kind of bugs Khronos wants to promote?

Are you going to propose banning `O` and `l` in identifiers?

And that doesn't even touch on what one can do with Unicode.

adoc/chapters/programming_interface.adoc
keryell: Ah, this is an important new point of information to me.
You do not only want that some typical C/C++ `std::malloc`/`std::free` can be used in a USM SYCL context but you also want that a code developed specifically in a USM SYCL context can be used, almost as is, back into a C/C++ context.
Interesting.
I am unsure this is a realistic use case, but at least this is to be considered.

Michoumichmich:
> I just rely on the fact that I can call free on it. Typical code that I wrote:
> 
> ```
> double foo(N)
>    int *a = malloc(N);
>    double s = 0;
>    for  (int i=0; i < N; i++) 
>       s+=compute(a,i)
>    free (a);
>    return s
> } 
> ```
> 
>  > IMO, I don't want to give it any more semantics than being able to free(p) it, because anything else makes the resulting code harder to reason about.
> 
> Work for me!

Hi, just saw the thread. Returning a nullptr, won't allow this to work. Any reasonable code would likely assert for `a != nullptr`

Michoumichmich:
> > Hi, just saw the thread. Returning a nullptr, won't allow this to work.
> 
> @Michoumichmich Why?

Hi @keryell, I was just thinking about the fact that most likely the snippet would be:
```
double foo(N)
   int *a = malloc(N);
   assert(a != NULL); // or deal with the error properly <-- 
   double s = 0;
   for  (int i=0; i < N; i++) 
      s+=compute(a,i)
   free (a);
   return s
} 
```
and you can't differentiate between alloc failure and 0-alloc. Anyway, it's not a big issue - just don't enter the function. But you won't be able to rely on these patterns working out of the box (thinking about https://github.com/KhronosGroup/SYCL-Docs/pull/356#issuecomment-1423210898)

keryell:
> and you can't differentiate between alloc failure and 0-alloc.

Well, this is how C/C++ works. :-) It is not a coincidence that `std::free(nullptr)` is a no-op.

> Anyway, it's not a big issue - just don't enter the function. But you won't be able to rely on these patterns working out of the box (thinking about [#356 (comment)](https://github.com/KhronosGroup/SYCL-Docs/pull/356#issuecomment-1423210898))

These patterns actually work with the 2 different possible C/C++ malloc behaviors.
The discussion here is whether we keep exactly the same 2 possible behaviors or not. What are the trade-offs, the cognitive burden, etc.?

gmlueck:
I think we decided that we would have a straw poll vote on this in today's meeting.  I think there are only two options being considered:

1. The SYCL definition of `sycl::malloc_shared` (and similar USM malloc-like function) should align with the C++ definition of `std::malloc` and say that the returned value is unspecified when the input size is `0`.  The function can return either null or non-null in this case.  Either way, it is UB if the application dereferences the pointer.  If the pointer is non-null, the application must pass it to `syc::free` to avoid a memory leak.

2. The SYCL definition of `sycl::malloc_shared` (and similar USM malloc-like function) should deviate from the C++ definition of `std::malloc` and require the implementation to return null when the input size is `0`.

Since this is my PR, I'll express my preference in the next comment and give my rationale.

gmlueck:
I think we should choose option 1 (align with C++ `std::malloc`).  I have several reasons:

* Even if we do not agree with the C++ definition of `std::malloc`, it is not the charter of our group to improve C++ in general.  Our philosophy in general has been to align SYCL with standard C++ when possible, and I do not see why this case should be considered differently.

* Many modern implementations of `std::malloc` do actually seem to return non-null when the input size is zero.  If we forbid this behavior in SYCL, it could make it harder for an implementation to reuse existing malloc pool management code when implementing `sycl::malloc_shared` and friends.  It seems interesting that modern implementations return non-null in this case, so maybe there is a good reason.

* Since many modern implementations of `std::malloc` return non-null when the input size is zero, there may be existing application code that assumes this.  Obviously, that code is not conformant to the C++ spec, but SYCL implementations may want to make it easy for these applications to port to `sycl::malloc_shared` anyway.  If the SYCL spec requires these functions to diverge from standard `std::malloc`, a SYCL implementation would not have this freedom.

keryell:
>     * Many modern implementations of `std::malloc` do actually seem to return non-null when the input size is zero.  If we forbid this behavior in SYCL, it could make it harder for an implementation to reuse existing malloc pool management code when implementing `sycl::malloc_shared` and friends.

The implementation I proposed in https://github.com/KhronosGroup/SYCL-Docs/pull/356#issuecomment-1430688956 does not seem hard or inefficient.

keryell:
I am curious about how the USM allocator crash was fixed in Intel implementation as a data point.

fraggamuffin:
Vote: 
1. The SYCL definition of sycl::malloc_shared (and similar USM malloc-like function) should align with the C++ definition of std::malloc and say that the returned value is unspecified when the input size is 0. The function can return either null or non-null in this case. Either way, it is UB if the application dereferences the pointer. If the pointer is non-null, the application must pass it to syc::free to avoid a memory leak.

2. The SYCL definition of sycl::malloc_shared (and similar USM malloc-like function) should deviate from the C++ definition of std::malloc and require the implementation to return null when the input size is 0.
Option 1: 6/12  everyone's implmentation.
Option 2: 6/12
No consensus: status quo remains? But just fix bad wording.

New poll result:
Option 1: 13/16
Option 2: 3/16

keryell:
Can anyone remember what was the real questions for the first and second poll? I am unsure about the formatting of the previous message for the questions and the results.

gmlueck:
> I am curious about how the USM allocator crash was fixed in Intel implementation as a data point.

It looks like it was not a crash, but rather that the function threw an exception.  The bug report is #355.

illuhad:
As a follow up, here we have a user having an issue with malloc(0). They are asking for consistent behavior across backends: https://github.com/OpenSYCL/OpenSYCL/discussions/1047

PR #357, opened 2023-02-07
Clarifications to device copyable

gmlueck:
* Section 3.13.1 "Device copyable" lists several types from the C++ standard library as device copyable.  Clarify that this does not create a requirement for an implementation to support these types in device code.  Rather, it is saying that *if* an implementation chooses to support these types in device code (as an extension), then those types should be device copyable.  Note that section 5.4 "Language restrictions for device functions" already says that an implementation is not required to support the use of C++ standard library functions in device code.  This clarification is consistent with the wording proposed in internal issue 533.

* Clarify that when an implementation supports these standard library classes in device code, then arrays of those classes and cv-qualified versions of those classes are also device copyable.  This is consistent with the standard C++ definition of "trivially copyable".

* We already stated that when an application explicitly declares a custom type to be device copyable via the `is_device_copyable` trait, then arrays of that type and cv-qualified versions of that type are also device copyable.  Make it extra clear that the implementation is responsible for setting the `is_device_copyable` trait to `true` for these array and cv-qualified types.

Closes internal issue 533.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1783,8 +1783,8 @@ implicitly device copyable.
 Although implementations are not required to support device code that calls
 library functions from the {cpp} core language, some implementations may
 provide device support for some of these functions.  If the implementation
-provides device support for one of the following classes in the {cpp} core
-language, that type is also implicitly device copyable:
+provides device support for one of the following classes, that type is also
+implicitly device copyable:
 
   * [code]#std::array<T, 0>#;
   * [code]#std::array<T, N># if [code]#T# is device copyable;
@@ -1798,7 +1798,8 @@ language, that type is also implicitly device copyable:
     [code]#Types# are device copyable;
   * [code]#std::basic_string_view<CharT, Traits>#;
   * [code]#std::span<ElementType, Extent># (the [code]#std::span# type has
-    been introduced in {cpp}20).
+    been introduced in {cpp}20);
+  * [code]#sycl::span<ElementType, Extent>#.
 
 If the implementation provides device support for one of the classes listed
 above, arrays of that class and cv-qualified versions of that class are also
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1778,8 +1778,13 @@ certain arguments to a <<sycl-kernel-function>> (as described in
 <<device-copyable>> as defined below.
 
 Any type that is trivially copyable (as defined by the {cpp} core language) is
-implicitly device copyable.  In addition, the following types from the {cpp}
-core language are implicitly device copyable:
+implicitly device copyable.
+
+Although implementations are not required to support device code that calls
+library functions from the {cpp} core language, some implementations may
+provide device support for some of these functions.  If the implementation
+provides device support for one of the following classes in the {cpp} core
+language, that type is also implicitly device copyable:
 
   * [code]#std::array<T, 0>#;
   * [code]#std::array<T, N># if [code]#T# is device copyable;
@@ -1794,6 +1799,10 @@ core language are implicitly device copyable:
   * [code]#std::basic_string_view<CharT, Traits>#;
   * [code]#std::span<ElementType, Extent># (the [code]#std::span# type has
     been introduced in {cpp}20).
+
+If the implementation provides device support for one of the classes listed
+above, arrays of that class and cv-qualified versions of that class are also
+device copyable.
 
 [NOTE]
 ====
@@ -1829,7 +1838,8 @@ have this support.  When the implementation has this support, a class type
 
 When the application explicitly declares a class type to be device copyable,
 arrays of that type and cv-qualified versions of that type are also device
-copyable.
+copyable, and the implementation sets the [code]#is_device_copyable_v# trait to
+[code]#true# for these array and cv-qualified types.
 
 [NOTE]
 ====


adoc/chapters/architecture.adoc
```diff
@@ -1778,8 +1778,13 @@ certain arguments to a <<sycl-kernel-function>> (as described in
 <<device-copyable>> as defined below.
 
 Any type that is trivially copyable (as defined by the {cpp} core language) is
-implicitly device copyable.  In addition, the following types from the {cpp}
-core language are implicitly device copyable:
+implicitly device copyable.
+
+Although implementations are not required to support device code that calls
+library functions from the {cpp} core language, some implementations may
+provide device support for some of these functions.  If the implementation
```
AlexeySachkov, line 1785: I wonder: how is user supposed to check if implementation supports a certain `std::` container? Is it responsibility of every implementation to document what's exactly is supported? 

I'm asking, because we have SYCL-CTS which checks that types listed below are device copyable:

https://github.com/KhronosGroup/SYCL-CTS/blob/4a9e68f81e4c63ace2c50ef1c68068b329ab9573/tests/common/type_coverage.h#L541

With the new wording, it seems like implementation is free to **not** mark them as device copyable if they are not supported in SYCL kernels. Should we remove the test then, because specification now provides no guarantees here?

gmlueck: The Khronos SYCL WG should discuss if this PR is what we intended.  (I think it is.)  Answering your questions assuming this PR is accepted:

> I wonder: how is user supposed to check if implementation supports a certain std:: container? Is it responsibility of every implementation to document what's exactly is supported?

Yes.  The core SYCL spec does not guarantee that any `std` functions or types are available in device code.  Of course, an implementation is allowed to support some of these functions as an  extension if it wants to, though.  If an application wants to be portable across SYCL implementations, it needs to use `#ifdef` around any use of `std` functions to make sure they are not used unless the implementation supports this extension.

> With the new wording, it seems like implementation is free to not mark them as device copyable if they are not supported in SYCL kernels. Should we remove the test then, because specification now provides no guarantees here?

Yes, I think we should remove this part of the test.

keryell: There are already some `std::`function usages in the CTS: https://github.com/KhronosGroup/SYCL-CTS/pull/505/files#diff-9f869743b3173b1554dab7dcd20be5a91239ec33aff3ef8cec9ad2aeac5dadf0R64

keryell: I have seen recently some `std::memcmp` into some CTS device code.

gmlueck: It would be great if we could come to an agreement about this PR soon.  Some of the CTS work for SYCL 2020 is being held up because of uncertainty about whether these `std` types need to be supported in device code.  If we accept this PR, then we would change the CTS to remove use of these `std` types in device code.

I've added the "P1" label to the associated internal issue 533 in order to raise visibility for the next WG meeting.

nliber: @gmlueck As of C++23, both `span` and `basic_string_view` are guaranteed to be trivially copyable, and no major compiler vendor expects that their current implementation will change.  `mdspan` is also specified to be trivially copyable if its component parts are trivially copyable, and all the ones provided by the standard are trivially copyable.

That leaves `array`, which may already be required to be trivially copyable if either `T` is trivially copyable or `N == 0` (because it has to be an aggregate) even though the standard doesn't explicitly say so.

gmlueck: Are you suggesting that we remove these types from the list because they are already trivially copyable, which means they are also device copyable?

I am not opposed to doing this.  My main goal in this PR to is to clarify that there is no requirement for an implement to support these types in device code.

nliber: @gmlueck I'm fine with removing them.

I think there is a deeper question here.  If a standard library implementation of a type is trivially copyable, how are users to know whether or not that type is supported in device code?

Making things more complicated:  while `array` of a trivially copyable `T` or `N==0` is trivially copyable, it contains operations that cannot be supported in device code.  An example is `array::at()`, because it throws exceptions.

gmlueck: > I agree we shouldn't have a requirement that device code support these types, but the other side of that coin is the question of how are users supposed to know?

For now, we are relying on vendor extensions for this.  If a vendor wants to support these (or any) `std` types in device code, they can create an extension for that.  We have an established way for application code to test whether an extension is supported (via macros), so applications would use that to determine if it is safe to use these types in device code.

In SYCL-Next, I think we should consider adding some guarantee to the core language saying that certain `std` types can be used in device code.

keryell: The problem is more an implementation problem. A `std::` thing might be completely valid, except you might not have all the C++ standard library digested by or implemented for the device compiler...
An advantage for the SSCP model cherished by @illuhad. (We have SMCP but not SSCP in the SYCL glossary. :-( No definition in the spec itself either).

nliber: What are the requirements on user defined trivially copyable types?  I don't see how there can be a practical distinction between those and standard library provided trivially copyable types.

gmlueck: The requirements for a trivially copyable type are the same whether it is user-defined or part of the `std` library.

However, we need to keep in mind that "trivially copyable" is not a sufficient requirement for the type to be used in device code.  Device code has other requirements, such as a prohibition on exceptions and recursion.  Therefore, even if we agree that the types listed in this section are trivially copyable, there is still no guarantee that they can be used in device code.

AerialMantis (approved):

illuhad (approved):

nliber (approved):

keryell (approved):

fraggamuffin:
spec is ambiguous and needs to be clarified before NEXT
approved LGTM

PR #358, opened 2023-02-07
Clarify sycl::atomic_ref

gmlueck:
The main change in this commit is to limit `sycl::atomic_ref` to be used only in device code and to provide a guarantee that `std::atomic_ref` will be atomic w.r.t. `sycl::atomic_ref` when the device supports host-to-device atomics.  This follows our decision in internal issue 637.

It also provides some other clarifications:

* Describes the effect of the `AddressSpace` template parameter.
* Clarifies that the type `T` must either be 32-bits or 64-bits.  If it is 64-bits, the device must have `aspect::atomic64`.

It also removes a suggestion that the compiler should issue a diagnostic if device code calls `std::atomic` or `std::atomic_ref`.  It seems odd to provide this suggestion only for these two functions since the SYCL spec does not allow device code to call any standard C++ function.  If we want to provide a suggestion (or requirement) like this, I think it should be a more comprehensive statement about calling standard C++ functions in general from device code.

Closes internal issue 637.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17945,43 +17945,77 @@ device_event(___unspecified___)
 [[sec:atomic-references]]
 === Atomic references
 
-The SYCL specification provides atomic operations based on the library
-syntax from the next {cpp} specification. The set of supported orderings is
+The [code]#sycl::atomic_ref# class provides the ability to perform atomic
+operations in device code with a syntax similar to the {cpp} standard
+[code]#std::atomic_ref#.  The [code]#sycl::atomic_ref# class must not be used
+in host code.
+
+Unlike [code]#std::atomic_ref#, [code]#sycl::atomic_ref# does not provide a
+default memory ordering for its operations.  Instead, the application must
+specify a default ordering via the [code]#DefaultOrder# template parameter.
+This ordering is used as a default for most of the atomic operations, but
+each member function also provides an optional parameter that allows the
+application to override this default.  The set of supported orderings is
 specific to a device, but every device is guaranteed to support at least
-[code]#memory_order::relaxed#. Since different devices have different
-capabilities, there is no default ordering in SYCL and the default order
-used by each instance of [code]#sycl::atomic_ref# is set by a template
-argument. If the default order is set to [code]#memory_order::relaxed#,
-all memory order arguments default to [code]#memory_order::relaxed#. If
-the default order is set to [code]#memory_order::acq_rel#, memory order
-arguments default to [code]#memory_order::acquire# for load operations,
+[code]#memory_order::relaxed#.  If the default order is set to
+[code]#memory_order::relaxed#, all memory order arguments default to
+[code]#memory_order::relaxed#.  If the default order is set to
+[code]#memory_order::acq_rel#, memory order arguments default to
+[code]#memory_order::acquire# for load operations,
 [code]#memory_order::release# for store operations and
-[code]#memory_order::acq_rel# for read-modify-write operations. If the
+[code]#memory_order::acq_rel# for read-modify-write operations.  If the
 default order is set to [code]#memory_order::seq_cst#, all memory order
 arguments default to [code]#memory_order::seq_cst#.
 
-The SYCL atomic library may map directly to the underlying {cpp} library in
-<<sycl-application>> code, and must interact safely with the host {cpp}
-atomic library when used in host code. The SYCL library must be used in
-device code to ensure that only the limited subset of functionality is
-available. SYCL device compilers should give a compilation error on use of
-the [code]#std::atomic# and [code]#std::atomic_ref# classes and
-functions in device code.
+The [code]#sycl::atomic_ref# class has a template parameter
+[code]#DefaultScope#, which allows the application to define a default memory
+scope for the atomic operations.  Each member function also provides an
+optional parameter that allows the application to override this default.
 
-The template parameter [code]#Space# is permitted to be
-[code]#access::address_space::generic_space#,
-[code]#access::address_space::global_space# or [code]#access::address_space::local_space#.
+The [code]#sycl::atomic_ref# class also has a template parameter
+[code]#AddressSpace#, which allows the application to make an assertion about
+the address space of the object of type [code]#T# that it references.  The
+default value for this parameter is
+[code]#access::address_space::generic_space#, which indicates that the object
+could be in either the global or local address spaces.  If the application
+knows the address space, it can set this template parameter to either
+[code]#access::address_space::global_space# or
+[code]#access::address_space::local_space# as an assertion to the
+implementation.  Specifying the address space via this template parameter may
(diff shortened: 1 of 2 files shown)


keryell (approved):
Thanks for the clarification.

adoc/chapters/programming_interface.adoc
```diff
@@ -17945,43 +17945,77 @@ device_event(___unspecified___)
...
+The [code]#sycl::atomic_ref# class provides the ability to perform atomic
+operations in device code with a syntax similar to the {cpp} standard
+[code]#std::atomic_ref#.  The [code]#sycl::atomic_ref# class must not be used
+in host code.
+
+Unlike [code]#std::atomic_ref#, [code]#sycl::atomic_ref# does not provide a
+default memory ordering for its operations.  Instead, the application must
+specify a default ordering via the [code]#DefaultOrder# template parameter.
+This ordering is used as a default for most of the atomic operations, but
+each member function also provides an optional parameter that allows the
```
Pennycook, line 17957: Do you think we need to state explicitly that the arithmetic operators always use the default and are always read-modify-write operations?  Or is that clear enough from the way we define them in the table (e.g. that `operator++` is equivalent to `fetch_add(1)`)?

adoc/chapters/programming_interface.adoc
```diff
@@ -17945,43 +17945,77 @@ device_event(___unspecified___)
...
+[code]#info::device::double_fp_config# (i.e. floating-point atomics may use
+different rounding modes and may have different exception behavior).
+
+If the code using [code]#sycl::atomic_ref# is submitted to a device that has
+[code]#aspect::usm_atomic_host_allocations# or
+[code]#aspect::usm_atomic_shared_allocations#, then atomic operations in device
+code (using [code]#sycl::atomic_ref#) are guaranteed to be atomic with respect
+to atomic operations performed in host code (using [code]#std::atomic_ref#),
+when they operate on the same object and when the memory scope is
+[code]#memory_scope::system#.
```
Pennycook, line 18018: I know this PR is focused on `atomic_ref`, but should we add similar clarifications for the interaction between `sycl::atomic_fence` and `std::atomic_thread_fence`, and the interaction between fences and atomic references?

One simple way to do this might be to lift this paragraph to the "Synchronization and atomics" section and add something like:

> Additionally, atomic operations in device code are guaranteed to be able to synchronize with fence operations performed in host code (using [code]#std::atomic_thread_fence#), and atomic operations in host code are guaranteed to be able to synchronize with fence operations performed in device code (using [code]#sycl::atomic_fence#), when the memory scope is [code]#memory_scope::system#.

I suggest "guaranteed to be able to" rather than "guaranteed to" in the above because fences only synchronize with atomics and other fences under certain conditions (see [atomics.fences](http://eel.is/c++draft/atomics.fences)).

adoc/chapters/programming_interface.adoc
gmlueck: I think we can rely on the description in the table for this level of detail.

I could change the wording to "most member functions also provide ..." if we think that's better.  I guess it depends on whether the term "member function" also includes the operators.

adoc/chapters/programming_interface.adoc
gmlueck: Yes, something like this probably makes sense.  It seems weird, though, to put that description at the start of the [section 4.15](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_synchronization_and_atomics) before we talk about either `sycl::atomic_fence` or `sycl::atomic_ref`.  Maybe it would be better to add a new section 4.15.5 "Synchronization between host and device code" where we describe this.

adoc/chapters/programming_interface.adoc
Pennycook: Yeah, that's what I was unsure of.  I did some digging and it looks like an "operator function" declared inside a class is considered a "member function" (https://eel.is/c++draft/over.oper.general#7).

Your suggested fix sounds good.

adoc/chapters/programming_interface.adoc
Pennycook: I agree, a new section would be better.

gmlueck: My proposed wording is in 5ef7fe95aee51656a85cb7025254892503f3604a.

Pennycook: I think this looks very good.  I'm worried I'm being a little pedantic, but I think we should consider dropping the "share the same atomic object _M_" bit from the last three paragraphs.

I don't think it's obvious what it means to "share the same atomic object _M_", because a fence doesn't directly operate on any memory objects.

I suggest a simpler "if the fence operations follow the rules for fence synchronization defined in the {cpp} core language.".  This still requires certain orderings of fence operations and atomic operations, but which orderings lead to synchronization is covered entirely by C++.

gmlueck: I'd like to keep that because it helps clarify why these fence operations only provide a guarantee of synchronization between the host and device when the device supports `aspect::usm_atomic_host_allocations` or `aspect::usm_atomic_shared_allocations`.  (Since C++ defines the fence operations in terms of atomic operations on some shared object _M_, there is only a synchronization guarantee when it is possible to have some shared atomic object _M_.)

I agree that the wording would be unclear to a casual reader because you would have to read the C++ standard to know about this relationship between fence operations and a shared atomic object.  I tried to fix this in 7d52276711524cae6596e3623cf8885426175c39 by clarifying that this relationship exists.

Pennycook (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -17945,43 +17945,69 @@ device_event(___unspecified___)
 [[sec:atomic-references]]
 === Atomic references
 
-The SYCL specification provides atomic operations based on the library
-syntax from the next {cpp} specification. The set of supported orderings is
+The [code]#sycl::atomic_ref# class provides the ability to perform atomic
+operations in device code with a syntax similar to the {cpp} standard
+[code]#std::atomic_ref#.  The [code]#sycl::atomic_ref# class must not be used
+in host code.
```
AerialMantis, line 17951: When we say host code here, do we mean application scope, or does this also extend to host tasks?

nliber (approved):

fraggamuffin:
LGTM

AerialMantis (commented):
Looks good to me, just a comment about the scope of the std::atomic_ref/sycl::atomic_ref interop.

adoc/chapters/programming_interface.adoc
gmlueck: The prohibition also extends to host tasks.  Therefore, I think "application scope" is not the right term here.

AerialMantis: That makes sense, just wanted to clarify, thanks.

AerialMantis (approved):

fraggamuffin:
LGTM

PR #359, opened 2023-02-08
Clarify queue profiling behavior when unsupported

gmlueck:
After a recent question from our development team, I realized that we did not specify the behavior when a queue is constructed with the `property::queue::enable_profiling` property when the device does not support profiling (`aspect::queue_profiling` is false).

I think it is natural for the queue constructor to fail in this case, and I clarified the spec to say this.


keryell (approved):
That makes sense!

TApplencourt (approved):

fraggamuffin:
LGTM

PR #370, opened 2023-03-03
Clarify any_device_has / all_devices_have

gmlueck:
We determined during testing that the description of these traits wasn't clear.  These traits can't tell you the set of devices that are available at runtime.  They can only tell you what devices are supported by the code that the compiler generates.  Make this more clear.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2833,14 +2833,13 @@ aspect::usm_system_allocations
 |====
 
 The implementation also provides two traits that the application can use to
-query aspects at compilation time.  The trait [code]#any_device_has<aspect>#
-inherits from [code]#std::true_type# if the compilation environment supports
-any device which has the specified aspect, and it inherits from
-[code]#std::false_type# if no device has the aspect.  The trait
-[code]#all_devices_have<aspect># inherits from [code]#std::true_type# if all
-devices supported by the compilation environment have the specified aspect,
-and it inherits from [code]#std::false_type# if any device does not have the
-aspect.
+query aspects at compilation time.  The traits [code]#any_device_has<aspect>#
+and [code]#all_devices_have<aspect># are set according to the collection of
+devices _D_ that can possibly execute device code, as determined by the
+compilation environment.  The trait [code]#any_device_has<aspect># inherits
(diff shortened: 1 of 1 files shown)


steffenlarsen (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -2833,14 +2833,13 @@ aspect::usm_system_allocations
...
-inherits from [code]#std::true_type# if the compilation environment supports
-any device which has the specified aspect, and it inherits from
-[code]#std::false_type# if no device has the aspect.  The trait
-[code]#all_devices_have<aspect># inherits from [code]#std::true_type# if all
-devices supported by the compilation environment have the specified aspect,
-and it inherits from [code]#std::false_type# if any device does not have the
-aspect.
+query aspects at compilation time.  The traits [code]#any_device_has<aspect>#
+and [code]#all_devices_have<aspect># are set according to the collection of
+devices _D_ that can possibly execute device code, as determined by the
```
TApplencourt, line 2838: Nitpicking, but should we say `_Ds_` or `devs`  here?

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
gmlueck: You are suggesting that we use the italic name _Ds_ instead of _D_ when referring to a set of devices?  I don't have a strong opinion, but there are other places where we use a singular letter to refer to a set of things.  For example, [section 3.7.1.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_command_groups_and_execution_order) "SYCL command groups and execution order" says:

> A SYCL command group object defines a set of requisites (R) ...

Where "R" is singular.

TApplencourt: Then nevermind! 

It was just that I interpreted the `the collection of devices _D_`  as a collection of devices, where the devices inside this collection are individually named `D`, and not that the collection itself is named `D`. Aka `set<D>` and not `D=set<device>` . Witch makes the sentence in the below paragraph `if all device in _D_` nonsensical. 

But it's just me who cannot read English.

keryell (approved):
Thanks.

PR #375, opened 2023-03-08
Small clarification to generic address space

gmlueck:
Clarify that objects in global, local, or private address space can also be accessed via the generic address space.


keryell (approved):
Thanks.

PR #381, opened 2023-03-15
Clarify "is_compatible"

gmlueck:
This PR addresses an issue that was raised during the review of KhronosGroup/SYCL-CTS#553.  It is also related to the clarification made in #370.

When we clarified the semantics of `any_device_has` and `all_devices_have`, we decided that these tell information about the "compilation environment" used to compile the SYCL translation unit. This implies that kernels in a TU might not be compatible with some devices depending on the "compilation environment".  For example, a vendor's compiler might support some compiler flag that allows kernels to run on GPU devices but not on other devices.

While writing the CTS, we realized that applications need some way to tell if a kernel was compiled in a way that restricts the devices on which it can run.  It seems like the existing API `is_compatible` is a good fit, so this PR extends the semantics to say that this API also considers the "compilation environment" when determining compatibility.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -14999,7 +14999,12 @@ application is compatible with a device unless:
   in <<sec:optional-kernel-features>>; or
 * It is decorated with a [code]#[[sycl::device_has()]]# {cpp} attribute that
   lists an aspect that is not supported by the device, as described in
-  <<sec:kernel.attributes>>.
+  <<sec:kernel.attributes>>; or
+* The translation unit containing the kernel was compiled in a compilation
+  environment that does not support the device.  Each implementation defines
+  the specific criteria for which devices are supported in its compilation
+  environment.  For example, this might be dependent on options passed to the
+  compiler.
 
 A device built-in kernel is only compatible with the device for which it is
 built-in.


rasoolmaghareh (approved):

keryell:
Need more reviews.

adoc/chapters/programming_interface.adoc
```diff
@@ -14999,7 +14999,12 @@ application is compatible with a device unless:
   in <<sec:optional-kernel-features>>; or
 * It is decorated with a [code]#[[sycl::device_has()]]# {cpp} attribute that
   lists an aspect that is not supported by the device, as described in
-  <<sec:kernel.attributes>>.
+  <<sec:kernel.attributes>>; or
```
AlexeySachkov, line 15002: `is_compatible` doesn't really sound like "was this kernel compiled for this device?". With this change, users won't have an easy way to understand whether or not all optional features used in a kernel are supported by the device, because `is_compatible` could return `false` for other reasons.

Would it be better to instead expand definition of `has_kernel_bundle` to say that it may return `false` if compilation environment doesn't support the device?

Kernel definition/properties, i.e. list of aspects used by it should not change depending on compilation environment, because it is encoded in the source code. Content of kernel bundle and list of available device images, however, is not defined by the source code alone and can also be heavily dependent on compilation environment.

Tagging @rolandschulz

gmlueck: I think the intent of `is_compatible` was that you can use it to determine whether a kernel can be submitted to a device.  If we think that was the intent, then I believe this clarification makes sense.

I agree that an application could have some kernel _K_ that uses features that are compatible with some device _D_, but due to the compilation environment _K_ cannot be run on device _D_.  This change to the spec makes it impossible for the application to test that _K_ would be compatible with device _D_ if the application was compiled with some other set of options.  However, I'm not sure that such a query is useful.  It seems like the thing the application cares about is whether _K_ can be run on _D_, not some hypothetical question about whether _K_ could be run on _D_ if the application was compiled differently.

Note that it's less convenient to use `has_kernel_bundle` because the application need to create a kernel bundle in order to use that API.  It's much more convenient to use `is_compatible` because you only need the kernel's type-name.

AlexeySachkov: > However, I'm not sure that such a query is useful. It seems like the thing the application cares about is whether K can be run on D, not some hypothetical question about whether K could be run on D if the application was compiled differently.

Let's say you provide a library which is AOT-compiled for a certain list of devices. You may want to improve user experience of the library users by emitting a better error saying that they need to re-compile the library with different flags to make it work on that device; instead of having some unspecified exception at runtime.
However, I agree that it is kind of an artificial case, because as a library author you could somehow configure your compilation environment to ensure that there are some fallback paths for non-target devices; or simply have clear documentation about device requirements.

> Note that it's less convenient to use has_kernel_bundle because the application need to create a kernel bundle in order to use that API. 

You don't need a kernel bundle to query if a kernel bundle exists. `has_kernel_bundle` accepts context, vector of `kernel_id` and optionally list of devices.

keryell (approved):

adoc/chapters/programming_interface.adoc
AerialMantis: I think I agree that since there are a few different reasons for why `is_compatible` may return false, it would be useful to have a way to expose the reason, I'm not sure how useful it would be as a mechanism for dispatching different kernels, but it could at least be a useful debug feature, to be able to identify exactly why the kernel couldn't be run on a certain device. Saying that, if we were to do this, I think that could be a different PR that doesn't need to hold up this one.

adoc/chapters/programming_interface.adoc
```diff
@@ -14999,7 +14999,12 @@ application is compatible with a device unless:
...
+* The translation unit containing the kernel was compiled in a compilation
+  environment that does not support the device.  Each implementation defines
```
AerialMantis, line 15004: Minor comment, should we phrase this as `... that is not supported by the device` to align the phrasing with the statements above?

AerialMantis (approved):

adoc/chapters/programming_interface.adoc
gmlueck: That phrasing doesn't seem accurate, though.  A device doesn't support a compilation environment, it's the other way around.

fraggamuffin:
delay until remaining open discussion closes

tomdeakin:
WG: ready to merge

PR #382, opened 2023-03-16
Clarify queries for memory scope / order caps

gmlueck:
This change clarifies some questions that arose during implementation in DPC++.

* Clarify that the query operations for supported memory scopes must return all "narrower" scopes when a "wider" scope is supported.  For example, if the device supports `memory_scope::sub_group` it must also support `memory_scope::work_item`.  This matches the OpeCL queries, which also returns all "narrower" scopes.

* Clarify that the query operations for supported memory orders must return all "weaker" orders when a "stronger" order is supported. Again, this matches OpenCL.

* Clarify in the architecture chapter that a device does not need to support every memory scope.  We said this for memory orders, but not for memory scopes.

* Clarify the wording for `context::atomic_memory_order_capabilities` and `context::atomic_fence_order_capabilities`.  These provide information about memory orders for *inter-device* operations, which could be a subset of the memory orders supported for operations within the device.  We said this before, but more tersely.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1033,10 +1033,9 @@ The meanings of these values are identical to those defined in the {cpp} core
 language.
 
 The complete set of memory orders is not guaranteed to be supported by every
-device, nor across all combinations of devices within a context.  The memory
-orders supported by a specific device and context can be queried using
-functionalities of the [code]#sycl::device# and [code]#sycl::context# classes,
-respectively.
+device, nor across all combinations of devices within a context.  The set of
+supported memory orders can be queried via the information descriptors for the
+[code]#sycl::device# and [code]#sycl::context# classes.
 
 [NOTE]
 ====
@@ -1071,6 +1070,10 @@ values:
     work-item or host thread in the system that is currently permitted to
     access the memory allocation containing the referenced object, as
     defined by the capabilities of <<buffer,buffers>> and <<usm>>.
+
+The complete set of memory scopes is not guaranteed to be supported by every
+device.  The set of supported memory scopes can be queried via the information
+descriptors for the [code]#sycl::device# and [code]#sycl::context# classes.
 
 The broadest scope that can be applied to an atomic operation corresponds to
 the set of work-items which can access the associated memory location.  For
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1012,6 +1012,7 @@ supporting concurrent atomic accesses to USM allocations and acquire-release or 
 consistent memory orderings, cross-device memory consistency can be
 enforced through the use of <<mem-fence>> and atomic operations.
 
+[[sec:memory-ordering]]
 ==== Memory ordering
 
 [source,,linenums]
@@ -1032,6 +1033,9 @@ values:
 The meanings of these values are identical to those defined in the {cpp} core
 language.
 
+These memory orders are listed above from weakest
+([code]#memory_order::relaxed#) to strongest ([code]#memory_order::seq_cst#).
+
 The complete set of memory orders is not guaranteed to be supported by every
 device, nor across all combinations of devices within a context.  The set of
 supported memory orders can be queried via the information descriptors for the
@@ -1045,6 +1049,7 @@ device kernel results in undefined behavior. Developers are encouraged to use
 [code]#sycl::memory_order::acquire# instead.
 ====
 
+[[sec:memory-scope]]
 ==== Memory scope
 
 [source,,linenums]
@@ -1071,15 +1076,18 @@ values:
     access the memory allocation containing the referenced object, as
     defined by the capabilities of <<buffer,buffers>> and <<usm>>.
 
+The memory scopes are listed above from narrowest
+([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
+
 The complete set of memory scopes is not guaranteed to be supported by every
 device.  The set of supported memory scopes can be queried via the information
 descriptors for the [code]#sycl::device# and [code]#sycl::context# classes.
 
-The broadest scope that can be applied to an atomic operation corresponds to
+The widest scope that can be applied to an atomic operation corresponds to
 the set of work-items which can access the associated memory location.  For
-example, the broadest scope that can be applied to atomic operations in
+example, the widest scope that can be applied to atomic operations in
 work-group local memory is [code]#sycl::memory_scope::work_group#.  If a
-broader scope is supplied, the behavior is as-if the narrowest scope containing
+wider scope is supplied, the behavior is as-if the narrowest scope containing
 all work-items which can access the associated memory location was supplied.
 
 [NOTE]
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1037,7 +1037,7 @@ These memory orders are listed above from weakest
 ([code]#memory_order::relaxed#) to strongest ([code]#memory_order::seq_cst#).
 
 The complete set of memory orders is not guaranteed to be supported by every
-device, nor across all combinations of devices within a context.  The set of
+device, nor across all combinations of devices within a platform.  The set of
 supported memory orders can be queried via the information descriptors for the
 [code]#sycl::device# and [code]#sycl::context# classes.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1428,9 +1428,11 @@ info::context::atomic_memory_order_capabilities
 
 Returns the set of memory orders supported by these atomic operations.  When a
 context returns a "stronger" memory order in this set, it must also return all
-"weaker" memory orders.  The memory orders [code]#memory_order::acquire#,
-[code]#memory_order::release#, and [code]#memory_order::acq_rel# are all the
-same strength.  If a context returns one of these, it must return them all.
+"weaker" memory orders.  (See <<sec:memory-ordering>> for a definition of
+"stronger" and "weaker" memory orders.) The memory orders
+[code]#memory_order::acquire#, [code]#memory_order::release#, and
+[code]#memory_order::acq_rel# are all the same strength.  If a context returns
+one of these, it must return them all.
 
 At a minimum, each context must support [code]#memory_order::relaxed#.
 
@@ -1449,7 +1451,8 @@ info::context::atomic_fence_order_capabilities
 
 Returns the set of memory orders supported by these [code]#atomic_fence#
 operations.  When a context returns a "stronger" memory order in this set, it
-must also return all "weaker" memory orders.
+must also return all "weaker" memory orders.  (See <<sec:memory-ordering>> for
+a definition of "stronger" and "weaker" memory orders.)
 
 At a minimum, each context must support [code]#memory_order::relaxed#,
 [code]#memory_order::acquire#, [code]#memory_order::release#, and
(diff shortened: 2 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -2258,9 +2287,12 @@ info::device::atomic_fence_order_capabilities
...
     @ [.code]#std::vector<memory_order>#
-   a@ Returns the set of memory orderings supported by [code]#atomic_fence# on the device,
-      which is guaranteed to include [code]#relaxed#, [code]#acquire#,
-      [code]#release# and [code]#acq_rel#.
+   a@ Returns the set of memory orders supported by [code]#atomic_fence# on
+      the device.  When a device returns a "stronger" memory order in this set,
+      it must also return all "weaker" memory orders.  At a minimum, each
+      device must support [code]#memory_order::relaxed#,
+      [code]#memory_order::acquire#, [code]#memory_order::release#, and
+      [code]#memory_order::acq_rel#.
```
keryell, line 2295: Is it true?

adoc/chapters/programming_interface.adoc
```diff
@@ -1447,7 +1462,11 @@ info::context::atomic_memory_scope_capabilities
 ----
 
     @ [.code]#std::vector<memory_scope>#
-   a@ Returns the set of memory scopes supported by atomic operations on all devices in the context, which is guaranteed to include [code]#work_group#.
+   a@ Returns the set of memory scopes supported by atomic operations on all
+      devices in the context.  When a context returns a "wider" memory scope
+      in this set, it must also return all "narrower" memory scopes.  At a
+      minimum, each context must support [code]#memory_scope::work_item#,
+      [code]#memory_scope::sub_group#, and [code]#memory_scope::work_group#.
```
keryell, line 1469: Would it be possible for a device to not have all of them?

adoc/chapters/programming_interface.adoc
```diff
@@ -1420,11 +1420,19 @@ info::context::atomic_memory_order_capabilities
...
 
-The memory ordering of the context determines the behavior of atomic operations applied
-to any memory that can be concurrently accessed by multiple devices in the context.
+Returns the set of memory orders supported by these atomic operations.  When a
+context returns a "stronger" memory order in this set, it must also return all
+"weaker" memory orders.  The memory orders [code]#memory_order::acquire#,
+[code]#memory_order::release#, and [code]#memory_order::acq_rel# are all the
+same strength.  If a context returns one of these, it must return them all.
+
+At a minimum, each context must support [code]#memory_order::relaxed#.
```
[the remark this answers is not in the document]
gmlueck: We are following OpenCL here, which has the same concept of "strength":

> Because atomic memory orders are hierarchical, a device that supports a strong memory order must also support all weaker memory orders.

(https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES)

adoc/chapters/programming_interface.adoc
gmlueck: Again, we are following OpenCL here, which has the same concept of "wide" and "narrow" memory scopes:

> Because atomic scopes are hierarchical, a device that supports a wide scope must also support all narrower scopes, except for the work-item scope, which is a special case.

(https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES)

The new wording does not require support for any "wider" scope than before.  The previous wording also required a device to support `work_group` scope.  The question from our implementation team was whether this meant that the device must also support the narrower scopes.  We thought that we should follow OpenCL in this regard and require support for the narrower scopes too.

Note that we do not adopt the OpenCL language about "except for the work-item scope, which is a special case" because work-item scope has a special meaning in OpenCL related to read/write images.  SYCL does not have that issue because we do not have read/write images, so there is no special semantic regarding `work_item` scope.

adoc/chapters/programming_interface.adoc
gmlueck: I think this is the same question you are asking above, so we can discuss it there.

adoc/chapters/programming_interface.adoc
```diff
@@ -1433,12 +1441,19 @@ info::context::atomic_fence_order_capabilities
...
-      [code]#relaxed#, [code]#acquire#, [code]#release# and [code]#acq_rel#.
+   a@ This query applies only to the capabilities of [code]#atomic_fence# when
+      applied to memory that can be concurrently accessed by multiple devices
+      in the context.  If these capabilities are not uniform across all devices
+      in the context, the query reports only the capabilities that are common
+      for all devices.
+
+Returns the set of memory orders supported by these [code]#atomic_fence#
+operations.  When a context returns a "stronger" memory order in this set, it
+must also return all "weaker" memory orders.
```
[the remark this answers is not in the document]
keryell: Now I realize using a `std::vector` is kind of overkill. To generalize for SYCL Next... I hope nobody will use this in the critical path anyway.

adoc/chapters/programming_interface.adoc
```diff
@@ -1433,12 +1441,19 @@ info::context::atomic_fence_order_capabilities
...
 
-The memory ordering of the context determines the behavior of fence operations applied
-to any memory that can be concurrently accessed by multiple devices in the context.
+At a minimum, each context must support [code]#memory_order::relaxed#,
+[code]#memory_order::acquire#, [code]#memory_order::release#, and
+[code]#memory_order::acq_rel#.
```
keryell, line 1456: All these? A fence with a relaxed memory order might be useless. But can we imagine a device specialized to produce data while another one could only consume?

adoc/chapters/programming_interface.adoc
```diff
@@ -2248,8 +2271,14 @@ info::device::atomic_memory_order_capabilities
...
 
     @ [.code]#std::vector<memory_order>#
-   a@ Returns the set of memory orderings supported by atomic operations on the device,
-      which is guaranteed to include [code]#relaxed#.
+   a@ Returns the set of memory orders supported by atomic operations on the
+      device.  When a device returns a "stronger" memory order in this set, it
+      must also return all "weaker" memory orders.  The memory orders
+      [code]#memory_order::acquire#, [code]#memory_order::release#, and
+      [code]#memory_order::acq_rel# are all the same strength.  If a device
+      returns one of these, it must return them all.
```
keryell, line 2279: As I said, would it be possible to have one device having only acquire while another one only release?

adoc/chapters/programming_interface.adoc
gmlueck: This PR does not add any new requirements here.  Even the old wording required all of these memory orderings.

adoc/chapters/programming_interface.adoc
gmlueck: This query (`info::device::atomic_memory_order_capabilities`) tells the operations that are supported within a single device, not between different devices.  Within a single device, it does not make sense to have "acquire" without "release".

The query `info::context::atomic_memory_order_capabilities` tells the capabilities between devices.  I agree that this query cannot express the situation where one device can do only "acquire" operations while another device can do only "release" operations.  We would need some completely different query to express this, though.  The existing `info::context::atomic_memory_order_capabilities` query can only describe operations that are allowed across a set of devices.  It cannot describe a situation where each device has a different set of allowable operations.  Fixing this is beyond the scope of this PR.

keryell (approved):
Thanks.
Anything else more complex can be handled by extensions anyway.

fraggamuffin:
need another approver

fraggamuffin:
Please review before next meeting.

tomdeakin:
WG: One more review

fraggamuffin:
GB will review.

adoc/chapters/architecture.adoc
```diff
@@ -1072,6 +1071,10 @@ values:
     access the memory allocation containing the referenced object, as
     defined by the capabilities of <<buffer,buffers>> and <<usm>>.
 
+The complete set of memory scopes is not guaranteed to be supported by every
```
AerialMantis, line 1074: Should this paragraph also have the "nor across all combinations of devices within a context", to account for the case where some devices in a platform may support system scope, where as another might be more limited?

adoc/chapters/architecture.adoc
```diff
@@ -1033,10 +1033,9 @@ The meanings of these values are identical to those defined in the {cpp} core
 language.
 
 The complete set of memory orders is not guaranteed to be supported by every
-device, nor across all combinations of devices within a context.  The memory
-orders supported by a specific device and context can be queried using
-functionalities of the [code]#sycl::device# and [code]#sycl::context# classes,
-respectively.
+device, nor across all combinations of devices within a context.  The set of
```
AerialMantis, line 1036: I wonder if we should describe this in terms of combinations of devices from the same platform, it's somewhat saying the same thing, but since this is querying static capabilities of the platform rather than the state of a context, this might better reflect that.

adoc/chapters/programming_interface.adoc
```diff
@@ -1420,11 +1420,19 @@ info::context::atomic_memory_order_capabilities
...
+   a@ This query applies only to the capabilities of atomic operations that are
+      applied to memory that can be concurrently accessed by multiple devices
+      in the context.  If these capabilities are not uniform across all devices
+      in the context, the query reports only the capabilities that are common
+      for all devices.
 
-The memory ordering of the context determines the behavior of atomic operations applied
-to any memory that can be concurrently accessed by multiple devices in the context.
+Returns the set of memory orders supported by these atomic operations.  When a
+context returns a "stronger" memory order in this set, it must also return all
```
AerialMantis, line 1430: We're using the terms stronger/weaker and wider/narrower, but I don't think we define what these mean, for these to make sense I think we need to define a total ordering for the `memory_order` and `memory_scope`.

gmlueck: Good point.  I'll add definitions for "stronger/weaker" to section [3.8.3.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_ordering) and definitions for "wider/narrower" to section [3.8.3.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_scope).

gmlueck: Addressed in 07393a9249343f061159cbd0b1f4fd16eeda858d.

adoc/chapters/architecture.adoc
gmlueck: Addressed in 30fa3952949391f2768cce819e9c18015e6b5f5b.

keryell (approved):
Aprove++

adoc/chapters/architecture.adoc
AerialMantis: > I think the current spec wording says that the application is not required to check whether D2 supports memory_scope::system. Instead, I think the implementation is required to support atomic synchronization between D1 and D2 because D1 reported memory_scope::system:

So yeah this was the case that I had in mind, I can imagine a theoretical platform where there are multiple devices, say D1 and D2 which provide `mermory_scope::system` atomic synchronization, in that both D1 and D2 can access and modify some object in shared memory concurrently with the host, but not with each other.

Furthermore, this could be extended to a platform where there are subsets of devices that share the same memory scope for synchronization, but not all, say D1, D2, and D3, where D1 and D2 provide `mermory_scope::system` atomic synchronization w.r.t each other but not w.r.t D3.

Obviously, this is all theoretical, and I'm not aware of any specific architectures where this is the case, but I wonder if we need to account for this. I think we perhaps should revise the concept of system-level synchronization such that it doesn't implicitly refer to synchronization w.r.t all other devices in the platform, and instead refers to the relationships between a group of devices and with the host.

I think the wording you suggest makes sense, though how do we define "the atomic operation is to memory that the device can concurrently access"? We currently don't currently specify device-to-device access and I don't think we can use the context as the basis for this, as the criteria for devices being in the same context is too broad.

As a side, I didn't mean to derail this PR, if you'd like I'm happy to leave the wording as you have it and open a new issue to continue this.

AerialMantis (approved):

adoc/chapters/architecture.adoc
gmlueck: > As a side, I didn't mean to derail this PR, if you'd like I'm happy to leave the wording as you have it and open a new issue to continue this.

Yes, I think this is better.  In order to solve the problem you are pointing out, we would need to add some new peer-to-peer API that queries about atomic access between arbitrary devices.  This is beyond the scope of the clarification I want to make in this PR.

AerialMantis: I've created a new issue for this - https://github.com/KhronosGroup/SYCL-Docs/issues/429.

Issue #383, opened 2023-03-19
Description of info::device::partition_type_property descriptor is incorrect


oleksandr-pavlyk:
The [description](https://github.com/KhronosGroup/SYCL-Docs/blob/SYCL-2020/master/adoc/chapters/programming_interface.adoc?plain=1#L2591-L2603) of `info::device::partitition_type_property` descriptor says:

> Returns the partition property of this SYCL [code]#device#. If this SYCL [code]#device# is not a sub device then the return value must be [code]#info::partition_property::no_partition#, otherwise it must be one of the following values:

The part "If this SYCL [code]#device# is not a sub device then the return value must be ..." is incorrect. 

It should say "If this SYCL [code]#device# can not be further partitioned then the return value must be..." instead.

gmlueck:
I think the current wording in the spec is correct.  It looks like you are confusing `info::device::partitition_type_property` with `info::device::partition_properties`.

The info descriptor `info::device::partition_properties` tells you how a device can be partitioned into sub-devices.  The info descriptor `info::device::partitition_type_property` tells you how a sub-device was created from its parent device (i.e. what type of partitioning).

oleksandr-pavlyk:
Thank you for the explanation @gmlueck. I will close the PR and this issue.

Pennycook:
Since there was some confusion here, should we clarify what these queries return?  I don't think the names are very clear.

Instead of:

> Returns the partition property of this SYCL device. 

What about:

> Returns the `info::partition_property` that was used to create this sub-device.  If the SYCL `device` is not a sub-device...

...with an equivalent change to the wording of `partition_type_affinity_domain`?

PR #384, opened 2023-03-19
Closes gh-383, corrects description of partition_type_property

oleksandr-pavlyk:
Closes gh-383, corrects description of ``info::device::partition_type_property`` in correcting the condition when the value of the descriptor must be ``info::partition_property::no_partition``.


oleksandr-pavlyk:
Closing since the original spec wording is correct as explained by @gmlueck in gh-383

PR #387, opened 2023-03-21
Clarify is_group and bool_constant alias relations

steffenlarsen:
This commit changes the wording for `is_group` to specify that it inherits from `std::true_type` or `std::false_type`, instead of stating that they are the same. This is in line with other traits in the specification.


gmlueck (approved):

nliber (approved):

keryell (approved):
Good catch.

PR #394, opened 2023-04-11
Editorial cleanup of multi-ptr

gmlueck:
This commit is preparation for a larger change to `multi_ptr` that I'm working on.  All of the changes in this commit are editorial cleanup. There are no meaningful changes to the specification:

* Add line breaks to the table of `multi_ptr` functions to avoid line wrap.
* Use `std::is_void_v<T>` instead of `std::is_void<T>::value`.
* Change the "Only if" comments in the `multi_ptr` synopsis header to use real C++ code to more precisely define the condition under which the various functions are available.
*  Consistently use the name `IsDecorated` for the template parameter of type `access::decorated`.  Previously, some member functions named this parameter `DecorateAddress2`.
* Correct the title of a table in the `accessor` section.

### DIFF
--- a/adoc/headers/multipointer.h
+++ b/adoc/headers/multipointer.h
@@ -135,17 +135,17 @@ class multi_ptr {
 
   // Implicit conversion to a multi_ptr<void>.
   // Available only when: (!std::is_const_v<value_type>)
-  template <access::decorated DecorateAddress2>
-  operator multi_ptr<void, Space, DecorateAddress2>() const;
+  template <access::decorated IsDecorated>
+  operator multi_ptr<void, Space, IsDecorated>() const;
 
   // Implicit conversion to a multi_ptr<const void>.
   // Available only when: (std::is_const_v<value_type>)
-  template <access::decorated DecorateAddress2>
-  operator multi_ptr<const void, Space, DecorateAddress2>() const;
+  template <access::decorated IsDecorated>
+  operator multi_ptr<const void, Space, IsDecorated>() const;
 
   // Implicit conversion to multi_ptr<const value_type, Space>.
-  template <access::decorated DecorateAddress2>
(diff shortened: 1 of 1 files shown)


TApplencourt:
LGTM!

PS: I don't know if you did the re-formating of `code block` by hand, but just in case, I wrote this little script back in the day (https://github.com/KhronosGroup/SYCL-Docs/pull/292#issuecomment-1270531337) to do it automatically. Maybe you can find it helpful too.

TApplencourt (approved):

keryell (approved):
Thanks!

gmlueck:
> I don't know if you did the re-formating of `code block` by hand, but just in case, I wrote this little script back in the day ([#292 (comment)](https://github.com/KhronosGroup/SYCL-Docs/pull/292#issuecomment-1270531337)) to do it automatically. Maybe you can find it helpful too.

Thanks for noting this.  I did do the formatting by hand in this PR.  Does your script split lines to avoid line wrap, even for code blocks that we have in the tables?  How does your script know the maximum line length?  This varies for each table, depending on the column widths.

TApplencourt:
> How does your script know the maximum line length? This varies for each table, depending on the column widths.

Oh. I didn't know that. The script only handles `[source]`, not the `[code]` one. Or maybe the default `line length` of LLVM format is short enough, so it didn't triger the issue. 

How do you know the size to use? Trial and error until it looks good? I can try to parse the `[cols="]` and do some heuristic based on that but it sound painful. I can also update the script to avoid any code block inside tables. But sadly, it's maybe the most of our usage.

gmlueck:
> How do you know the size to use? Trial and error until it looks good?

Yes, exactly.  Specifically, I create the PDF render and adjust line breaks until there is no wrap in the PDF render.

fraggamuffin:
LGTM

PR #395, opened 2023-04-12
Multi-ptr bug fixes, phase 1

gmlueck:
This is the first of several commits to fix bugs in the specification of `multi_ptr`.  This phase aligns the synopsis of the `multi_ptr` class with the table describing each member function, focusing just on the primary class synopsis (not the specialization for `void` types).

Describing the changes that are not obvious bugs:

* The table specified the two conversion operators to the "no" and "yes" version of the multi-pointer like this:

  ```
  operator multi_ptr<const value_type, Space, access::decorated::no>() const

  operator multi_ptr<const value_type, Space, access::decorated::yes>() const
  ```

  The `const value_type` seems weird because it would only allow the conversion if the underlying pointer is also changed to `const`. This does not match the synopsis, so remove the `const` bringing the description in the table into alignment with the synopsis.

* The synopsis specified several conversion operators that converted from `generic_space` to the other three address spaces:

  ```
  explicit operator multi_ptr<value_type, access::address_space::private_space,
                              DecorateAddress>();
  ```

  However, the table specified these conversions as template operators, allowing them to also change the decoration:

  ```
  template <access::decorated IsDecorated>
  explicit operator multi_ptr<value_type, access::address_space::private_space,
                              IsDecorated>() const
  ```

  In addition, the versions in the table are `const` operators, while the versions in the synopsis were not.

  In this case, I kept the versions in the table and updated the synopsis.  It makes sense to me that it should be possible to change both the address space and the decoration with a single conversion. I'm not sure why the synopsis versions of the operators were missing `const`; I assume this was just a typo.

* The table defines the `prefetch` function with the restriction that it is only available for multi-pointers that are `generic_space`.  By contrast, the synopsis did not have this restriction.

  I also kept the version in the table in this case.  The description says that this function prefetches in the "global memory cache", so I think we intended it to be available only for `global_space`.

* The template parameters for the deduction guide from `local_accessor` made no sense.  I think this was a copy / paste error, and I corrected them in the obvious way.


gmlueck:
I made this PR "draft" only because I'd like our development team to implement this first and make sure there aren't any surprises.    I'll remove "draft" once that happens.  In the meantime, any comments are welcome.

adoc/chapters/programming_interface.adoc
```diff
@@ -9372,7 +9372,7 @@ operator multi_ptr<const value_type, Space, IsDecorated>() const
 a@
 [source]
 ----
-operator multi_ptr<const value_type, Space, access::decorated::no>() const
+operator multi_ptr<value_type, Space, access::decorated::no>() const
```
keryell, line 9375: Good catch.

keryell (approved):
Thanks.

gmlueck:
> I made this PR "draft" only because I'd like our development team to implement this first and make sure there aren't any surprises. I'll remove "draft" once that happens.

It turns out our development team won't focus on `multi_ptr` for a little while.  I'm pretty confident that these changes are correct and are certainly an improvement over the current state of the spec.  I've removed "draft" status.

tomdeakin:
One more review needed

adoc/headers/multipointer.h
```diff
@@ -155,6 +161,7 @@ class multi_ptr {
   // Available only when: (is_decorated == false)
   operator multi_ptr<value_type, Space, access::decorated::yes>() const;
 
+  // Available only when: (Space == address_space::global_space)
   void prefetch(size_t numElements) const;
```
AerialMantis, line 165: I think we could support `prefetch` for both `global_space` and `generic_space` specializations, to allow for a generic pointer which is points to global memory. I was speaking with @Naghasan about this and we concluded that it could be done safely by casting to a global pointer and delegating to the global built-in if it's a valid conversion, otherwise doing nothing.

Though we don't really have a precedent in the specification for whether functions that take a `multi_ptr` to `global_space` or `local_space`, should also allow for `generic_space` or whether users should be expected to do an explicit conversion, so maybe there is a discussion to be had there.

AerialMantis (approved):
Changes look good to me, just one point about whether we should also support `generic_space` for `prefetch`.

adoc/headers/multipointer.h
tomdeakin: Implementors please review.

gmlueck: I'm only trying to fix bugs in this PR.  If we want to extend this API to also support `generic_space`, should we do that as a separate PR (probably to SYCL-NEXT)?

Note that the table describing this API already lists this restriction:

> Available only when: `Space == access::address_space::global_space`.

fraggamuffin:
LGTM

keryell:
SYCL WG meeting 2023/06/15:

To be merged once the merge conflicts are solved.

PR #396, opened 2023-04-13
Editorial change with "any_device_has_v"

gmlueck:
We had feedback from a reviewer that our description of `any_device_has` and `all_devices_have` is confusing because we sometimes refer to the shortcuts with the "_v" suffix (`any_device_has_v` and `all_devices_have_v`).  Avoid this confusion by consistently using the trait names without the "_v".

The description of the example continues to use the "_v" names because the example uses those shortcuts, so it seemed appropriate to explain the code that is actually used in the example.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2867,14 +2867,14 @@ function is never instantiated with support for the [code]#sycl::half# type.
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks.

TApplencourt (approved):

fraggamuffin:
LGTM

PR #402, opened 2023-04-24
[SYCL] Modify signature of `async_work_group_copy` in `nd_item` and `group` 

mmoadeli:
Currently the source argument in `async_work_group_copy` of `nd_item` and `group` shares the same type with destination for their corresponding `multi_ptr` types.  Ideally The source should be declared as a `multi_ptr<const>` type.
Apart from above reasoning, I faced an issue on an attempt to update `accessor::get_pointer/accessor::get_multi_ptr`  to conform to the specifications in [PR-8874](https://github.com/intel/llvm/pull/8874).
An `access::mode::read` specialised accessor should return a  `multi_ptr<const DataT>` on call to `get_multi_ptr`. If the created `multi_ptr` is needed for a call to `async_work_group_copy` it requires an undesired conversion of `multi_ptr<const DataT>` to `multi_ptr<DataT>`.

This PR defines source of `async_work_group_copy` as a `multi_ptr<const DataT>` to address both conceptual and technical issues mentioned.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11881,7 +11881,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(decorated_local_ptr<DataT> dest,
-                                   decorated_global_ptr<DataT> src,
+                                   decorated_global_ptr<const DataT> src,
                                    size_t numElements) const
 ----
    a@ Permitted types for [code]#DataT# are all scalar and vector types.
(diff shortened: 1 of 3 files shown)


mmoadeli:
> I think this looks good. However, please also change the synopses in "adoc/headers/group.h" and "adoc/headers/nditem.h".

Thanks @gmlueck , done.

keryell (approved):
Good point!

tomdeakin:
WG: ready to merge

PR #418, opened 2023-05-26
Add missing noexcept to sampled_image member functions

steffenlarsen:
The `byte_size` and `size` member functions in `sampled_image` are `noexcept` in the description but are not in the class declaration. This commit adds the missing `noexcept` to the class declaration, which seems to be the intention given other classes with these member functions have the same.


gmlueck (approved):

TApplencourt (approved):

keryell (approved):
Good catch.

fraggamuffin:
LGTM

Issue #164, opened 2021-07-13
Clarify use of "sycl::error_category_for"


gmlueck:
This question was raised by a member of our QA team.  I'm raising it here, so they can participate if they want.

SYCL 2020 errors are generally reported by throwing `sycl::exception`, which contains an `std::error_code` identifying the error.  An `std::error_code` contains two things: a numeric `value` and an `std::category`.  The idea is that the category is used to identify the "domain" of the value.  For example, error code value `1` in category *A* and error code value `1` in category *B* could identify completely different errors.

SYCL 2020 also defines a free function `sycl::sycl_category()` which returns a reference to the singleton error category object that is used for SYCL errors.  Although the spec doesn't state this, one might assume that all `sycl::exception` objects thrown by the SYCL runtime will have `sycl::sycl_category()` as their error category.

However, the SYCL 2020 spec also defines a templated free function `sycl::error_category_for()`:

```
template<backend b>
const std::error_category& error_category_for() noexcept;
```

This function is not well defined in the spec, but it appears to return a reference to an error category object that is used for a specific backend.

It now becomes unclear what error category is used by `sycl::exception` objects that are thrown by the SYCL runtime.  Will they always be `sycl::sycl_category()`?  Or, will they sometimes be a backend-specific error category?  If so, how does the application know which category to expect?

To add to the confusion, the SYCL 2020 spec contains two code examples that illustrate different ways to handle SYCL runtime errors.  One example seems to assume that all `sycl::exception` objects have error category `sycl::sycl_category()`:

```
void catch_invalid_errors(sycl::context const& ctx) {
  try {
    do_something_to_invoke_error(ctx);
  }
  catch(sycl::exception const& e) {
    if(e.code() == sycl::errc::invalid) {
      std::cerr << "Invalid error: " << e.what();
    }
    else {
      throw;
    }
  }
}
```

However, the other example seems to indicate that the application must first test the error category:

```
void catch_backend_errors(sycl::context const& ctx) {
  try {
    do_something_to_invoke_error(ctx);
  }
  catch(sycl::exception const& e) {
    if(e.category() == sycl::error_category_for<sycl::backend::opencl>()) {
      switch(e.code().value()) {
        case CL_INVALID_PROGRAM:
          std::cerr << "OpenCL invalid program error: " << e.what();
        /* ...*/
      }
      else {
        throw;
      }
    }
    else {
      if(e.code() == sycl::errc::invalid) {
[...]

PR #420, opened 2023-06-01
Introduce CUDA backend specification (revised)

AerialMantis:
This pull request is a revised version of the original pull request for introducing the CUDA backend (https://github.com/KhronosGroup/SYCL-Docs/pull/197). I created a new PR as the original PR was becoming very difficult to continue rebasing.

Introduces the CUDA backend specification covering a mapping of the platform, execution and memories models and interoperability with the CUDA API for SYCL implementations targeting a CUDA backend.

The backend specification has been added as an appendix to the SYCL 2020 specification.

A pull request for introducing a test plan to the SYCL CTS for the CUDA backend interoperability defined in this backend specification is here - https://github.com/KhronosGroup/SYCL-CTS/pull/207. This will still need to be updated with recent changes.

### DIFF
--- a/adoc/chapters/cuda_backend.adoc
+++ b/adoc/chapters/cuda_backend.adoc
@@ -1,0 +1,551 @@
+// %%%%%%%%%%%%%%%%%%%%%%%%%%%% begin cuda_backend %%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+[appendix]
+[[chapter:cuda-backend]]
+= CUDA backend specification
+
+This chapter describes the behavior of the [code]#sycl::backend::cuda# backend
+and how it relates to the SYCL general programming model.
+This backend is implemented on top of the CUDA SDK and exposes NVIDIA
+devices to the SYCL general programming model.
+This chapter also describes how the SYCL generic interoperability interface is
+implemented for the [code]#sycl::backend::cuda# backend.
+
+The CUDA backend is enabled using the [code]#sycl::backend::cuda# value of [code]#enum
+class backend#. That means that when the CUDA backend is active the
+preprocessor macro [code]#SYCL_BACKEND_CUDA# will be defined.
+
+The CUDA backend requires an installation of CUDA SDK as well as one or more
+CUDA devices available in the system.
+[[sec:cuda:introduction]]
+== Introduction
+
+[[sec:cuda:mapping_of_sycl_programming_model]]
+== Mapping of SYCL programming model
+
+This section gives a general overview of how the SYCL programming model maps to
+CUDA. These two programming models are pretty similar in essence however they do
+have a few differences in terminology and architecture.
+
+[[sub:cuda:platform_model]]
+=== Platform Model
+
+A SYCL device maps to a single CUDA device.  As CUDA does not split into
+separate platforms there is no 'platform' concept in CUDA corresponding to the
+SYCL platform. Instead, a SYCL platform maps to a collection of CUDA devices
+represented by [code]#std::vector<CUdevice>#.
+
+A SYCL <<context>> simply maps to one, or multiple CUDA contexts. Indeed while
+a CUDA context is tied to a single device, this is not the case for a SYCL
+<<context>> and the CUDA backend implementation may use multiple CUDA contexts
+to emulate a SYCL <<context>> containing multiple devices.
+
+In CUDA, contexts and devices may need to be set as active on a thread for the
+CUDA Driver or CUDA Runtime APIs to use. Therefore the SYCL API entry points
+may change the active context or device whenever required for proper execution
+of SYCL operations, and no guarantee is provided that any prior active context
+or active device will be restored by the SYCL API entry points before returning
+to the user application.
+
+A SYCL <<queue>> simply maps to one or multiple CUDA streams. Indeed while a
+CUDA stream is in-order, a SYCL <<queue>> isn't, so a CUDA backend implementation
+may use multiple CUDA streams to implement an out of order SYCL <<queue>>.
+
+[[sub:cuda:memory_model]]
+=== Memory model
+
+==== Accessing memory on a different GPU
+
+Devices belonging to the same context must be able to access (directly or indirectly) each other's global memory. This is done in one of the following ways:
+
+- Device directly accesses memory on another device (peer-to-peer memory access).
+- CUDA-runtime-managed memory is used. CUDA runtime copies the data from one device to another.
+- Peer copy (one of [code]#CuMemcpyPeerAsync# or [code]#CuMemcpy3DPeerAsync#) is used to directly copy the data from one device to another.
+- The data is copied from one device to the host, and then from host to another device.
+
+==== Shared USM memory advices
+
+Values for the [code]#advice# parameter of [code]#sycl::queue::mem_advise# and [code]#sycl::handler::mem_advise# and their mapping to CUDA equivalent are defined in table <<table.cuda.memmodel.advices>>.
+
+[[table.cuda.memmodel.advices]]
+.Valid shared USM advices and their equivalents in CUDA
+[width="100%",options="header",cols="40%,30%,30%"]
+|====
+| SYCL shared USM advice | CUDA managed memory advice | processor, the advice is set for
+| [code]#sycl::cuda::advice::cuda_mem_advise_set_read_mostly# | [code]#cudaMemAdviseSetReadMostly# | device associated with the queue/handler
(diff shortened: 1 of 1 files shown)


AerialMantis:
Since https://github.com/KhronosGroup/SYCL-Docs/pull/197 I've made some additional changes based on feedback:
* Added a `get_error_code` free function for retrieving native CUDA error codes.
* Removed the optional features table as it's not necessary.
* Clarified that the CUDA backend can use either CUDA Driver API or the CUDA Runtime API.
* Clarified the context and queue mappings.
* Improved wording on synchronization of streams for interop.
* Specified that a queue created from a stream should be in-order.
* Added host task interop section.
* Removed image interop.
* Added sections describing synchronization and setting the current context.
* Added note about using the primary context.
* Clarified that it's defined as a KHR extension.
* Cleaned up some duplicated wording and fixed merge conflicts.

adoc/chapters/cuda_backend.adoc
```diff
@@ -0,0 +1,551 @@
...
+[[sec::programming_interface]]
+== Programming Interface
+
+[[table.cuda.get_error_code]]
+.Free function for querying whether an event has a native CUevent
+[width="100%",options="header",cols="40%,60%"]
+|====
+| Function | Description
+| [code]#bool cuda::has_native_event()# | Returns true if the SYCL [code]#event# has a native [code]#CUevent# associated with it.
+|====
```
gmlueck, line 335: How does this relate to `sycl::error_category_for` (see #164)?  If the idea is that each backend interop spec will provide a `get_error_code` function like this, then can we remove the ill-defined `sycl::error_category_for` from the core spec?

It would be very good to resolve #164 because we deferred any CTS testing of `sycl::error_category_for` due to its unclear specification.

keryell:
SYCL WG meeting 2023/06/15:

No recent progress, need to come back in a couple of weeks

adoc/chapters/cuda_backend.adoc
AerialMantis: Sorry I missed your comment here. Are you referring to `cuda::get_error_code(exception &)` and how that relates to `sycl::error_category_for`?

Yeah, I think the consensus was that we should move towards per-backend `get_error_code` functions, so then we can remove `sycl::error_category_for`. Though I'm not sure if it was decided exactly what we should do with `sycl::error_category_for`, whether it's removed immediately from SYCL 2020 or deprecated, I think there is justification for removing immediately it as it's been found to not work as it was intended.

tomdeakin:
Discussed in Oct/5/23 - @AerialMantis to say when to discuss again.

adoc/chapters/cuda_backend.adoc
```diff
@@ -0,0 +1,551 @@
...
+level grouping of threads, not including the whole kernel. Both 
+nd_range and grid can segment the groups of threads into one, two, or 
+three dimensions. SYCL sub-groups roughly map to
+cooperative groups [code]#thread_block_tile# as it allows for the
+work-group/thread block to be further subdivided into concurrent threads.
+Likewise, thread blocks map directly to work-groups, and a
+single thread is a SYCL work-item.
+
+CUDA primarily synchronizes the threads through two functions,
+[code]#cudaStreamSynchronize()# and [code]#\__syncthreads()#.
```
hdelan, line 277: I'm pretty sure all of the mentions of `cudaStreamSynchronize` in this paragraph should be the more general `cudaDeviceSynchronize`

adoc/chapters/cuda_backend.adoc
```diff
@@ -0,0 +1,551 @@
...
+| [code]#queue# | [code]#CUstream# | [code]#CUstream# | A SYCL [code]#queue# created from a [code]#CUstream# will encapsulate that stream and will have the [code]#property::queue::in_order# property . A single [code]#CUstream# is retrieved from a SYCL[code]#queue#, and before it is returned all native [code]#CUstream#s associated with the SYCL [code]#queue# synchronize with the calling thread.
+| [code]#event# | [code]#CUevent# | [code]#CUevent# | A SYCL [code]#event# created from a [code]#CUstream# will encapsulate that event. A single [code]#CUevent# is potentially retrieved from a SYCL [code]#event# if there is a valid native [code]#CUevent# associated with it, otherwise [code]#nullptr# is returned instead. The CUDA backend-specific free function [code]#cuda::has_native_event# can be used to query whether the SYCL [code]#event# has a valid native [code]#CUevent# associated with it.
```
hdelan, line 389:
```suggestion
| [code]#event# | [code]#CUevent# | [code]#CUevent# | A SYCL [code]#event# created from a [code]#CUevent# will encapsulate that event. A single [code]#CUevent# is potentially retrieved from a SYCL [code]#event# if there is a valid native [code]#CUevent# associated with it, otherwise [code]#nullptr# is returned instead. The CUDA backend-specific free function [code]#cuda::has_native_event# can be used to query whether the SYCL [code]#event# has a valid native [code]#CUevent# associated with it.
```

adoc/chapters/cuda_backend.adoc
```diff
@@ -0,0 +1,551 @@
...
+|====
+
+==== Synchronization
+
+When retireving a [code]#CUstream# from a SYCL [code]#queue# the SYCL runtime
```
hdelan, line 394:
```suggestion
When retrieving a [code]#CUstream# from a SYCL [code]#queue# the SYCL runtime
```

adoc/chapters/cuda_backend.adoc
```diff
@@ -0,0 +1,551 @@
...
+must synchronize all [code]#CUstream#s associated with the SYCL [code]#queue#
+in order to guarnatee consistent ordering of commands previously enqueued to
+the SYCL [code]#queue# in relation to any commands enqueued using the native
+[code]#CUstream#. The native [code]#stream# must be synchornized with and
```
hdelan, line 398:
```suggestion
[code]#CUstream#. The native [code]#stream# must be synchronized with and
```

hdelan:
Some things missing here:

- When we call `ih.get_native_queue();` are we always guaranteed to get the same `CUstream/cudaStream`? I think this should be a yes
- From within a HT function, is it necessary to set the context? This is a bit tricky because we are essentially saying: Is a HT function executed in a new thread? If the context needs to be set then how should this happen? If we call an `ih.get_native...` method should this set the `cuCtx`?
- How do we reconcile the CUDA types that are equivalent with SYCL types if different implementations may use CUDA driver types or CUDA runtime types? Since the CUDA driver can mix and match with runtime, maybe it is better to make SYCL types map to CUDA runtime types instead of driver types

t4c1:
Why the preference for the same cuda stream? I think for performance reasons it is better to leave this unspecified to allow different streams to be returned.

hdelan:
If I want `ih.get_native_queue` to give me different streams for any two separate HTs then I have no way of checking that this has indeed happened. It's better to always return the same stream for every call to `ih.get_native_queue`, or indeed to give all streams associated with the queue. However I think that the latter is still a bit confusing as the stream pool may grow or shrink according to the implementation's preferences (unless we specified that it doesn't, which is over prescriptive I think), so it would be tricky to reason about which stream is which in separate HT funcs.

I think if a secondary stream is needed for a queue then it can be created within a HT or the user can use in order SYCL queues as CUstreams, which seems to be what a lot of users do anyway.

hdelan:
The primary motivation is that something like AdaptiveCPP's `enqueue_custom_operation` can work (being able to submit async work in a HT-like lambda) only if we have guarantees that we will always get the same stream from `get_native_queue`. Similar behaviour might be desirable to have in the future in core SYCL. 

This is a small step towards being able to submit async work in the HT lambda. 

At the minimum I think we should define this for in order queues, whatever about out of order.

hdelan:
@t4c1 , @jle-quel  and I were talking about this offline. We think it's important to guarantee that for an in order queue, `ih.get_native_queue` is guaranteed to return the same queue every time, whereas in order queues have no such guarantees, even within the same host task func.

So for out of order queues:
```c++
q.host_task(...
   bool sameStream = ih.get_native_queue() == ih.get_native_queue();
});
```
Might be false, whereas for in order queues this is always true.

And 
```c++
q.host_task(...
   stream_acc[0] = ih.get_native_queue();
}).wait();
q.host_task(...
   bool sameStream = ih.get_native_queue() == stream_acc[0];
   // Using the stream stored in previous HT lambda through accessor
});
```
Will always be true for in order queues. This may or may not evaluate to true for out of order queues.

Maybe also worth mentioning that if dispatching to the same CUstream across different HTs it is important to make sure the correct command dependencies are used, so async work is dispatched to CUstreams in the correct order.

AerialMantis:
This PR has been open for a while, the conclusion of the most recent discussions was to go in a different direction, and this is now awaiting a new proposal. So we are closing this.

PR #425, opened 2023-06-07
Type for copy and fill should be device-copyable

gmlueck:
Clarify that the type used in the `handler::copy` and `handler::fill` functions must be "device copyable".  This avoids confusion when `T` is a class type because people might expect the class's copy constructor to be called, when our intent is that the implementation should just do a byte copy.

There is no need to clarify the element type of the accessors in these functions becuse we already state that the element type of a buffer must be device copyable.  Therefore, this PR only changes the wording for the non-accessor parameters.


TApplencourt (approved):
We don't have any `prefetch` or `mem_advise` that are templated with `T` so LGTM!

fraggamuffin:
LGTM

PR #427, opened 2023-06-08
Relax exception requirement for placeholder capture

steffenlarsen:
Currently the specification mandates that an implementation throws a synchronous exception when a command is submitted that has captured an accessor that has not previously been through `handler::require`. However, it may be difficult for `host_task` and general commands in library-only implementations to guarantee. To alleviate these problems, this commit aims to relax these exception requirements make the synchronous exception optional and require select accessor member functions to trigger asynchronous exceptions if the corresponding command is executed.

Additionally, this commit adds notes to select member functions in the deprecated accessors specializations for `target::local` and `target::constant` specifying that they can only be used in commands. This corresponds to the same interfaces for the non-deprecated accessor classes.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6554,10 +6554,16 @@ yet bound to a <<command-group>>.  Before such an accessor can be used in a
 <<command>>, it must be bound by calling [code]#handler::require()#.  If a
 placeholder accessor is passed as an argument to a <<command>> without first
 being bound to a <<command-group>> with [code]#handler::require()#, the
-implementation throws a synchronous [code]#exception# with the
-[code]#errc::kernel_argument# error code when the <<command>> is submitted.
+implementation may throw a synchronous [code]#exception# with the
+[code]#errc::kernel_argument# error code when the <<command>> is submitted. If
+the implementation does not throw an synchronous [code]#exception# upon
+<<command>> submission, an asynchronous [code]#exception# will be thrown during
+the execution of the <<command>> if it calls any of the member functions stating
+so (see <<sec:accessors.command.buffer.interface>> and
+<<sec:accessor.common.members>>.)
 
 
+[[sec:accessors.command.buffer.interface]]
 ===== Interface for buffer command accessors
 
 A synopsis of the [code]#accessor# class is provided below, showing the
@@ -6918,6 +6924,11 @@ accessor_ptr<IsDecorated> get_multi_ptr() const noexcept
 
 This function may only be called from within a <<command>>.
 
+A <<command>> calling this function will throw an asynchronous
+[code]#exception# with [code]#exception#errc::kernel_argument# if this is a
+placeholder accessor that has not been bound to a <<command>> by calling
+[code]#handler::require()#.
+
 a@
 [source]
 ----
@@ -6931,6 +6942,11 @@ Assignment to the single element that is accessed by this accessor.
 
 This function may only be called from within a <<command>>.
 
+A <<command>> calling this function will throw an asynchronous
+[code]#exception# with [code]#exception#errc::kernel_argument# if this is a
+placeholder accessor that has not been bound to a <<command>> by calling
+[code]#handler::require()#.
+
 a@
 [source]
 ----
@@ -6943,6 +6959,11 @@ const accessor& operator=(value_type&& other) const
 Assignment to the single element that is accessed by this accessor.
 
 This function may only be called from within a <<command>>.
+
+A <<command>> calling this function will throw an asynchronous
+[code]#exception# with [code]#exception#errc::kernel_argument# if this is a
+placeholder accessor that has not been bound to a <<command>> by calling
+[code]#handler::require()#.
 
 |====
 
@@ -7078,6 +7099,7 @@ size_t get_count() const
 |====
 
 
+[[sec:accessor.command.buffer.constant.specialization]]
 ====== Accessor specialization with [code]#target::constant_buffer#
 
 The [code]#accessor# class may be specialized with target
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -6554,10 +6554,16 @@ yet bound to a <<command-group>>.  Before such an accessor can be used in a
...
 being bound to a <<command-group>> with [code]#handler::require()#, the
-implementation throws a synchronous [code]#exception# with the
-[code]#errc::kernel_argument# error code when the <<command>> is submitted.
+implementation may throw a synchronous [code]#exception# with the
+[code]#errc::kernel_argument# error code when the <<command>> is submitted. If
+the implementation does not throw an synchronous [code]#exception# upon
+<<command>> submission, an asynchronous [code]#exception# will be thrown during
+the execution of the <<command>> if it calls any of the member functions stating
+so (see <<sec:accessors.command.buffer.interface>> and
+<<sec:accessor.common.members>>.)
```
illuhad, line 6563: How do you imagine this be implemented in the library-only case? I'm concerned that then any of these functions need to have some runtime check for every access, which could harm performance.

steffenlarsen: I believe you're right, this would likely require some sort of runtime check for `is_placeholder() && IsThisAccessorBoundInThisCommand()`. Though we know  `IsThisAccessorBoundInThisCommand()` would be the same in each call to it inside the given command, I struggle to see how the compiler would be able to discern that. As such, I fear you're right that the implementation would effectively have to check with every call.

Do you have alternatives in mind? We could just drop the idea of throwing asynchronously for this, leaving putting the burden onto the user for cases where we can't (or don't want the overhead of) checking for unbound accessors. If we do that, we could add an interface to check if an accessor is bound, but to that end the user might as well just make sure to call require with all their accessors if they've gotten that far anyway.
Implementations can of course offer a mode that adds an asynchronous exception like what is proposed here if the user enables it. Defining such a mode seems outside the spec's responsibilities though.

illuhad: I don't really know how to handle that in a better way either. The easiest fix would likely be to adopt a "*may* throw an asynchronous exception" wording here. Generally, accessors are constructed such that once they are inside a kernel/host task little checking needs to happen for performance reasons. So I think it shouldn't be surprising to users (and maybe well-aligned with other accessor uses) that such checks may not be in place.

steffenlarsen: After all, the exception is a service to inform the user that they're doing something they shouldn't. As such, I am okay with letting it be entirely optional for an implementation to gauge whether it is worth doing. @gmlueck - Do you feel strongly about that solution?

TApplencourt: Without exception thrown, how can one debug this kind of bug (aka using a placeholder without behind bound)?

keryell: It would be possible to switch to another implementation throwing an exception during debug time, or using a pure CPU implementation with instrumentation of `operator[]` and `get_pointer()`, or use a pure CPU implementation compiled with UBsan or equivalent mode.

adoc/chapters/programming_interface.adoc
```diff
@@ -6554,10 +6554,16 @@ yet bound to a <<command-group>>.  Before such an accessor can be used in a
 <<command>>, it must be bound by calling [code]#handler::require()#.  If a
 placeholder accessor is passed as an argument to a <<command>> without first
 being bound to a <<command-group>> with [code]#handler::require()#, the
-implementation throws a synchronous [code]#exception# with the
-[code]#errc::kernel_argument# error code when the <<command>> is submitted.
+implementation may throw a synchronous [code]#exception# with the
+[code]#errc::kernel_argument# error code when the <<command>> is submitted. If
+the implementation does not throw an synchronous [code]#exception# upon
+<<command>> submission, an asynchronous [code]#exception# will be thrown during
```
keryell, line 6560:
```suggestion
<<command>> submission, an asynchronous [code]#exception# may be thrown during
```
Perhaps we could also add a rationale in a non normative note to avoid too much puzzling.

adoc/chapters/programming_interface.adoc
gmlueck: If we think the error check will cause too much overhead, then we can remove the requirement.  The wording would then be much simpler:

> If a placeholder accessor is passed as an argument to a \<\<command\>\> without first
being bound to a \<\<command-group\>\> with [code]#handler::require()#, the **behavior is undefined**.

We would also remove all the proposed wording below about throwing an async exception in this case.

I think it would be good to re-examine error checking requirements in SYCL-Next.  We need some overall philosophy about error checking.  Currently, the spec is very inconsistent here.  In some cases, we require an error check and in others we just say the behavior is undefined.  I don't think we have a rationale for choosing one vs. the other.

> Without exception throwed, how can one debug this kind of bug (aka using a placeholder without behind bound)?

Even if the spec does not require an exception in this case, implementations are still allowed to do this.  I think DPC++ will continue to throw an exception in the case of kernels that use an unbound placeholder. Library only implementations could provide this ability under an compile-time option.

adoc/chapters/programming_interface.adoc
```diff
@@ -7294,6 +7321,13 @@ constant_ptr<DataT> get_pointer() const noexcept
       at the beginning of the buffer.  The return value is unspecified if the
       accessor is empty.
 
+This function may only be called from within a <<command>>.
```
gmlueck, line 7324: It would be nice to keep these additions in this PR, as they fix another bug in the spec.

adoc/chapters/programming_interface.adoc
TApplencourt: Agree with greg that a coherent philosophy is required.

Also Maybe for SYCL-next we can add a new queue optional parameter, `require-throwing` or something like that?  It's always better to give control to users about the overhead /behavior they want (see discussion about returning an event) ? This can also be interesting for the SC user.

keryell:
SYCL WG meeting 2023/06/15:

Relax exception requirement for placeholder capture

Sometimes this should throw, sometimes not...

Proposition to make it UB and add normative note to suggest good implementation behavior like throwing when possible.

adoc/chapters/programming_interface.adoc
```diff
@@ -7669,6 +7685,9 @@ When [code]#AccessTarget# is [code]#target::local#, available only when
 
 Returns a reference to the single element that is accessed by this accessor.
 
+When [code]#AccessTarget# is [code]#target::local# or [code]#target::constant#,
```
gmlueck, line 7688:
```suggestion
When [code]#AccessTarget# is [code]#target::local# or [code]#target::constant_buffer#,
```

The target name is `constant_buffer` not `constant`.  Same for the other statements below.

steffenlarsen: Good eye! Fixed.

gmlueck (approved):

TApplencourt (approved):

keryell (approved):
Thanks.

AerialMantis (approved):

fraggamuffin:
LGTM

PR #428, opened 2023-06-09
Clarify "gentype" function signatures

gmlueck:
This commit clarifies the expected signatures of the functions defined
in section 4.17.5 "Math functions" - 4.17.9 "Relational functions".

* Previously, it was not clear how much templating (if any) is present
  in these functions.  We now clarify that these functions are mostly
  exposed as overloads, using template parameters only for the `vec`
  and `marray` extents and for the `multi_ptr` parameters.

* Previously, it was not clear how the "generic type names" should be
  interpreted when a function signature uses several generic type
  names like this:

  ```
  genfloat fmax(genfloat x, sgenfloat y)
  ```

  Did we intend `fmax` to have overloads for all possible combinations
  of types from `genfloat` and `sgenfloat`, or are only certain
  combinations expected?  Cases like this have been clarified by
  rewriting the spec to avoid multiple generic type names in the same
  signature.  For example:

  ```
  genfloatf fmax(genfloatf x, float y)
  genfloatd fmax(genfloatd x, double y)
  genfloath fmax(genfloath x, half y)
  ```

  To resolve cases that weren't obvious, I aligned with the OpenCL
  behavior, which makes sense because all of these functions were
  originally derived from OpenCL.

* There are still two cases when a function could be defined with
  two different generic type names.  These are resolved by adding
  the names `unsignedtype` and `elementtype` and specifying exactly
  how these generic types interact with the other generic types.

* The `upsample`, `mad24`, and `mul24` functions presented an
  interesting problem.  These were defined in terms of the
  `{i,u}genintegerNbit` generic types, which lead to ambiguities in
  cases like this:

  ```
  igeninteger32bit upsample(igeninteger16bit hi, ugeninteger16bit lo)
  ```

  The `igeninteger32bit` type was defined to be all signed integer
  types that are 32 bits wide.  It was therefore unclear whether
  this function should return `int` or `long` for an implementation
  where both types are 32-bits.  This was resolved by defining these
  three functions exclusively in terms of the fixed-width integer
  types.  In the example above, the return type is now `int32_t`
  (or a `vec` or `marray` of `int32_t`).  This makes sense because all
  three of these functions assume input values which have a specific
  number of bits.

* The overloads involving `marray` of integer types were also
  simplified.  To give an example, we previously defined overloads for
  both `mshort{n}` and `marray<short,{N}>`.  This may seem purely
  redundant, but `mshort{n}` is defined as `marray<int16_t,{N}>`.  If
  we assume that each of the fixed-width integer types aliases to one
  of the fundamental integer types, there is no need to define
  overloads for both.  We therefore define overloads only for the
  fundamental integer types.  This will only make a difference for a
  bizarre implementation where a fixed-width integer type is a distinct
  type, different from all the fundamental types.  We do not think that
  SYCL needs to define special overloads for this case, which is
  consistent with the C++ standard (which also does not define special
  library function overloads for the fixed-width integer types).

  Note that all the overloads involving `vec` of integer types are
  defined exclusively for the fixed-width integer types (and not the
  fundamental types).  This was the case even before this commit.  This
  makes sense if we think the main use for `vec` is for compatibility
  with OpenCL, where the sizes of the integer types are all fixed.

* The structure of the table of "Generic type names" was also improved
  for clarity.  Previously, there was a deep hierarchy of generic type
  names, where many generic types were defined in terms of other
  generic types.  A reader needed to traverse down through the
  hierarchy in order to understand exactly which concrete types were
  included in each generic type.  The table no longer has any
  hierarchy, so each generic type now lists the full set of concrete
  types that it represents, which makes it much easier to read.

* The following statement was removed from section 4.17.1 "Description
  of generic type names" as part of my rewrite of the introduction of
  that section:

  > All of the OpenCL built-in types are available in the namespace
  > `sycl`.

  I think this statement is erroneous because we did not intend to
  export all of these types into the `sycl` namespace.  In addition,
  this would be a very strange place for us to say that.

Closes internal issue 278.
Closes internal issue 590.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19401,26 +19401,109 @@ that all of the built-in functions fulfill the same requirements for both
 host and device.
 
 
-=== Description of the built-in types available for SYCL host and device
+=== Description of generic type names
 
-All of the OpenCL built-in types are available in the namespace
-[code]#sycl#. For the purposes of this document we use
-generic type names for describing sets of valid SYCL types. The
-generic type names themselves are not valid SYCL types, but they
-represent a set of valid types, as defined in
-<<table.gentypes>>. Each generic type within a section is
-comprised of a combination of scalar, SYCL [code]#vec# and/or [code]#marray# class
-specializations. The letters [code]#{n}# and [code]#{N}# define valid sizes for
-class specializations, where [code]#{n}# means 2,3,4,8,16 and [code]#{N}# means
-any positive value of size_t type. Note that any reference to the base type
-refers to the type of a scalar or the element type of a SYCL [code]#vec# or
-[code]#marray# specialization.
+The functions defined in sections <<sec:math-functions>> through
+<<sec::relational-functions>> are specified using a set of generic type names.
+These are not actual type names available to SYCL applications.  Instead, they
+are a shorthand used in this document to describe the set of overloads for
+these functions.  For example the [code]#fmax# function is specified as:
 
+[source]
+----
+genfloat fmax(genfloat x, genfloat y)
+----
 
+The generic type [code]#genfloat# is a stand-in for a set of a actual type
+names: [code]#float#, [code]#double#, etc.  To construct the actual overloads
+that are available, replace the generic type name with the name of each of
+these actual types:
+
+[source]
+----
+float fmax(float x, float y)
+double fmax(double x, double y)
+template<int N> vec<float,N> fmax(vec<float,N> x, vec<float,N> y)
+template<size_t M> marray<float,M> fmax(marray<float,M> x, marray<float,M> y)
+// etc.
+----
+
+When determining the set of overloads that are available, use the following
+rules:
+
+* If the specification of a function has several occurrences of the same
+  generic type name, replace each occurrence with the same actual type.  For
+  example, the [code]#fmax# example above has three occurrences of
+  [code]#genfloat#: one for the return type and one for each parameter type.
+  The available overloads include only the cases where both parameters have
+  the same type, and the return type always matches the type of the parameters.
+
+* Some of the generic type names can be replaced with a templated type, where
+  the identifiers [code]#N#, [code]#M#, [code]#Space#, and [code]#IsDecorated#
+  represent template parameters.  When this happens, the overload is a
+  templated function, as shown above with the [code]#fmax# example.  Only a few
+  combinations of template parameters are possible, and the following list
+  tells the order and type of the template parameters for each possible
+  combination:
+
+** If there is a single template parameter [code]#N# (number of dimensions of
+   the [code]#vec# type):
++
+[source]
+----
+template<int N>
+----
+
+** If there is a single template parameter [code]#M# (number of dimensions of
+   the [code]#marray# type):
++
+[source]
+----
+template<size_t M>
+----
+
+** If there are two template parameters [code]#Space# and [code]#IsDecorated#
+   (address space and decoration for the [code]#multi_ptr# type):
++
+[source]
+----
+template<access::address_space Space, access::decorated IsDecorated>
+----
+
+** If there are three template parameters [code]#N#, [code]#Space# and
+   [code]#IsDecorated#:
++
+[source]
+----
+template<int N, access::address_space Space, access::decorated IsDecorated>
+----
+
+** If there are three template parameters [code]#M#, [code]#Space# and
+   [code]#IsDecorated#:
++
+[source]
+----
+template<size_t M, access::address_space Space, access::decorated IsDecorated>
+----
+
+* Unless specified otherwise, the template parameter [code]#N# is constrained
+  to the values 2, 3, 4, 8, or 16.
+
+* Unless specified otherwise, the template parameter [code]#M# is constrained
+  to a positive value.
+
+* Unless specified otherwise, the template parameter [code]#Space# is
+  constrained to the values [code]#access::address_space::global_space#,
+  [code]#access::address_space::local_space#, or
+  [code]#access::address_space::private_space#.
+
+* Unless specified otherwise, the template parameter [code]#IsDecorated# is
+  constrained to the values [code]#access::decorated::yes# or
+  [code]#access::decorated::no#.
 
 [[table.gentypes]]
-.Generic type name description, which serves as a description for all valid types of <<opencl12, parameters to kernel functions>>.
-[width="100%",options="header",separator="@",cols="65%,35%"]
+.Generic type name descriptions.
+[width="100%",options="header",separator="@",cols="25%,75%"]
 |====
 @ Generic type name  @ Description
 a@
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,697 @@ that all of the built-in functions fulfill the same requirements for both
...
+[source]
+----
+template<size_t M, access::address_space Space, access::decorated IsDecorated>
+----
+
+* Unless specified otherwise, the template parameter [code]#N# is constrained
+  to the values 2, 3, 4, 8, or 16.
+
+* Unless specified otherwise, the template parameter [code]#M# is constrained
+  to a positive value.
```
Pennycook, line 19493: Is this required?  If `M` is `size_t`, isn't it always unsigned?

gmlueck: This prevents the value from being zero.

Thinking about this more, though ... maybe we don't need either of these constraints on `N` or `M`.  Presumably, the types `vec` and `marray` should have some sort of `static_assert` that their template parameter is in range.  In that case, there is no need to redundantly put the same constraint on these functions.

Pennycook: Oh, I see.  If you do keep these constraints, I think "M must be greater than zero" is a clearer way to say that.

But I'm in favor of removing the duplicated constraints.

adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,697 @@ that all of the built-in functions fulfill the same requirements for both
...
+
+* Unless specified otherwise, the template parameter [code]#Space# is
+  constrained to the values [code]#access::address_space::global_space#,
+  [code]#access::address_space::local_space#, or
+  [code]#access::address_space::private_space#.
```
Pennycook, line 19498: Reading this again this morning... Why aren't these functions available for generic pointers?  Pointers are only ever used for return types, so it seems odd to limit the available address spaces here.

It would make sense to keep a note that `Space` cannot be `constant_space`, but we could remove that in SYCL-Next (since `constant_space` is deprecated).

gmlueck: I noticed this too, and I was thinking of creating an issue for this.  The spec had the same limitation before my PR, and I didn't want to expand the scope of this PR to include new additions.  In addition to adding `generic_space`, I think we should consider adding new overloads that take raw pointers.

adoc/chapters/programming_interface.adoc
gmlueck: I removed the constraints for `N` and `M` in 1133eefba7b155f68320b0776417ed9afa337040.

adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,694 @@ that all of the built-in functions fulfill the same requirements for both
...
+The generic type [code]#genfloat# is a stand-in for a set of a actual type
+names: [code]#float#, [code]#double#, etc.  To construct the actual overloads
+that are available, replace the generic type name with the name of each of
+these actual types:
+
+[source]
+----
+float fmax(float x, float y)
+double fmax(double x, double y)
+template<int N> vec<float,N> fmax(vec<float,N> x, vec<float,N> y)
```
steffenlarsen, line 19426: Is there a benefit to keep the `vec` variants as templates? Since the set of valid values for `{n}` is so limited we could just have overloads for each one.

gmlueck: I agree that this is possible.  I did it this way because I think this is what we agreed on when we discussed internal issue 278.

One advantage to implementing the `vec` versions as many overloads is that it allows implicit conversion from `__swizzled_vec__`.  For example:

```'
vec<float, 3> v{1.0f, 2.0f, 3.0f};
fmax(v.swizzle<2, 0, 1>(), v.swizzle<0, 1, 2>());
```

The calls to `v.swizzle` above return the unnamed type `__swizzled_vec__` which is convertible to `vec<float, 3>`.  If we implement `fmax` as a template, I believe the code above won't compile because template resolution doesn't consider conversions.

If we implement `fmax` as separate overloads for all of the `{n}` possibilities, then I think the code above does compile because overload resolution does consider conversions.

There is no similar issue with `marray` because there is no type like `__swizzled_vec__` that is convertible to an `marray`.  In any event, we cannot do multiple overloads for `marray` because there aren't a fixed number of `{N}` values.

@illuhad how does hipSYCL deal with the `vec` variants of these math functions?  Is the `{n}` a template parameter or are there many different overloads for all the `{n}` possibilities?

gmlueck: It seems like this would be visible to applications, though, so this could cause portability problems.  For example, if the spec says the synopsis is:

```
template<int N> fmax(vec<N> x, vec<N> y)
```

Then, application code might do something like this:

```
template<int N>
void foo() {
  sycl::vec<N> x;
  sycl::vec<N> y;
  fmax<N>(x, y);
}
```

This won't compile unless the implementation actually implements `fmax` with a template parameter of type `int`.

keryell: The spec should just say 1 line to rule them all:
```c++
template <typename X, typename Y> fmax(X x, Y y) -> return_t<X, Y>;
```
and just explain the conversion rule implemented under the hood and when it is valid, in a modern intensive generic way instead of in an old extensive overloaded way,
Having a special template for `int N` like
```c++
template<int N> vec<float,N> fmax(vec<float,N> x, vec<float,N> y)
```
looks like syntactic noise because it allows to write the cryptic
```c++
fmax<3>(1.f, 42.f) // returns a vfloat3 obviously o_o
```
instead of just
```c++
fmax(vfloat3 { 1.f }, 42.f) // returns a vfloat3 more naturally
``````

Comment (gmlueck) [reply adoc/chapters/programming_interface.adoc]: > Is there a benefit to keep the `vec` variants as templates? Since the set of valid values for `{n}` is so limited we could just have overloads for each one.

I did this in a8bcf6296d683a377dc613a7c5bb53a2e6c3b215.

File: adoc/chapters/programming_interface.adoc [hunk 4]
```diff
@@ -19423,8 +19423,13 @@ these actual types:
...
 double fmax(double x, double y)
-template<int N> vec<float,N> fmax(vec<float,N> x, vec<float,N> y)
-template<size_t M> marray<float,M> fmax(marray<float,M> x, marray<float,M> y)
+vec<float,1> fmax(vec<float,1> x, vec<float,1> y)
+vec<float,2> fmax(vec<float,2> x, vec<float,2> y)
+vec<float,3> fmax(vec<float,3> x, vec<float,3> y)
+vec<float,4> fmax(vec<float,4> x, vec<float,4> y)
+vec<float,8> fmax(vec<float,8> x, vec<float,8> y)
+vec<float,16> fmax(vec<float,16> x, vec<float,16> y)
+template<size_t N> marray<float,N> fmax(marray<float,N> x, marray<float,N> y)
```
keryell, line 19432: I guess it is a discrepancy compared to `vec` to solve with concepts in SYCL Next...

keryell (approved):
Ouch, this was verbose...

adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,1161 @@ that all of the built-in functions fulfill the same requirements for both
...
+// etc.
+----
+
+When determining the set of overloads that are available, use the following
+rules:
+
+* If the specification of a function has several occurrences of the same
+  generic type name, replace each occurrence with the same actual type.  For
```
AerialMantis, line 19440:
```suggestion
  generic type name, replace each occurrence in the same overloads with the same actual type.  For
```

adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,1161 @@ that all of the built-in functions fulfill the same requirements for both
...
+float fmax(float x, float y)
+double fmax(double x, double y)
+vec<float,1> fmax(vec<float,1> x, vec<float,1> y)
+vec<float,2> fmax(vec<float,2> x, vec<float,2> y)
+vec<float,3> fmax(vec<float,3> x, vec<float,3> y)
+vec<float,4> fmax(vec<float,4> x, vec<float,4> y)
+vec<float,8> fmax(vec<float,8> x, vec<float,8> y)
+vec<float,16> fmax(vec<float,16> x, vec<float,16> y)
+template<size_t N> marray<float,N> fmax(marray<float,N> x, marray<float,N> y)
+// etc.
```
AerialMantis, line 19433:
```suggestion
// ...
```

adoc/chapters/programming_interface.adoc
```diff
@@ -19401,481 +19401,1161 @@ that all of the built-in functions fulfill the same requirements for both
...
+----
+func(vec<float,1>, multi_ptr<vec<float,1>, Space, IsDecorated>)
+func(vec<float,2>, multi_ptr<vec<float,2>, Space, IsDecorated>)
+// ...
+----
++
+However, there are no overloads where the two parameters have different
+vector lengths.
+
+* Some of the generic type names can be replaced with a templated type, where
```
AerialMantis, line 19483:
```suggestion
* Some of the generic type names are replaced with a templated type, where
```
Should we be stronger here and say they **are** replaced with a template?

adoc/chapters/programming_interface.adoc
AerialMantis: I have addressed this in https://github.com/KhronosGroup/SYCL-Docs/pull/432 which is rebased on this PR.

AerialMantis (approved):
LGTM.

adoc/chapters/programming_interface.adoc
gmlueck: Added in bd1229a146abf422525dd25563865087e0d2fa1a

adoc/chapters/programming_interface.adoc
gmlueck: Added in bd1229a146abf422525dd25563865087e0d2fa1a

adoc/chapters/programming_interface.adoc
gmlueck: Added in bd1229a146abf422525dd25563865087e0d2fa1a

adoc/chapters/programming_interface.adoc
gmlueck: Our thinking is that `vec` will probably go away at some point.  Once that happens, all of these overloads on `vec` can be removed.  In the meantime, most people wanted to do whatever was easiest for implementors w.r.t. `vec`, and the consensus from implementors was that adding these overloads is the easiest solution.

keryell: The problem is nit the `vec` but that the `marray` with this template is not user-friendly without deeper rework.

PR #431, opened 2023-06-21
Remove errc_for / error_category_for

gmlueck:
Remove `errc_for` and `error_category_for`, which were underspecified.

I think `errc_for` was intended to be the type of an error code enumeration that each backends provides.  In fact, the spec used to say that each backend was required to define such an enumeration. However, the OpenCL backend interop spec does not do this, and we do not know of any backend that does.

I think `error_category_for` was supposed to return the static error category object for each backend, although the spec did not provide any explanation for this function.  Again, the OpenCL backend interop spec does not document any such object, and we do not know of any backend that does this.

Remove an example that used `error_category_for` for the OpenCL backend.  I do not think this example would work given the current state of the OpenCL backend interop spec.

The direction for backend specific errors seems to have moved in a different direction.  The OpenCL backend interop spec defines a free function `opencl::get_error_code`, which returns a backend specific error code from a SYCL `exception` object.  Other backends seem to be following this model.  With this model, there is no need for the core SYCL specification to mandate an error code enumeration or error category for each backend.

Addresses #164.


fraggamuffin:
wait another week, but all so far agree to removal as long as no one has implemented this

keryell (approved):
Thanks.

gmlueck:
Ping to the group: Looking for another reviewer here ...

TApplencourt (approved):

fraggamuffin:
LGTM

PR #432, opened 2023-06-22
Default multi_ptr to legacy to avoid code break with SYCL 1.2.1

AerialMantis:
We have discovered that some of the changes in SYCL 2020 for `multi_ptr` have introduced breaking changes for SYCL 1.2.1 applications moving to SYCL 2020. At the time of these changes it was thought to be necessary but since then the SYCL working group has adopted a philosophy of deprecating interfaces before removing them outright.

This pull request aims to address these changes, such that SYCL 1.2.1 code will still be compatible with SYCL 2020 implementations, avoiding errors in moving from one version to the other and in portability between implementations.

### 1. Default `access::decorated` value for `multi_ptr`

The first issue that was identified was that the SYCL 2020 wording introduces the `access::decorated` template parameter in `multi_ptr`, however, does not provide a default. This means that any code which declares the `multi_ptr` without this template argument, as was the case in SYCL 1.2.1 would no longer compiler.

For example:

```
multi_ptr<float, access::address_space::global_space> multiPtr;
```

To solve this issue, this PR has the `multi_ptr` class template default to `access::decorated::legacy`. It also specifies in the `access::decorated` enum class that `access::decorated::legacy` is deprecated.

### 2. `accessor` returning a raw pointer

The second issue that was identified was that the SYCL 2020 wording changes the return type of `accessor::get_pointer` from a `multi_ptr` to a raw pointer. This means that any code which expects the return value of `accessor::get_pointer` to be a `multi_ptr` will fail to compile.

For example:

```
multi_ptr<float, access::address_space::global_space> multiPtr = acc.get_pointer();
```

or

```
float f4;
f4.load(offset, acc.get_pointer());
```

To solve this issue, this PR reverts back to `accessor::get_pointer` returning a `multi_ptr` with `access::decorated::legacy`. It also introduces wording for `accessor::get_pointer` which was previously missing and specifies that it is deprecated.

### 3. Missing interfaces for `multi_ptr` with `access::decorated::legacy`

The third issue that was identified was that the SYCL 2020 wording introduces new interfaces for the `multi_ptr` class; specifically the `multi_ptr::get_raw` and `multi_ptr::get_decorated` member functions, but not for the `access::decorated::legacy` specialization. This means any new code which uses these member functions will not be compatible with old code which uses the `access::decorated::legacy` specialization of `multi_ptr` and will therefore have to multi-version for the two specializations.

For example:

```
template <typename T, access::address_space AddressSpace, access::decorated IsDecorated>
void foo(multi_ptr<T, AddressSpace,IsDecorated> multiPtr) {
  T *ptr = multiPtr.get_decorated();
  // do something with ptr
}
```

To solve this issue, this PR adds the `multi_ptr::get_raw` and `multi_ptr::get_decorated` member functions to the `access::decorated::legacy` specialization of `multi_ptr` as well.

### 4. `async_work_group_copy` requiring `access::decorated::yes`

The fourth issue that was identified was that the SYCL 2020 wording changes the `nd_item::async_work_group_copy` and `group::async_work_group_copy` member functions such that they now take a `multi_ptr` with `access::decorated::yes` explicitly. This means that any code which uses the `access::decorated::legacy` specialization will not be able to call these functions.

For example:

```
multi_ptr<float, access::address_space::global_space> srcPtr = srcAcc.get_pointer();
multi_ptr<float, access::address_space::global_space> destPtr= destAcc.get_pointer();
async_work_group_copy(destPtr, srcPtr, numElements);
```
To solve this issue, this PR re-introduces the original SYCL 1.2.1 overloads of ` async_work_group_copy` which take a `multi_ptr` with `access::decorated::legacy`, and specifies them as deprecated.

Note, the examples for these are based on the previous changes being applied, however, those issues are all reproducible even if they weren't using other SYCL 2020 interfaces.

### Edits

* https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/b84963dbb7a353890317fcaa01c8c172a727ae68: Have `local_accessor::get_pointer() also return ` multi_ptr` for consistency and address comments.
* Rebased onto https://github.com/gmlueck/SYCL-Docs/tree/gmlueck/gentype-funcs.
* https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/a101c3b6618c9d28dba474fe5c9927934c9d9910: Removed constraints on the Space and IsDecorated template parameters for builtin functions taking a multi_ptr.
* https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/323de84e12d9df7b364a438260669cfd6df17a0c: Maintained builtin functions constraint on mutli_ptr parameters with constant_space and removed const qualifiers on global_ptr parameters to legacy `async_work_group_copy` overloads.
* https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/99c0b3b0a9527316bb7375d38779763ba83ff602: Added fixes based on feedback; fix for async_work_group_copy definition, removal of accessor_ptr for get_pointer() and additional get_pointer() declaration for the host_task case.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11863,7 +11863,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(local_ptr<DataT> dest,
-                                   global_ptr<const DataT> src,
+                                   global_ptr<DataT> src,
                                    size_t numElements) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -11876,7 +11876,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(global_ptr<DataT> dest,
-                                   local_ptr<const DataT> src,
+                                   local_ptr<DataT> src,
                                    size_t numElements) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -11889,7 +11889,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(local_ptr<DataT> dest,
-                                   global_ptr<const DataT> src,
+                                   global_ptr<DataT> src,
                                    size_t numElements, size_t srcStride) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -11902,7 +11902,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(global_ptr<DataT> dest,
-                                   local_ptr<const DataT> src,
+                                   local_ptr<DataT> src,
                                    size_t numElements, size_t destStride) const
 ----
     a@ Deprecated in SYCL 2020.
@@ -12391,7 +12391,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(local_ptr<DataT> dest,
-                                   global_ptr<const DataT> src,
+                                   global_ptr<DataT> src,
                                    size_t numElements) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -12404,7 +12404,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(global_ptr<DataT> dest,
-                                   local_ptr<const DataT> src,
+                                   local_ptr<DataT> src,
                                    size_t numElements) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -12417,7 +12417,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(local_ptr<DataT> dest,
-                                   global_ptr<const DataT> src,
+                                   global_ptr<DataT> src,
                                    size_t numElements, size_t srcStride) const
 ----
    a@ Deprecated in SYCL 2020.
@@ -12430,7 +12430,7 @@ a@
 ----
 template <typename DataT>
 device_event async_work_group_copy(global_ptr<DataT> dest,
-                                   local_ptr<const DataT> src,
+                                   local_ptr<DataT> src,
                                    size_t numElements, size_t destStride) const
 ----
     a@ Deprecated in SYCL 2020.
@@ -19707,8 +19707,11 @@ template<access::address_space Space, access::decorated IsDecorated>
 template<size_t N, access::address_space Space, access::decorated IsDecorated>
 ----
 
-* Unless specified otherwise, the template parameters [code]#Space# and
-  [code]#IsDecorated# are unconstrained.
+* Unless specified otherwise, the template parameter [code]#Space# is
+  constrained to the values [code]#access::address_space::global_space#,
+  [code]#access::address_space::local_space#,
+  [code]#access::address_space::private_space# or 
+  [code]#access::address_space::generic_space.
 
 * Other template parameters are not constrained to particular values unless a
   constraint is specifically mentioned in <<table.gentypes>>.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19707,14 +19707,8 @@ template<access::address_space Space, access::decorated IsDecorated>
 template<size_t N, access::address_space Space, access::decorated IsDecorated>
 ----
 
-* Unless specified otherwise, the template parameter [code]#Space# is
-  constrained to the values [code]#access::address_space::global_space#,
-  [code]#access::address_space::local_space#, or
-  [code]#access::address_space::private_space#.
-
-* Unless specified otherwise, the template parameter [code]#IsDecorated# is
-  constrained to the values [code]#access::decorated::yes# or
-  [code]#access::decorated::no#.
+* Unless specified otherwise, the template parameters [code]#Space# and
+  [code]#IsDecorated# are unconstrained.
 
 * Other template parameters are not constrained to particular values unless a
   constraint is specifically mentioned in <<table.gentypes>>.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6962,16 +6962,31 @@ accessor was constructed.  For other accessors, returns the default constructed
 a@
 [source]
 ----
-accessor_ptr<access::decorated::legacy> get_pointer() const noexcept
+global_ptr<access::decorated::legacy> get_pointer() const noexcept
 ----
-   a@ Returns a [code]#multi_ptr# to the start of this accessor's underlying
-      buffer, even if this is a <<ranged-accessor>> whose range does not start
-      at the beginning of the buffer.  The return value is unspecified if the
-      accessor is empty.
+   a@ Available only when [code]#(AccessTarget == target::device)#.
+
+Returns a [code]#multi_ptr# to the start of this accessor's underlying buffer,
+even if this is a <<ranged-accessor>> whose range does not start at the
+beginning of the buffer.  The return value is unspecified if the accessor is
+empty.
 
 This function may only be called from within a <<command>>.
 
 Deprecated in SYCL 2020.  Use [code]#get_multi_ptr# instead.
+
+a@
+[source]
+----
+std::add_pointer_t<value_type> get_pointer() const noexcept
+----
+   a@ Available only when [code]#(AccessTarget == target::host_task)#.
+
+Returns a pointer to the start of this accessor's underlying buffer, even if
+this is a <<ranged-accessor>> whose range does not start at the beginning of
+the buffer.  The return value is unspecified if the accessor is empty.
+
+This function may only be called from within a <<command>>.
 
 a@
 [source]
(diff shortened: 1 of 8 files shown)


adoc/headers/accessorBuffer.h
```diff
@@ -184,7 +184,7 @@ class accessor {
   cl::sycl::atomic<DataT, access::address_space::global_space>
   operator[](id<Dimensions> index) const;
 
-  std::add_pointer_t<value_type> get_pointer() const noexcept;
+  accessor_ptr<access::decorated::legacy> get_pointer() const noexcept;
```
gmlueck, line 187: The table says this function is deprecated.  I think it would be good to add a comment saying that here too, similar to what we have above for the other deprecated functions.

adoc/chapters/programming_interface.adoc
```diff
@@ -6954,6 +6954,20 @@ If this is a <<ranged-accessor>>, returns the offset that was specified when the
...
+----
+   a@ Returns a [code]#multi_ptr# to the start of this accessor's underlying
+      buffer, even if this is a <<ranged-accessor>> whose range does not start
+      at the beginning of the buffer.  The return value is unspecified if the
+      accessor is empty.
+
+This function may only be called from within a <<command>>.
+
+Deprecated in SYCL 2020.  Use [code]#get_multi_ptr# instead.
+
```
gmlueck, line 6970: I think it makes sense to remove `get_pointer` from Table 80 "Common buffer and local accessor member functions" (section 4.7.6.12).

You have already added a description here that is specific to `accessor`.  You should also add descriptions to:

* Table 72 "Member functions of the `host_accessor` class" (section 4.7.6.10.1)
* Table 78 "Member functions of the `local_accessor` class" (section 4.7.6.11.1)

adoc/headers/group.h
```diff
@@ -49,6 +49,32 @@ template <int Dimensions = 1> class group {
...
+                                     size_t numElements,
+                                     size_t srcStride) const;
+
+  // Deprecated in SYCL 2020.
+  template <typename DataT>
+  device_event async_work_group_copy(global_ptr<DataT> dest,
+                                     local_ptr<const DataT> src,
+                                     size_t numElements,
+                                     size_t destStride) const;
+
```
gmlueck, line 77: These new overloads should also be added to Table 118 "Member functions for the group class" (section 4.9.1.7).  I think the description could be simplified like this, though:

> Deprecated in SYCL 2020.  Has the same effect as the overload taking `decorated_local_ptr` and `decorated_global_ptr` except that the `dest` and `src` parameters have `access::decorated::legacy`.

adoc/headers/nditem.h
```diff
@@ -47,6 +47,33 @@ template <int Dimensions = 1> class nd_item {
...
+                                     global_ptr<const DataT> src,
+                                     size_t numElements,
+                                     size_t srcStride) const;
+
+  // Deprecated in SYCL 2020.
+  template <typename DataT>
+  device_event async_work_group_copy(global_ptr<DataT> dest,
+                                     local_ptr<const DataT> src,
+                                     size_t numElements,
+                                     size_t destStride) const;
```
gmlueck, line 74: Same comment here.  These functions should be added to Table 116 "Member functions for the nd_item class" in section  4.9.1.5.

adoc/headers/accessorBuffer.h
AerialMantis: Fixed in https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/b84963dbb7a353890317fcaa01c8c172a727ae68.

adoc/chapters/programming_interface.adoc
AerialMantis: That's a good point, `get_pointer` is no longer common, I've made this change in https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/b84963dbb7a353890317fcaa01c8c172a727ae68.

adoc/headers/group.h
AerialMantis: Good catch, fixed in https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/b84963dbb7a353890317fcaa01c8c172a727ae68.

adoc/headers/nditem.h
AerialMantis: Fixed in https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/b84963dbb7a353890317fcaa01c8c172a727ae68.

gmlueck:
> Currently based on https://github.com/KhronosGroup/SYCL-Docs/pull/428, will rebase once merged.

I think something went wrong with the rebase because the changes in this PR also show all the changes from #428 now.

keryell (approved):
Thanks.

AerialMantis:
> > Currently based on #428, will rebase once merged.
> 
> I think something went wrong with the rebase because the changes in this PR also show all the changes from #428 now.

I've re-targeted the PR to https://github.com/KhronosGroup/SYCL-Docs/pull/428 so it will only show the relevant changes now, but we'll need to update the target prior to merging.

fraggamuffin:
merge if 2nd approval comes in

gmlueck (commented):
Looks good, but I found a typo.

adoc/chapters/programming_interface.adoc
```diff
@@ -8068,6 +8082,16 @@ If this is a <<ranged-accessor>>, returns the offset that was specified when the
 accessor was constructed.  For other accessors, returns the default constructed
 [code]#id<Dimensions>{}#.
 
+a@
+[source]
+----
+std::add_pointer_t<value_type> get_pointer() const noexcept
+----
+   a@ Returns a [code]#multi_ptr# to the start of this accessor's underlying
```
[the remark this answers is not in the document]
AerialMantis: Good catch, thanks!

gmlueck (approved):

gmlueck:
> merge if 2nd approval comes in

I added an approval.  Can this be merged now?

AerialMantis:
Thanks @gmlueck!

@keryell we've noticed a couple of things we'd like to tweak in the wording so can we hold off merging this just yet, I should have an update later today.

AerialMantis:
@gmlueck @keryell I've pushed a patch with a few fixes based on feedback (https://github.com/KhronosGroup/SYCL-Docs/pull/432/commits/99c0b3b0a9527316bb7375d38779763ba83ff602), if you want re-review this. Thanks!

adoc/chapters/programming_interface.adoc
```diff
@@ -6959,6 +6959,35 @@ If this is a <<ranged-accessor>>, returns the offset that was specified when the
...
+
+Returns a pointer to the start of this accessor's underlying buffer, even if
+this is a <<ranged-accessor>> whose range does not start at the beginning of
+the buffer.  The return value is unspecified if the accessor is empty.
+
+This function may only be called from within a <<command>>.
+
 a@
 [source]
 ----
```
gmlueck, line 6993: While reviewing this, I noticed that there is a bug in the table entry below this comment.  The `get_multi_ptr` member function should be available only when `(AccessTarget == target::device)`.  We do not want to make a `multi_ptr` available to code running in a host task.

We could either include that fix as part of this PR, or create a new PR to fix it.

If you do fix it in this PR, remember also to add an "available only when" comment to the synopsis.

gmlueck (approved):
Your latest changes all look good.

I noticed an existing bug while doing this review, which I noted below.  I think it's OK to fix that in a separate PR, though, if you prefer, so I'm still approving this PR.

keryell:
> Your latest changes all look good.
> 
> I noticed an existing bug while doing this review, which I noted below. I think it's OK to fix that in a separate PR, though, if you prefer, so I'm still approving this PR.

Let's fix this in another PR then.

adoc/headers/multipointerlegacy.h
```diff
@@ -61,6 +61,10 @@ class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
   // Returns the underlying OpenCL C pointer
   pointer_t get() const;
 
+  std::add_pointer_t<value_type> get_raw() const;
```
steffenlarsen, line 64: Should this have used `ElementType` instead of `value_type`? `value_type` doesn't seem to be a thing in this specialization.

AerialMantis: Yeah good catch, the legacy specialization doesn't have `value_type` so this should be either `element_type` or `ElementType`.

I've created a follow-up PR to address this: https://github.com/KhronosGroup/SYCL-Docs/pull/437.

adoc/chapters/programming_interface.adoc
AerialMantis: Yeah that's a good point, I was thinking about this when I was making this change, we say that `accessor_ptr` is unspecified when targeting `host_trask` but we don't restrict `get_multi_ptr` from being defined, I think it makes sense to do this.

I've created a follow-up PR to address this: https://github.com/KhronosGroup/SYCL-Docs/pull/437.

adoc/headers/multipointerlegacy.h
AerialMantis: In second thought, I've decided to introduce `value_type` to the legacy specialization rather than changing `get_raw`, so that the `legacy` and non-`legacy` specializations are closer together.

PR #437, opened 2023-07-04
Add minor fixes to accessor/multi_ptr wording

AerialMantis:
Some minor issues were raised in https://github.com/KhronosGroup/SYCL-Docs/pull/432 after the PR was merged, so this is a follow-up PR to address those.

This PR adds a restriction for `accessor::get_multi_ptr` that it's only available when `AccessTarget` is `target::device` as it can't be used in a host task.

It adds the `value_type` alias to the `decorated::legacy` specialization of `multi_ptr` and adds the `get_raw` and `get_decorated` member functions to the `void`/`decorated::legacy` specializations of `multi_ptr`.

It also changes the alias used in `multi_ptr::get_raw` in the `decorated::legacy` specialization to `ElementType` as `value_type` is not defined.

cc @gmlueck @steffenlarsen

### DIFF
--- a/adoc/headers/multipointerlegacy.h
+++ b/adoc/headers/multipointerlegacy.h
@@ -61,7 +61,7 @@ class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
   // Returns the underlying OpenCL C pointer
   pointer_t get() const;
 
-  std::add_pointer_t<value_type> get_raw() const;
+  std::add_pointer_t<ElementType> get_raw() const;
 
   pointer_t get_decorated() const;
 
--- a/adoc/headers/multipointerlegacy.h
+++ b/adoc/headers/multipointerlegacy.h
@@ -8,6 +8,7 @@ namespace sycl {
 template <typename ElementType, access::address_space Space>
 class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
  public:
+  using value_type = ElementType;
   using element_type = ElementType;
   using difference_type = std::ptrdiff_t;
 
@@ -61,7 +62,7 @@ class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
   // Returns the underlying OpenCL C pointer
   pointer_t get() const;
 
-  std::add_pointer_t<ElementType> get_raw() const;
+  std::add_pointer_t<element_type> get_raw() const;
 
   pointer_t get_decorated() const;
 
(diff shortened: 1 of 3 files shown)


adoc/headers/multipointerlegacy.h
```diff
@@ -149,6 +150,7 @@ class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
 template <access::address_space Space>
 class [[deprecated]] multi_ptr<VoidType, Space, access::decorated::legacy> {
  public:
+  using value_type = ElementType;
```
steffenlarsen, line 153:
```suggestion
  using value_type = VoidType;
```

AerialMantis: Good catch, thanks!

adoc/headers/multipointerlegacy.h
```diff
@@ -61,7 +62,7 @@ class [[deprecated]] multi_ptr<ElementType, Space, access::decorated::legacy> {
   // Returns the underlying OpenCL C pointer
   pointer_t get() const;
 
-  std::add_pointer_t<value_type> get_raw() const;
+  std::add_pointer_t<element_type> get_raw() const;
```
gmlueck, line 65: Why change this now that you've added the `value_type` alias?  It seems better to keep it as `std::add_pointer_t<value_type>` to be consistent with the non-legacy specialization.

adoc/headers/multipointerlegacy.h
```diff
@@ -191,6 +193,10 @@ class [[deprecated]] multi_ptr<VoidType, Space, access::decorated::legacy> {
   // Returns the underlying OpenCL C pointer
   pointer_t get() const;
 
+  std::add_pointer_t<element_type> get_raw() const;
```
gmlueck, line 196: Here too, it seems better to use `std::add_pointer_t<value_type>`.

gmlueck (commented):
A couple of comments below.  In addition, this statement in the PR description seems confusing:

> It also changes the alias used in `multi_ptr::get_raw` in the `decorated::legacy` specialization to `ElementType` as `value_type` is not defined.

Because this PR adds the definition of `value_type`.

adoc/headers/multipointerlegacy.h
AerialMantis, line 65:
```suggestion
  std::add_pointer_t<value_type> get_raw() const;
```

AerialMantis: Good catch, this was a typo, should have been `value_type`, thanks. Same for the second case too.

gmlueck (approved):

keryell (approved):

fraggamuffin:
LGTM

PR #443, opened 2023-07-21
Fix range for half-precision cos/sin/tan

gmlueck:
Fix the precondition range for the half-precision `cos`, `sin`, and `tan` functions.  This appears to be a long-standing transcription bug when we first copied these function definitions from OpenCL.  Somehow we dropped the superscript notation, changing "2^16" into "216".

Also clarify that the range is inclusive, matching the definition in OpenCL 3.0.

Closes internal issue 659.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -23644,7 +23644,7 @@ template<typename NonScalar>      (2)
 
 *Overload (1):*
 
-_Preconditions:_ The value of [code]#x# must be in the range -216 to +216.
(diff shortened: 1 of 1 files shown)


gmlueck:
IMPORTANT!

keryell (approved):
Haha! This was a good one. I always thought it was some magical implementation details of OpenCL. :-)
But now it makes more sense.

AerialMantis (approved):
Good catch!

tomdeakin:
Good one! Ready to merge.

PR #444, opened 2023-07-21
Add "half" to geometric functions

gmlueck:
We seem to have unintentionally omitted `half` support from the Geometric Functions.  These functions derive from OpenCL, which does have "half" variations.  We added `half` to all the other builtin functions, so we just seem to have missed the Geometric ones.

Note that the `fast_distance`, `fast_length`, and `fast_normalize` functions do NOT have `half` variants (or `double` variants).  This is intentional, and it matches OpenCL.  The rationale is that these three functions all rely on the half-precision functions, which are only defined for `float`.

Closes internal issue 660.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -25427,12 +25427,16 @@ following types:
 
 * [code]#float#
 * [code]#double#
+* [code]#half#
 * [code]#marray<float, N>#, where [code]#N# is 2, 3, or 4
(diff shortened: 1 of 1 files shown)


gmlueck:
IMPORTANT!

keryell (approved):
Thanks!

AerialMantis (approved):

tomdeakin:
Ready to merge. Please find and close related issue.

PR #445, opened 2023-07-21
Remove reference to "denorms are flushed to zero"

gmlueck:
This appears to be an error from when we transcribed these function definitions from OpenCL.  There is no mode called "denorms are flushed to zero" in SYCL.  In OpenCL, this corresponds to the compiler flag "-cl-denorms-are-zero", but SYCL does not define any compiler flags.

Vendors could define a compiler flag like this if they want.  In this case, it would be the vendor's responsibility to document the effect.

Closes internal issue 657.


gmlueck:
IMPORTANT!

keryell (approved):
Good catch!

TApplencourt (approved):

AerialMantis (approved):

tomdeakin:
Ready to merge.

PR #446, opened 2023-07-21
Add marray<bool> support to "any" / "all"

gmlueck:
The "any" and "all" functions are not sensibly defined for SYCL when the input type is scalar or `marray`.  I think this happened because we copied the OpenCL semantics without thinking through the use cases.

The current definition looks at the most significant bit of each input element to determine if the result is true or false.  This makes sense for OpenCL, where true values are represented as -1 (all 1 bits) and false values are represented by 0.  It also makes sense in SYCL for the `vec` type because the relational functions set `vec` elements to -1 / 0.  However, it does not make sense when the parameter is `marray` or a scalar.

Consider a typical use of `any` like this:

```
marray<float, N> m1 = /*...*/;
marray<float, N> m2 = /*...*/;
marray<bool, N> comparisons = isequal(v1, v2);
if (any(comparisons)) {/*...*/}
```

Here, the `comparisons` variable is passed to `any`, where it does not make sense to test the most significant bit of each `bool` element.

This commit adds a new overload to `any` / `all` for `marray<bool>` with sensible C++ semantics.  It also deprecates the existing overloads for `marray` of non-bool types.

The scalar overloads of `any` / `all` are also deprecated.  The existing semantics (where the MSB is tested) do not make sense for SYCL.  In addition, these versions of `any` and `all` have already been deprecated in OpenCL.

Closes internal issue 658.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -26557,10 +26557,39 @@ a@
 a!
 [source]
 ----
-template<typename GenInt>
+template<typename GenInt>      (1)
 /*return-type*/ any(GenInt x)
+
+template<typename GenInt>      (2)  /* deprecated */
+bool any(GenInt x)
 ----
 !====
+
+*Overload (1):*
+
+_Constraints:_ Available only if [code]#GenInt# is one of the following types:
+
+* [code]#marray<bool, N>#
+* [code]#vec<int8_t, N>#
+* [code]#vec<int16_t, N>#
+* [code]#vec<int32_t, N>#
+* [code]#vec<int64_t, N>#
+* [code]#+__swizzled_vec__<int8_t, N>+#
+* [code]#+__swizzled_vec__<int16_t, N>+#
+* [code]#+__swizzled_vec__<int32_t, N>+#
+* [code]#+__swizzled_vec__<int64_t, N>+#
+
+_Returns:_ When [code]#x# is [code]#marray#, returns a Boolean telling whether
+any element of [code]#x# is true.  When [code]#x# is [code]#vec# or the
+[code]#+__swizzled_vec__+# type, returns the value -1 if any element in
+[code]#x# has its most significant bit set, otherwise returns the value 0.
+
+The return type is [code]#bool# if [code]#GenInt# is [code]#marray#.
+Otherwise, the return type is [code]#int#.
+
+*Overload (2):*
+
+This overload is deprecated in SYCL 2020.
 
 _Constraints:_ Available only if [code]#GenInt# is one of the following types:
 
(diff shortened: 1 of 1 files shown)


gmlueck:
IMPORTANT!

adoc/chapters/programming_interface.adoc
```diff
@@ -26557,11 +26557,40 @@ a@
...
+* [code]#+__swizzled_vec__<int32_t, N>+#
+* [code]#+__swizzled_vec__<int64_t, N>+#
+
+_Returns:_ When [code]#x# is [code]#marray#, returns a Boolean telling whether
+any element of [code]#x# is true.  When [code]#x# is [code]#vec# or the
+[code]#+__swizzled_vec__+# type, returns the value -1 if any element in
+[code]#x# has its most significant bit set, otherwise returns the value 0.
+
+The return type is [code]#bool# if [code]#GenInt# is [code]#marray#.
+Otherwise, the return type is [code]#int#.
```
keryell, line 26588: Wouldn't be an `int` with the same bit size of the input element instead?

gmlueck: No, I think the return type is `int` regardless of the element type of the `vec`.  This is consistent with the OpenCL definition of `any` and `all`, which also return an `int` regardless of the element type:

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#relational-functions

This is also consistent with the existing SYCL definitions of `any` and `all`.

AerialMantis (approved):

adoc/chapters/programming_interface.adoc
keryell: Furthermore I find suspicious the OpenCL design choice
```
int islessequal(double x, double y)
longn islessequal(doublen x, doublen y)
```
with returning a vector of masks for the vector API and a `bool` represented as an `int` for the scalar API.
Why a scalar comparison result could not be used as a mask too?
Anyway, the ship left the harbor long time ago. :-(
But I think the intent of OpenCL is to return a `bool` in the case of the scalar API but they returned an `int` because `bool` was not a native type before C23...
So, could SYCL be just type-safe and return a `bool` when we mean a `bool` instead of an `int`?

gmlueck: We do already return `bool` when the input type is a scalar or `marray`.  It's only for `vec` (and `__swizzled_vec__`) inputs that we return `int`.  This has been the case for a while in SYCL, so changing the return type to `bool` even for `vec` would be an API break.

I assume our logic was that these functions should be as close as possible to OpenCL for `vec` inputs because `vec` exists mostly for compatibility with OpenCL.

BTW, this comment made me realize that I made a mistake when clarifying the `any` and `all` functions.  For `vec` inputs, they should return 1 for true (not -1).  This is fixed in 84c9896292079020ad989fb72ee567029a3bfce1.

keryell (approved):
Thanks.

PR #447, opened 2023-07-25
Add restriction to "vec::as"

gmlueck:
This change was prompted by a discussion in KhronosGroup/SYCL-CTS#756, relating to a test of `vec::as`.  That test tried to use `vec::as` to reinterpret a `sycl::vec<float, 3>` as `sycl::vec<int, 4>`, which is legal according to the spec because they both have the same storage size.  (This is because a `vec` of length 3 has the same storage size as a `vec` of length 4.)  However, the semantics are unclear because the content of the 4th element is undefined for a 3 element `vec`.

It seems reasonable to simply state that such a use of `vec::as` is not allowed.  I think our intent was to allow `vec::as` to reinterpret only the bits of defined elements in a `vec`, which is what this change does.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -16989,7 +16989,12 @@ a@
 ----
 template <typename asT> asT as() const
 ----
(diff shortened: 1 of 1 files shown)


keryell (approved):
Avoiding UB when it is simple to solve makes sense.

TApplencourt (approved):

fraggamuffin:
LGTM

AerialMantis (approved):

PR #448, opened 2023-07-25
Limit vec alignment to 64 bytes

gmlueck:
We currently require `vec` to be aligned according to its total storage size.  For example, a `vec<uint32_t, 8>` must be aligned on a 32 byte boundary (32 = sizeof(uint32_t) * 8).  This comes from OpenCL, which has the same requirement.  The largest required alignment is for `vec<double, 16>`, which is typically 128 bytes.

However, this is problematic in SYCL because we want to support a variety of host compilers.  The MSVC compiler has a limit of 64 byte alignment for function parameters.  Therefore, the SYCL alignment requirements makes it impossible to pass a `vec<double, 16>` as a function parameter when compiling with MSVC.  This is particularly problematic because many of the SYCL "builtin" math function take `vec` as a parameter type.

We therefore propose weakening the alignment requirement to make it possible to support MSVC.  Of course, implementations are still allowed to align `vec` more highly if this results in more efficient code.

We considered removing the special alignment requirements entirely for `vec` and leaving this as an implementation detail.  However, existing code may depend on this alignment, so we opted for the most conservative change that still allows MSVC to work.  Therefore, we propose capping the required alignment to 64 bytes.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17654,6 +17654,13 @@ functions.
 
 In no case, however, is the alignment guaranteed to be greater than 64 bytes.
 
+[NOTE]
+====
+The alignment guarantee is limited to 64 bytes because some host compilers
+(e.g. on Microsoft Windows) limit the maximum alignment of function parameters
+to this value.
+====
+
 
 ==== Performance note
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17652,6 +17652,8 @@ This is true for both host and device code in order to allow for instances
 of the [code]#vec# class template to be passed to SYCL kernel
 functions.
 
+In no case, however, is the alignment guaranteed to be greater than 64 bytes.
+
 
 ==== Performance note
 


keryell (commented):
That is a new interesting case of conflicting requirements between host and device.
It is not clear about what are the consequences for some hardware.
A program failing to run on a pedantic OpenCL device?
I have the feeling that it prevents using compilation flow with different compilers for host and device when they do not agree on this alignment constraint because it would not be possible to inject a compiler pass into MSVC to arbitrary increase the alignment by adding some padding elements in some structs for example.

AerialMantis (approved):

keryell:
I am still confused about how it can work with a pedantic OpenCL implementation.
You have a `sycl::vec<double, 16>` on the host and you use it with USM à la HMM as a capture by address in a kernel.
The pointer you get has a 0.5 probability to be misaligned. So, what you propose is to have the device compiler to strip-mine any `sycl::vec<double, 16>`  operation by operations on `sycl::vec<double, 8>`? This could work, even if I admit some ugliness to it. :-)
More annoying is that if you use OpenCL interoperability mode to call an existing OpenCL kernel, then we fail on using `sycl::vec<double, 16>` which was designed to be compatible with OpenCL... :-(

gmlueck:
> More annoying is that if you use OpenCL interoperability mode to call an existing OpenCL kernel, then we fail on using `sycl::vec<double, 16>` which was designed to be compatible with OpenCL... :-(

This is a good point.  I think you're right, it's probably not possible to have a conformant implementation of SYCL on Windows that uses msvc as the host compiler and also uses a pedantic version of OpenCL that requires high alignment of `vec<double, 16>`.

The change proposed in this PR still gives implementors a choice, though.  They can either use msvc and an OpenCL implementation that tolerates `vec<double, 16>` aligned at 64 bytes, or they can use a different host compiler, or they can use a backend other than OpenCL.

keryell (approved):
Thanks for the argumentation.

tomdeakin:
Please add non-normative note justifying this.

psalz:
> Please add non-normative note justifying this.

Following up on this week's discussion on whether we should keep a changelog, this would be another good use-case IMO: including a quick note as to *why* a change was made, if it is not immediately obvious (like for a bug fix).

gmlueck:
> Please add non-normative note justifying this.

Done in 8d41edcab0e76dffea686ea61463876367873b47

PR #449, opened 2023-07-27
Fix definition of "nextafter" for double and half

gmlueck:
The `nextafter` function returns the next floating-point value that is representable in the given type.  The previous description was wrong because it specifically stated that it returns the next "single-precision" value.

I think we adopted the previous wording from OpenCL, and the OpenCL committee recently fixed this same wording in their spec.

Closes internal issue 661.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -22379,10 +22379,10 @@ template<typename NonScalar1, typename NonScalar2>     (4)
 
 *Overloads (1) - (3):*
 
-_Returns:_ The next representable single-precision floating-point value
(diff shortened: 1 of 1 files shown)


gmlueck:
IMPORTANT!

This is currently a "stacked" PR on top of https://github.com/KhronosGroup/SYCL-Docs/pull/440. Once https://github.com/KhronosGroup/SYCL-Docs/pull/440 is merged, this PR should be retargeted to the "SYCL-2020/master" branch.

keryell (approved):
Good catch!

TApplencourt (approved):

AerialMantis (approved):

tomdeakin:
Ready to merge.

PR #453, opened 2023-08-10
Introduce a default constructor for sycl::range

### LINKED ISSUE #90: Consider adding default constructor for cl::sycl::range
Currently the spec does define a default constructor for `cl::sycl::id`, but not for ` cl::sycl::range` . I'm not sure whether this is intentional or an oversight, however it can be quite a hassle when trying to write generic code.

For example consider this:
```c++
template <int dimensions>
struct my_range_holder {  cl::sycl::range<dimensions> range; };
```

There is no easy way of writing a constructor for this struct, without having to resort to some kind of range-initialization functor, or `if constrexpr` in C++17, both of which are far from pretty.

This issue first came to my attention because the Intel implementation seems to be the only one that actually doesn't provide some kind of default constructor for `cl::sycl::range`  (thus "breaking" my code :)). I did a quick survey of the rest of the implementations, and it looks like everybody is doing something slightly different:

| Implementation |                     |
| -------------- | ------------------- |
| ComputeCpp     | Initializes to 1    |
| hipSYCL        | Initializes to 0    |
| triSYCL        | Does not initialize  |

Clearly this is not great for portability, and even if you decide not to add a default constructor, at least consider explicitly deleting it in the spec. In my opinion, initializing ranges with 1, as ComputeCpp is doing, seems to be the most natural solution.

psalz:
As discussed in last week's call, this adds a default constructor for `sycl::range` that initializes all components to 0, analogous to `id`'s default constructor, making it more convenient to use in generic contexts. I initially brought this up over three years ago in #90, and there has been some disagreement on whether to initialize all components to 0 or 1. We finally agreed to all 0s for consistency with `sycl::id` (among other reasons).

Resolves #90.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11032,6 +11032,13 @@ include::{header_dir}/range.h[lines=4..-1]
 a@
 [source]
 ----
+range()
+----
+   a@ Construct a SYCL [code]#range# with the value [code]#0# for each dimension.
+
+a@
+[source]
+----
 range(size_t dim0)
 ----
    a@ Construct a 1D range with value dim0.
--- a/adoc/headers/range.h
+++ b/adoc/headers/range.h
@@ -6,6 +6,8 @@ template <int Dimensions = 1> class range {
  public:
   static constexpr int dimensions = Dimensions;
 
+  range();
+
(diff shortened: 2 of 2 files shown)


keryell (approved):
Make sense for consistency.

fraggamuffin:
LGTM

PR #460, opened 2023-09-04
Fix a table contexts swapped problem.

yingcong-wu:
The contents of `info::device::max_parameter_size` and `info::device::mem_base_addr_align` are swapped, fixed with this PR.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2052,8 +2052,9 @@ a@
 info::device::max_parameter_size
 ----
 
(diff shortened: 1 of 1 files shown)


gmlueck:
Hi @yingcong-wu,

When I look at the descriptions of `info::device::max_parameter_size` and `info::device::mem_base_addr_align`, the order looks correct to me.  Can you explain why you think something should be changed?  Maybe there is some clarification that we should make.

yingcong-wu:
Sorry, I must have misunderstood the meanings of the items. I will close this PR.

PR #462, opened 2023-09-07
Clarify operations on swizzled vec

gmlueck:
[This comment](https://github.com/KhronosGroup/SYCL-Docs/pull/440#discussion_r1302035263) in #440 noted that the number of elements in a `__swizzled_vec__` is not clearly defined.  Clarify this and also clarify that the member functions operate on the result of the swizzle operation.


keryell (approved):
Thanks for the clarification.

TApplencourt (approved):

gmlueck:
This was approved in the WG call 9/7/2023.

PR #464, opened 2023-09-21
Fix prvalue to xvalue conversion address space assignment

Naghasan:
The common address deduction rules assumed prvalue-to-xvalue conversion can only happens inside a function.
The patch aligns the rules to match declaration rules.

It is a bit of a corner case, but the spec treated all 3 conversions bellow in the same way.

```cpp
constexpr const int& foo = 5;

class Foo {
public:
  static constexpr const int& foo = 42;
};

void bar() {
    const int& foo = 42;
}
```

### DIFF
--- a/adoc/chapters/device_compiler.adoc
+++ b/adoc/chapters/device_compiler.adoc
@@ -737,8 +737,8 @@ The variable declarations get assigned to an address space depending on their
 scope and storage class:
 
   * Namespace scope
-  ** If the type is [code]#const#, the address space the declaration is assigned
-     to is implementation-defined.
+  ** If the type is [code]#const#, the address space of the declaration is
+     assigned to is implementation-defined.
      If the target of the SYCL backend can represent the generic address space,
      then the assigned address space must be compatible with the generic address
      space.
@@ -761,9 +761,11 @@ device kernel or code called by the device kernel.
   ** Static data members are treated the same way as for variable in namespace
      scope
 
-The result of a prvalue-to-xvalue conversion is assigned to the local address
-space if it happens in a hierarchical context or to the private address space
-otherwise.
+If a prvalue-to-xvalue conversion happens in a block scope or function parameter
+scope, the result is assigned to the local address space if it happens a
+hierarchical context otherwise it is assigned to the private address space.
+It the prvalue-to-xvalue conversion happens in another scope, the result is
+assigned in the same way as declaration in namespace scope.
 
 
 [[subsec:genericAddressSpace]]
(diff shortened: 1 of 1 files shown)


adoc/chapters/device_compiler.adoc
```diff
@@ -748,9 +748,11 @@ code called by the device kernel.
   ** Static data members are treated the same way as for variable in
      namespace scope
 
-The result of a prvalue-to-xvalue conversion is assigned to the local
-address space if it happens in a hierarchical context or to the private
-address space otherwise.
+If a prvalue-to-xvalue conversion happens in a block scope or function parameter scope,
+the result is assigned to the local address space if it happens a hierarchical context
```
keryell, line 752:
```suggestion
the result is assigned to the local address space if it happens in a hierarchical context
```

keryell:
@Naghasan ping!

adoc/chapters/device_compiler.adoc
Naghasan: I addressed it in the follow up commit.

Naghasan:
@keryell  I extended the initial text to also account for the context in which prvalue to xvalue is taking place. I find it a bit "hand wavy" but still acceptable.

The new commit also account for this:

```cpp
void bar() {
    static constexpr const int& foo = 42;
}
```

Naghasan:
Resolved conflict: rebased to head and ran reflow

adoc/chapters/device_compiler.adoc
```diff
@@ -737,8 +737,8 @@ The variable declarations get assigned to an address space depending on their
 scope and storage class:
 
   * Namespace scope
-  ** If the type is [code]#const#, the address space the declaration is assigned
-     to is implementation-defined.
+  ** If the type is [code]#const#, the address space of the declaration is
+     assigned to is implementation-defined.
```
gmlueck, line 741: This wording no longer makes sense.  The original wording was a little awkward, but it did make sense.  Are we trying to say this:

> If the type is `const`, the declaration is assigned to an implementation-defined address space.

adoc/chapters/device_compiler.adoc
```diff
@@ -761,9 +761,14 @@ device kernel or code called by the device kernel.
   ** Static data members are treated the same way as for variable in namespace
      scope
 
-The result of a prvalue-to-xvalue conversion is assigned to the local address
-space if it happens in a hierarchical context or to the private address space
-otherwise.
+If a prvalue-to-xvalue conversion happens as part of the an initialization
```
gmlueck, line 764:
```suggestion
If a prvalue-to-xvalue conversion happens as part of an initialization
```

adoc/chapters/device_compiler.adoc
Naghasan: reads better, I'll apply this.

gmlueck (approved):

PR #468, opened 2023-10-05
Allow platforms to be empty

illuhad:
As discussed, this PR allows platforms to be empty. The motivating example was that of a CUDA backend where `CUDA_VISIBILE_DEVICES` has been set to expose no devices.

Most info queries like `platform::vendor` or `platform::name` can still be answered in this case. For the list of supported extensions, this PR assumes that an empty platform will not advertise any extension.

Another option might be to say that if an empty platform chooses to still advertise some extensions, they must be supported for all devices potentially supported by the platform, even if those devices are unavailable at the moment. In the example of the CUDA backend, we would e.g. know that any device that could be exposed once `CUDA_VISIBLE_DEVICES` is unset would also support USM.
I don't have a strong opinion on whether this additional rule is worth the effort.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1030,8 +1030,9 @@ The SYCL [code]#platform# class encapsulates a single SYCL platform on
 which SYCL kernel functions may be executed. A SYCL platform must be
 associated with a single <<backend>>.
 
-A SYCL [code]#platform# is also associated with one or more SYCL
-[code]#devices# associated with the same <<backend>>.
+A SYCL [code]#platform# is also associated with any number of SYCL
+[code]#devices# associated with the same <<backend>>. A platform might also
+contain no devices.
 
 All member functions of the [code]#platform# class are synchronous and
 errors are handled by throwing synchronous SYCL exceptions.
@@ -1139,7 +1140,8 @@ a@
 bool has(aspect asp) const
 ----
    a@ Returns true if all of the SYCL [code]#devices# associated with this
-      SYCL [code]#platform# have the given <<aspect>>.
+      SYCL [code]#platform# have the given <<aspect>>. If the platform does not
+      contain any devices, returns [code]#false#.
 
 a@
 [source]
@@ -1148,7 +1150,7 @@ bool has_extension(const std::string& extension) const
 ----
    a@ Deprecated, use [code]#has()# instead.
 
-Returns true if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension.
+Returns true if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension. Returns [code]#false# for a platform that does not contain any devices.
 
 a@
 [source]
@@ -1236,7 +1238,7 @@ info::platform::extensions
    a@ Deprecated, use [code]#device::get_info()# with
       [code]#info::device::aspects# instead.
 
-Returns the extensions supported by the <<platform>>.
+Returns the extensions supported by the <<platform>>. Returns an empty list if the platform does not contain any devices.
 
 |====
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1031,7 +1031,7 @@ which SYCL kernel functions may be executed. A SYCL platform must be
 associated with a single <<backend>>.
 
 A SYCL [code]#platform# is also associated with any number of SYCL
-[code]#devices# associated with the same <<backend>>. A platform might also
+[code]#devices# associated with the same <<backend>>. A platform may
 contain no devices.
 
 All member functions of the [code]#platform# class are synchronous and
@@ -1140,8 +1140,8 @@ a@
 bool has(aspect asp) const
 ----
    a@ Returns true if all of the SYCL [code]#devices# associated with this
-      SYCL [code]#platform# have the given <<aspect>>. If the platform does not
-      contain any devices, returns [code]#false#.
+      SYCL [code]#platform# have the given <<aspect>>. Returns [code]#false# if this SYCL [code]#platform# does not
+      contain any devices.
 
 a@
 [source]
@@ -1150,7 +1150,7 @@ bool has_extension(const std::string& extension) const
 ----
    a@ Deprecated, use [code]#has()# instead.
 
-Returns true if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension. Returns [code]#false# for a platform that does not contain any devices.
+Returns [code]#true# if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension. Returns [code]#false# if this SYCL [code]#platform# does not contain any devices.
 
 a@
 [source]
@@ -1238,7 +1238,7 @@ info::platform::extensions
    a@ Deprecated, use [code]#device::get_info()# with
       [code]#info::device::aspects# instead.
 
-Returns the extensions supported by the <<platform>>. Returns an empty list if the platform does not contain any devices.
+Returns the extensions supported by this [code]#platform#. Returns an empty list if this [code]#platform# does not contain any devices.
 
 |====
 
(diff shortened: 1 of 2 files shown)


TApplencourt (approved):

gmlueck:
One of the changes we have discussed for the next version of the SYCL specification is to add the concept of a platform "default context" and provide a guarantee that `queue` constructors that do not specifically request a context use the platform's default context.  This is an important addition to the spec because it provides a guarantee for most simple applications that queues for devices on the same platform use the same context, and this provides guarantees about the ability to share USM memory amongst kernels submitted to those devices.

This PR interacts with that proposed addition because it would provide a way for the user to get a `context` that has no devices.  (The proposal adds a way to get the default context for a platform, so a platform with no devices would necessarily have a default context with no devices.)

Therefore, if we decide to allow platforms to have no devices, I think we also need to clarify the semantics of a context that has no devices.

It's also not clear to me what requirements this PR imposes on an implementation.  Must an implementation expose platforms with no devices if they could theoretically support some platform?  It's not clear to me what this would even mean.  Just for the sake of argument, let's suppose a vendor has a SYCL implementation that can support backend _B_, but that support must be enabled when the SYCL compiler itself is compiled (e.g. via a build option when building the compiler).  Must this implementation expose an empty platform for backend _B_ when it is built _without_ that support?

As another scenario, let's suppose a SYCL implementation could potentially support the OpenCL backend, but only when the OpenCL runtime is installed.  If that runtime is not installed, must the implementation expose an empty platform for OpenCL?

The motivation listed above for this PR is the `CUDA_VISIBLE_DEVICES` environment variable.  The DPC++ implementation has a similar environment variable `ONEAPI_DEVICE_SELECTOR` which allows the user to select which devices and backends are exposed to the SYCL application.  Our current behavior is to not expose a platform if this environment variable deselects the associated backend (or deselects all the devices in the backend).  We have had an environment variable like this for a long time and no one has complained about that behavior.  Would this PR require that behavior to change?

TApplencourt:
I think the implementation is free to decide if it wants to show an empty platform.

>(The proposal adds a way to get the default context for a platform, so a platform with no devices would necessarily have a default context with no devices.)

Can't it just `throw` if the platform has no device?

illuhad:
> One of the changes we have discussed for the next version of the SYCL specification is to add the concept of a platform "default context" and provide a guarantee that queue constructors that do not specifically request a context use the platform's default context. This is an important addition to the spec because it provides a guarantee for most simple applications that queues for devices on the same platform use the same context, and this provides guarantees about the ability to share USM memory amongst kernels submitted to those devices.
> This PR interacts with that proposed addition because it would provide a way for the user to get a context that has no devices. (The proposal adds a way to get the default context for a platform, so a platform with no devices would necessarily have a default context with no devices.)
> Therefore, if we decide to allow platforms to have no devices, I think we also need to clarify the semantics of a context that has no devices.

It's a bit difficult to discuss the impact on a change that so far has only been vaguely discussed. I don't recall that any actual API or presentation on details was made on the topic of default context.
With that said, I don't think that there's a fundamental issue. In similar fashion to what the PR already states, we could also say that the atomic capability info queries for `context` return empty list if there are no devices.

We also have had default-context semantics in our implementation since the very beginning (and indeed, the `context` class is largely meaningless for us), so I don't think that a default context is fundamentally at odds with this PR.

> It's also not clear to me what requirements this PR imposes on an implementation. Must an implementation expose platforms with no devices if they could theoretically support some platform? It's not clear to me what this would even mean. Just for the sake of argument, let's suppose a vendor has a SYCL implementation that can support backend B, but that support must be enabled when the SYCL compiler itself is compiled (e.g. via a build option when building the compiler). Must this implementation expose an empty platform for backend B when it is built without that support?

The PR makes no such requirements. It just says that an implementation is allowed to return empty platforms. When that happens - if at all - is up to the implementation. It gives implementations more freedom.

> As another scenario, let's suppose a SYCL implementation could potentially support the OpenCL backend, but only when the OpenCL runtime is installed. If that runtime is not installed, must the implementation expose an empty platform for OpenCL?

Again, this PR makes no such requirements. It just gives implementations the freedom to return empty platforms if they wish to do so.

> The motivation listed above for this PR is the CUDA_VISIBLE_DEVICES environment variable. The DPC++ implementation has a similar environment variable ONEAPI_DEVICE_SELECTOR which allows the user to select which devices and backends are exposed to the SYCL application. Our current behavior is to not expose a platform if this environment variable deselects the associated backend (or deselects all the devices in the backend). We have had an environment variable like this for a long time and no one has complained about that behavior. Would this PR require that behavior to change?

No, this PR would not require any change to your implementation. Implementations can continue to choose to *not* return empty platforms. It just allows an alternative behavior.
Let me know if you have suggestions how this can perhaps be expressed more clearly in the PR.

psalz (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -1030,8 +1030,9 @@ The SYCL [code]#platform# class encapsulates a single SYCL platform on
 which SYCL kernel functions may be executed. A SYCL platform must be
 associated with a single <<backend>>.
 
-A SYCL [code]#platform# is also associated with one or more SYCL
-[code]#devices# associated with the same <<backend>>.
+A SYCL [code]#platform# is also associated with any number of SYCL
+[code]#devices# associated with the same <<backend>>. A platform might also
```
AerialMantis, line 1034:
```suggestion
[code]#devices# associated with the same <<backend>>. A platform may
```

adoc/chapters/programming_interface.adoc
```diff
@@ -1148,7 +1150,7 @@ bool has_extension(const std::string& extension) const
 ----
    a@ Deprecated, use [code]#has()# instead.
 
-Returns true if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension.
+Returns true if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension. Returns [code]#false# for a platform that does not contain any devices.
```
AerialMantis, line 1153:
```suggestion
Returns [code]#true# if this SYCL [code]#platform# supports the extension queried by the [code]#extension# parameter. A SYCL [code]#platform# can only support an extension if all associated SYCL [code]#devices# support that extension. Returns [code]#false# if this SYCL [code]#platform# does not contain any devices.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -1139,7 +1140,8 @@ a@
 bool has(aspect asp) const
 ----
    a@ Returns true if all of the SYCL [code]#devices# associated with this
-      SYCL [code]#platform# have the given <<aspect>>.
+      SYCL [code]#platform# have the given <<aspect>>. If the platform does not
+      contain any devices, returns [code]#false#.
```
AerialMantis, line 1144:
```suggestion
      contain any devices.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -1236,7 +1238,7 @@ info::platform::extensions
    a@ Deprecated, use [code]#device::get_info()# with
       [code]#info::device::aspects# instead.
 
-Returns the extensions supported by the <<platform>>.
+Returns the extensions supported by the <<platform>>. Returns an empty list if the platform does not contain any devices.
```
AerialMantis, line 1241:
```suggestion
Returns the extensions supported by this [code]#platform#. Returns an empty list if this [code]#platform# does not contain any devices.
```

AerialMantis:
The changes looks good to me, I have suggested some minor edits for consistency. I also noticed that there are some references to the requirement that a platform always contain at least on device is also described in the architecture section of the specification, so I think we need to change the wording there too.

_§3.5. The SYCL platform model_

> The SYCL platform model is based on the OpenCL platform model. The model consists of a host connected to one or more heterogeneous devices, called [devices](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#device).

_§3.6.1. Platform mixed version support_

> The SYCL generic programming model exposes a number of [platforms](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#platform), each of them exposing a number of [devices](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#device).

_§3.7.1.1. SYCL backend resources managed by the SYCL application_

> Devices: platforms provide one or more devices for executing SYCL kernels.

We may also want to drop the "based on the OpenCL platform model" wording.

> Therefore, if we decide to allow platforms to have no devices, I think we also need to clarify the semantics of a context that has no devices.

I think we could handle this by having the `context` and `queue` default constructors throw if there is no device found.

illuhad:
@AerialMantis I've changed the places you have highlighted to make clear that empty platforms are allowed too. I've also removed the reference to the OpenCL platform model.

keryell (approved):

AerialMantis (approved):
Looks good, thanks!

PR #471, opened 2023-10-11
Remove confusing wording about errors

gmlueck:
I noticed this while reviewing work on the SYCL reference guide.

The specification for the member functions `queue::get_context` and `queue::get_device` implied that they could throw an exception for an error scenario.  I cannot think of any reason these functions could fail because a queue always has a context and a device.  Therefore, I think we should remove these sentences.  For example, notice that there is no similar wording for `queue::is_in_order`, which also has no obvious reason to fail.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3184,7 +3184,6 @@ a@
 context get_context() const
 ----
 a@ Returns the SYCL queue's context.
-Reports errors using SYCL exception classes.
 The value returned must be equal to that returned by [code]#get_info<info::queue::context>()#.
 
 a@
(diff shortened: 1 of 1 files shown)


keryell:
I guess there is always the possibility that the device is no longer available for some reason.

gmlueck:
> I guess there is always the possibility that the device is no longer available for some reason.

But we have already decided that SYCL 2020 does not hot-swapped devices:

> The execution environment for a SYCL application has a fixed number of [root devices](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#root-device) which does not vary as the application executes.

(https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-class)

TApplencourt:
I agree with Greg that for SYCL 2020, I don't see how they can fail. (for SYCL SC Next this is another topic :) )

TApplencourt (approved):

keryell (approved):

PR #473, opened 2023-10-12
Don't require an implementation to expose at least one device

psalz:
As discussed in last week's meeting, we're removing the requirement for an implementation to expose at least one device. This is mainly to avoid undefined behavior in cases where for example all devices are hidden from the implementation through environment variables such as `CUDA_VISIBLE_DEVICES`.

It looks like the only location that really needed changing was the wording regarding the default device selector. Other than that, I wasn't sure whether we may also want to add an additional sentence or two to the descriptions of the default constructors for `platform`, `context` and `queue`, which currently say something like

> Constructs a SYCL platform instance that is a copy of the platform which contains the device returned by default_selector_v.

### DIFF
--- a/adoc/chapters/what_changed.adoc
+++ b/adoc/chapters/what_changed.adoc
@@ -373,8 +373,7 @@ The [code]#cl::sycl::byte# has been deprecated and now the {cpp17}
 [code]#std::byte# should be used instead.
 
 A SYCL implementation is no longer required to provide a host device.
-Instead, an implementation is only required to provide at least one
-device.  Implementations are still allowed to provide devices that are
+Implementations are still allowed to provide devices that are
 implemented on the host, but it is no longer required.  The specification
 no longer defines any special semantics for a "host device" and APIs
 specific to the host device have been removed.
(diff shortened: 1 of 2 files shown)


adoc/chapters/what_changed.adoc
```diff
@@ -373,8 +373,7 @@ The [code]#cl::sycl::byte# has been deprecated and now the {cpp17}
 [code]#std::byte# should be used instead.
 
 A SYCL implementation is no longer required to provide a host device.
-Instead, an implementation is only required to provide at least one
-device.  Implementations are still allowed to provide devices that are
+Implementations are still allowed to provide devices that are
```
psalz, line 376: I'm actually not really happy with changing the changelog. Alternative ideas:

- Use ~~strikethrough~~ and add a sentence that this was later reverted.
- Add a new subsection `Changes made in revision 9` that details this.

tomdeakin:
WG will discuss this for SYCL Next, but least SYCL 2020 behaviour as it is.

psalz:
Moving to https://gitlab.khronos.org/sycl/Specification/-/merge_requests/737.

PR #479, opened 2023-10-25
Refine `vec` interface code snippet

AlexeySachkov:
Changes:
  - removed some unnecessary indentation;
  - removed `{ /* ... */ }`;
  - re-ordered operator declarations to have `DataT, vec` overloads next to other overloads for the same operator;
  - moved declarations of `operator||` and `operator&&` closer to other relational operators to avoid disrupting biwtise operators block;

### DIFF
--- a/adoc/headers/vec.h
+++ b/adoc/headers/vec.h
@@ -121,112 +121,65 @@ template <typename DataT, int NumElements> class vec {
   // OP is: +, -, *, /, %
   /* If OP is %, available only when: DataT != float && DataT != double
   && DataT != half. */
-  friend vec operatorOP(const vec& lhs, const vec& rhs) { /* ... */
-  }
-  friend vec operatorOP(const vec& lhs, const DataT& rhs) { /* ... */
-  }
+  friend vec operatorOP(const vec& lhs, const vec& rhs);
+  friend vec operatorOP(const vec& lhs, const DataT& rhs);
+  friend vec operatorOP(const DataT& lhs, const vec& rhs);
 
   // OP is: +=, -=, *=, /=, %=
   /* If OP is %=, available only when: DataT != float && DataT != double
   && DataT != half. */
-  friend vec& operatorOP(vec& lhs, const vec& rhs) { /* ... */
-  }
-  friend vec& operatorOP(vec& lhs, const DataT& rhs) { /* ... */
-  }
+  friend vec& operatorOP(vec& lhs, const vec& rhs);
+  friend vec& operatorOP(vec& lhs, const DataT& rhs);
 
   // OP is prefix ++, --
   // Available only when: DataT != bool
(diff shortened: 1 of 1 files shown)


AlexeySachkov:
I've just realized that the table below the snippet follows the same order of operators as the snippet.

If there are no major objections against this change and we want to have matching order between the snippet and the table, I can align them as well, just let me know

keryell (approved):
Thanks for making the specification clearer and shorter.

adoc/headers/vec.h
```diff
@@ -121,112 +121,65 @@ template <typename DataT, int NumElements> class vec {
...
+  friend vec<RET, NumElements> operatorOP(const vec& lhs, const DataT& rhs);
+  friend vec<RET, NumElements> operatorOP(const DataT& lhs, const vec& rhs);
+
+  // OP is: ==, !=, <, >, <=, >=
+  friend vec<RET, NumElements> operatorOP(const vec& lhs, const vec& rhs);
+  friend vec<RET, NumElements> operatorOP(const vec& lhs, const DataT& rhs);
+  friend vec<RET, NumElements> operatorOP(const DataT& lhs, const vec& rhs);
+
+  vec& operator=(const vec<DataT, NumElements>& rhs);
+  vec& operator=(const DataT& rhs);
```
gmlueck, line 178: Let's move these assignment operators, so they appear right after the subscript operators.  This will match the order they have in the table.

adoc/headers/vec.h
```diff
@@ -121,112 +121,65 @@ template <typename DataT, int NumElements> class vec {
...
+
+  /* Available only when: DataT != float && DataT != double && DataT != half. */
+  friend vec operator~(const vec& v);
+  friend vec<RET, NumElements> operator!(const vec& v);
```
gmlueck, line 182: Can you put a blank link between these two?  This will help make it clear that the constraint ("Available only when ...") applies only to `operator~`.  The table already makes this clear too.

gmlueck (commented):
Thanks for making this change; it definitely helps with the readability of the synopsis.

> I've just realized that the table below the snippet follows the same order of operators as the snippet.
>
> If there are no major objections against this change and we want to have matching order between the snippet and the table, I can align them as well, just let me know

Yes, I think we want the order of the entries in the table to continue to match the order in the synopsis.  I like the arrangement you have in the synopsis now.  Can you also reorder the "operatorOP" entries in the table to match your new synopsis order?

I think we plan to merge #481 soon, so you might want to wait for that to happen in order to avoid a big merge conflict.

adoc/headers/vec.h
AlexeySachkov: Done in c42cf1a23905537df233b96500914040b772b1ce

gmlueck (approved):

Issue #472, opened 2023-10-12
Switch to one sentence per line?


psalz:
I'd like to propose switching to "[semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/)", i.e., having one sentence per line.

The spec currently mostly uses a 80-ish character limit, like in the olden days. Imposing a limit might make sense for program code, but for text it doesn't, in my opinion. One issue this introduces is that it is more difficult to CTRL+F something in the spec source, as a sentence might randomly contain a line break. In some places on the other hand, entire paragraphs are written without a single line break.

Both of these styles are not ideal from a version control perspective, as changing a single word can often cause large diffs spanning several lines, making it difficult to see what exactly has changed (yes, I'm aware there's word-diffs, but still), and may result in annoying merge conflicts.

Some might argue that having hard breaks between line-wrapped sentences makes the adoc source more difficult to read. This may be true for prose, but for a specification, where sentences often are more akin to disjoint enumerations of rules, it makes it easier read, in my opinion.

Here's an example from `Table 103. Parameterized USM Allocation Functions`, which currently uses a 80 character limit:

![image](https://github.com/KhronosGroup/SYCL-Docs/assets/791348/077cccc8-1da8-458d-937f-aeb53e851d96)

After changing it to one sentence per line:

![image](https://github.com/KhronosGroup/SYCL-Docs/assets/791348/36dda8e7-eef2-44b0-ba76-3ba969c62417)

And here's a paragraph from `Table 143. Hidden friend functions of the vec class template`, which currently has multiple sentences per line:

![image](https://github.com/KhronosGroup/SYCL-Docs/assets/791348/f16feef9-7b61-411e-a03b-a1468e6bdeb1)

After adding line breaks after each sentence:

![image](https://github.com/KhronosGroup/SYCL-Docs/assets/791348/cba03009-6a46-4c05-8f53-9303fb7d7e8e)

Yes, it's no longer as "neat", but IMO it is much easier to find a relevant sentence when quickly looking over it.

Thoughts?

gmlueck:
[...]

PR #481, opened 2023-10-25
Run reflow

TApplencourt:
Update our `reflow.py` copy and run it.

Link: https://github.com/KhronosGroup/Vulkan-Docs/blob/main/scripts/reflow.py and thanks @oddhack  :)

This does what we wanted:
 - Force a maximum line width (default 76 char)
 - Split on period 
 - Delete 2+ contiguous spaces
 
 This is a small PR, so it should be easy to review...


adoc/chapters/acknowledgements.adoc
```diff
@@ -48,8 +48,10 @@
   * Michael Wong, Codeplay
   * Peter Žužek, Codeplay
   * Matt Newport, EA
-  * Rasool Maghareh, Huawei Technologies Co. Ltd.
-  * Guansong Zhang, Huawei Technologies Co. Ltd.
+  * Rasool Maghareh, Huawei Technologies Co.
+    Ltd.
+  * Guansong Zhang, Huawei Technologies Co.
+    Ltd.
```
gmlueck, line 54: Not sure we want these newlines, but it does not hurt.

adoc/chapters/device_compiler.adoc
```diff
@@ -129,96 +136,98 @@ void h() {
 }
 ----
 
-In order for the SYCL device compiler to correctly compile <<device function, device functions>>, all
-functions in the source file, whether <<device function, device functions>> or not, must be
-syntactically correct functions according to this specification. A syntactically
-correct function adheres to at least the minimum required {cpp} version
-defined in <<sec:progmodel.minimumcppversion>>.
+In order for the SYCL device compiler to correctly compile <<device
```
gmlueck, line 139: This took me a while to track down ...

Asciidoctor does not allow you to split a document reference across two lines like this.  Therefore, you should break the line before the "<<device".  As it is now, this causes the following warning when you build the spec:

```
INFO: possible invalid reference: device
```

adoc/chapters/opencl_backend.adoc
```diff
@@ -703,18 +727,19 @@ code will be [code]#CL_SUCCESS#.
...
-[code]#cl_program# using [code]#clBuildProgram#,
+In <<opencl12, OpenCL>> any kernel function that is enqueued over an
+nd-range is represented by a [code]#cl_kernel# and must be compiled and
+linked via a [code]#cl_program# using [code]#clBuildProgram#,
 [code]#clCompileProgram# and [code]#clLinkProgram#.
 
-For OpenCL <<backend>> this detail is abstracted away by <<kernel bundle,kernel bundles>> and
-a [code]#kernel_bundle# object containing all <<sycl-kernel-function,SYCL kernel functions>>
-is retrieved by calling the free function [code]#get_kernel_bundle#.
+For OpenCL <<backend>> this detail is abstracted away by <<kernel
```
gmlueck, line 735: Same thing here.  The line break should be before "<<kernel".

gmlueck (commented):
Wow, this is great!  There are a couple problems I point out below, but I think they can be easily fixed.

To validate that nothing got inadvertently changed, I built the PDF both before and after this change and compared the PDF versions.  There were no unexpected changes.  (There were some expected changes on the title page related to the git version and the build date.)

To people who were not at lunch today, this addresses #472.  Thomas was noting how this could be easily done via a script, so I guess he volunteered!

adoc/chapters/device_compiler.adoc
oddhack: In the specs I am editor for, we use the convention that anchor names are hyphenated, e.g. \<\<kernel-bundle, kernel bundles>>.

oddhack:
A few other notes about this script:

- Requires pretty strict adherence to style guidelines, because it's not running the asciidoctor processing pipeline and does not know about every possible asciidoc construct. When I initially did the SYCL format conversion I followed the Vulkan style guidelines in general, so it's not surprising this basically works for you. The Vulkan style guide contains a lot of stuff that is irrelevant to SYCL but much of the markup / writing / miscellaneous sections may be worth extracting.
- Will not reflow long lines in captions and titles (not supported by asciidoc), nor inside tables (possible, but $TOOCOMPLEX for the minimal upside).

It is pretty robust within the scope of what it's intended for. We use it frequently on Vulkan.

oddhack:
@gmlueck it is an orthogonal issue, but you may want to review the Khronos spec copyright and make sure that's up to date with the latest version Neil has (V10, I think?).

adoc/chapters/device_compiler.adoc
TApplencourt: Good catch! I should have try to build it myself. I did a visual review, was obviously not enough. 

My 2cents: It may be easier for us to change the SYCL spec to use hyphen (the regex should be relatively pretty easy) than to try to patch `reflow.py`. At least for me :)

oddhack: For this type of thing, you can also do 'wdiff -3' to compare tokens in the HTML outputs with usually good results.

FWIW,

> git grep -E '<<[[:alpha:]:]+ '

turns up only these xrefs:

* 'command group'
* 'control flow'
* 'device copyable'
* 'device function'
* 'kernel bundle'

and AFAICT all of them are glossary terms which already have a corresponding hyphenated anchor.

TApplencourt: And we also use `<<device-function, Device functions>>`, so I guess it was so incoherence in our spec. Will fix.

gmlueck: Yes, I agree that we should change these references to use the (hyphenated) anchor name.  I did this in #482.

I'm surprised that Asciidoctor doesn't issue an error when there is a reference that does not have a matching anchor.  I guess the syntax "\<\<foo\>\>" will link to *either* an anchor "[[foo']]" or to a glossary term "foo"?

oddhack: You get anchors implicitly generated in some circumstances, for example section titles (and probably glossary entries, too, though not sure about that). I prefer having explicit control over this so I use explicit anchors everywhere.

There are varying levels of warning / error behavior selectable through command line options. They probably are not turned all the way up at present.

oddhack: BTW I think if you do something like '\<\<device
anchor, some text>>' you get a generated '\<a href="device
anchor">' which I guess? is legal HTML? But does not map to 'device-anchor'.

keryell (approved):
Looks globally like a good idea.
Probably you can rephrase "Keep a maximum line width," in the description to fit the mathematical property you want to express.

adoc/chapters/acknowledgements.adoc
oddhack: You can add more to the endAbbrev regexp in reflow.py.

TApplencourt: @oddhack would you prefer us to `fork` reflow.py or should we move `reflow` in a separate repo so it can be shared between projects?

oddhack: Go ahead and make a copy (you'll need a couple of other .py files from the scripts/ directory, see the import statements). I keep relevant scripts loosely synced up between the different repos using them but it's not worth the complexity of separate repos.

oddhack:
> @oddhack do you know why 76 characters in "Force a maximum line width (default 76 char)"? Why not 79?

Historical dating back to OpenGL extension days - use whatever is convenient to you. If you use 999999 you get one sentence/line :-) Just make sure you like the value you start with so you don't generate lots of git churn by changing it later.

keryell:
I mean, if we agree on 80 columns inherited from punch cards which percolated for decades into hardware terminals quite before OpenGL and computer graphics, why 76? You loose 4 characters... :-)

TApplencourt:
@gmlueck Added a CI test.

TApplencourt:
This is ready to merge (if we agree on line width...). Summary of changes:

- Copied Vulkan `reflow.py` and dependencies in our repo + updated abbrev list (thanks again @oddhack)
- Run `reflow` on all `chapters` 
- Added a CI script to verify chapters conformance with reflow  
- Minor: Updated some github receipt version

gmlueck:
What column width are you enforcing now?  Is it 76 or 80?  I agree with @keryell, 80 makes more sense than 76.  From @oddhack's response, it didn't seem like there was any particular reason to choose 76.

nliber:
80 columns makes sense -- if folks are still using ADM-5 terminals!  Even the mighty VT-100 had a 132 column mode.

(Honestly, just pick something...)

TApplencourt:
The default so 76. Agree, that 80 if most popular.  `black` -- the python formater -- did choose [88](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)  🤷🏽 

> You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).

If we have already two people pushing for 80, I can change to that. Don't want tobikeshedding it too much :)

TApplencourt:
Change default to 80. Funnily enough, it doesn't change anything :)

keryell:
The reason I propose 79 instead of 80 is that sometimes 80 is a problem when there is a 80-hard limit and some editors or modes display some EOL char at the end, like my Emacs whitespace-mode to show all the weird chars.

nliber:
Doesn't making it more than 72 columns make us incompatible with Fortran?

(Sorry; couldn't resist...)

adoc/scripts/vkconventions.py
```diff
@@ -0,0 +1,298 @@
...
+                structure_type_parts.append(word.upper())
+        name = '_'.join(structure_type_parts)
+
+        # The simple-minded rules need modification for some structure names
+        subpats = [
+            [ r'_H_(26[45])_',              r'_H\1_' ],
+            [ r'_VULKAN_([0-9])([0-9])_',   r'_VULKAN_\1_\2_' ],
+            [ r'_VULKAN_SC_([0-9])([0-9])_',r'_VULKAN_SC_\1_\2_' ],
+            [ r'_DIRECT_FB_',               r'_DIRECTFB_' ],
+            [ r'_VULKAN_SC_10',             r'_VULKAN_SC_1_0' ],
```
ProGTX, line 113: A lot of this is Vulkan specific, do we need this, or even this file?

ProGTX (approved):
LGTM, though I wonder if we need all of the Vulkan related stuff?

Also I agree with the 80 column limit, it's the most standard.

adoc/scripts/vkconventions.py
TApplencourt: It depends on your definition of `need` :). This file is "required" because of some import. But indeed, they seem to be related to some Vulkan conversion, so in theory, we can just do some modification and remove it. 

But saying that, I think the maintenance will be easier if we just copy/paste everything without "un-necessary-to-get-it-working" modification.  I agree that we don't need it, but it doesn't seem to hurt. The reformated code seems OK.

To be honest, I don't want to learn the inner architecture of `reflow.py` and neither supports a fancy fork. Nobody have time for that :)

ProGTX: Easy maintenance is always good to have :)

oddhack: The different APIs using these scripts have different conventions and they are capture in the APIConventions interface. So if you needed to, you could create syclconventions.py with the SYCL-specific things and make apiconventions.py import that, instead of vkconventions.py. However, most of the conventions are used by other scripts which you have no current use for in SYCL, so you may not even need that much - you could try replacing the code in apiconventions.py with just

> from spec_tools.conventions import ConventionsBase as APIConventions

and if that works as desired, remove vkconventions.py.

TApplencourt: Tried and got 
```
TypeError: Can't instantiate abstract class ConventionsBase with abstract methods api_name, api_prefix, extension_file_path, file_suffix, generate_structure_type_from_name, nextpoin
ter_member_name, null, structtype_member_name, xml_api_name
```

And indeed, you were "too clean", and used `abc.abstractmethod` on those methods. IMO we can merge as is. In the future, if we need a new `syclConvention` we can implement those methods (and maybe refactor the `apiconventions.py` to get a new ENV variable or something around those line)

.github/workflows/CI.yml
```diff
@@ -40,8 +40,11 @@ jobs:
         run: |
           cd adoc
           make OUTDIR=/tmp/out QUIET= html pdf
+      - name: Verify reflow conformance
+        run: |
+          ./adoc/scripts/verify_reflow_conformance.sh
```
oddhack, line 45: If I'm reading the script right, it's requiring that running reflow does nothing? That seems likely to make essentially every PR fail CI until you condition people to do this whenever they make a spec update. I've found it easier as the Vulkan spec editor to just do this when integrating PRs.

oddhack:
@TApplencourt I made a PR against this branch on your fork that just implements the required SYCL conventions for the script to succeed.

.github/workflows/CI.yml
TApplencourt: >If I'm reading the script right, it's requiring that running reflow does nothing?

Yes. If I remember correctly, it was an ask from @gmlueck .  

It looks like GitHub support issuing [warning](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message) if we want to go this way.

adoc/scripts/reflow.py
```diff
@@ -726,18 +535,27 @@ def reflowAllAdocFiles(folder_to_reflow, args):
...
                         help='Specify maximum VUID instead of the value wired into vuidCounts.py')
     parser.add_argument('-branch', action='store', dest='branch',
-                        help='Specify branch to assign VUIDs for.')
+                        help='Specify branch to assign VUIDs for')
     parser.add_argument('-noflow', action='store_true', dest='noflow',
-                        help='Do not reflow text. Other actions may apply.')
+                        help='Do not reflow text. Other actions may apply')
+    parser.add_argument('-margin', action='store', type=int, dest='margin',
+                        default='76',
+                        help='Width to reflow text, defaults to 76 characters')
```
gmlueck, line 558: It looks like commit aa7f7a41187765c6603e8569a228df0c7da810f4 did not actual change the default margin width.  I thin these two lines above are what we need to change.

gmlueck (commented):
> The reason I propose 79 instead of 80 is that sometimes 80 is a problem when there is a 80-hard limit and some editors or modes display some EOL char at the end, like my Emacs whitespace-mode to show all the weird chars.

If 79 makes life easier for Emacs users, we can use that.  This PR hasn't changed the column width yet, so it's still easy to choose a different number.  Let's try to decide today, though.

@TApplencourt, there are a number of other PRs that are ready to merge, so I propose the following order:

* We merge whatever PRs the WG approves in the meeting today.
* You merge latest SYCL-2020/master branch into this PR.
* Run reflow with whatever margin limit we decide on (79 or 80).
* I'll merge this PR as an editorial change before the next WG meeting.

adoc/scripts/reflow.py
TApplencourt: Oh, good catch. I will just do a big grep on `76`.

gmlueck:
@TApplencourt, other PRs have all been merged now, so we are ready for you to merge the latest "SYCL-2020/master" into this PR, resolve conflicts, and rerun reflow.

I suspect the merge conflicts might be substantial.  If it is easier for you, you could create a new PR with the reflow scripts and run reflow in that new PR.

However you do it, I'll do a final check of the PDF to make sure nothing gets accidentally changed before merging.

TApplencourt:
Please see  #495

PR #482, opened 2023-10-26
Use hyphens in reference names

gmlueck:
Some Asciidoc references to the glossary terms were using the name of the glossary term instead of the link name.  For glossary terms that have multiple words, this results in a reference that has a space in the name.  This space is causing problems in #481.

All these glossary terms have an anchor that is the same as the glossary term, but with a hyphen instead of the space that separates the words.  Change the references to use this anchor name instead of using the glossary term.

### DIFF
--- a/adoc/chapters/device_compiler.adoc
+++ b/adoc/chapters/device_compiler.adoc
@@ -129,8 +129,8 @@ void h() {
 }
 ----
 
-In order for the SYCL device compiler to correctly compile <<device function, device functions>>, all
-functions in the source file, whether <<device function, device functions>> or not, must be
+In order for the SYCL device compiler to correctly compile <<device-function, device functions>>, all
+functions in the source file, whether <<device-function, device functions>> or not, must be
 syntactically correct functions according to this specification. A syntactically
 correct function adheres to at least the minimum required {cpp} version
 defined in <<sec:progmodel.minimumcppversion>>.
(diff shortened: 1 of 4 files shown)


oddhack:
LGTM

TApplencourt:
LGTM (my script found the same token)

TApplencourt:
We also may want to always for a space after `,` to avoid too long line?
For example `For OpenCL <<backend>> this detail is abstracted away by <<kernel-bundle,kernel bundles>> and a `

gmlueck:
> We also may want to always for a space after `,` to avoid too long line / potential weird split? For example `For OpenCL <<backend>> this detail is abstracted away by <<kernel-bundle,kernel bundles>> and a `.

Do we need to change the spec to add a space?  Can't the script know it's safe to add a line break after this comma?

TApplencourt:
>Do we need to change the spec to add a space? Can't the script know it's safe to add a line break after this comma?

Definitely doable, but not trivial. AFAIK, Currently the script split line into word (using `split()` and then deal which those word. This is not easy extendable with `comma  separated word`. 

And it may lean to some surprising effect like
````
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa # 80 charac limits
, aaaaaaaaaaaaaaaaa
```

But we can't avoid changing the spec. It's purely a minor aestheticism change. Hence my approval :)

keryell (approved):
Thanks.

TApplencourt:
Can we merge? So I don't loos momentum working on the reflow.

PR #483, opened 2023-10-26
Fix longstanding informational messages

gmlueck:
The build of the spec has long generated the following informational messages:

```
asciidoctor: INFO: possible invalid reference: backend
asciidoctor: INFO: possible invalid reference: async-handler
```

These were caused by using a reference to a glossary term in the title of a section.  I don't see anything in Asciidoc saying that this is forbidden, so it might be a bug in Asciidoctor.  However, they don't seem that valuable, and it's helpful to have a clean build of the spec.

This commit removes these references and also changes the build process, so that an INFO message will cause a failure.  This will ensure that we don't accidentally introduce new INFO messages in the future.

### DIFF
--- a/adoc/Makefile
+++ b/adoc/Makefile
@@ -136,7 +136,7 @@ ATTRIBOPTS   = --attribute revdate="$(SPECDATE)" \
   --attribute generated=$(GENERATED) \
   --attribute refprefix \
   $(VERSIONATTRIBS)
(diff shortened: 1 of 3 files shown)


tomdeakin (approved):
This looks just editorial, so feel free to merge.

keryell (approved):
Thanks.

PR #484, opened 2023-10-30
Clarify sections previously using "synchronize"

Pennycook:
ISO C++ defines a "synchronizes-with" relationship between library calls, and defines functions in terms of their "synchronization" effects. Additionally, it refers to the new std::barrier and std::latch types as "coordination mechanisms" rather than "synchronization mechanisms".

Before this PR, several usages of "synchronize" (or "synchronization") were incompatible with these definitions, or at least potentially ambiguous. For example, several sections talked about "synchronizing" host and device data, but really meant that data would be copied from one place to another. Other sections talked about work-items or different devices "synchronizing with" one another, but without reference to specific library calls.

This PR rewords several sections, as follows:

- Where "synchronize" was used to mean "wait", the specification now says that the caller "blocks" until a certain condition holds.

- Where "synchronize" was used to mean "two copies of data are made consistent", the specification now says that data is "copied".

- Where "synchronize" was used to mean "call a barrier", the specification now talks about "coordination" of work-items and/or devices.

Any remaining occurrences of "synchronize" are related to mutexes (which ISO C++ describes as a "synchronization primitive") or to atomics and fences (which ISO C++ describes as "synchronizing with" other atomics and fences).

Closes internal issue 629.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -561,7 +561,7 @@ image::{images}/three-cg-three-queue.svg[align="center",opts="{imageopts}"]
 Submitting an action for execution returns an [code]#event# object. Programmers
 may use these events to explicitly coordinate host and device execution. Host
 code can wait for an event to complete, which will block execution on the host
-until the action(s) represented by the event has completed.  The [code]#event#
+until the action(s) represented by the event have completed.  The [code]#event#
 class is described in greater detail in <<sec:interface.event>>.
 
 Events may also be used to explicitly order the execution of kernels. Host code may
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -458,7 +458,7 @@ each other without side effects (i.e., given two requirements _r~j~_ and _r~k~_,
 _(r~j~, r~k~)_ {equiv} _(r~k~, r~j~)_). The intersection of two
 actions is not necessarily empty.
 *Actions* can include (but are not limited to): memory copy operations,
-mapping operations, host side synchronization, or implementation-specific
+mapping operations, coordinating with the host, or implementation-specific
 behavior.
 
 Finally, _Performing an action_ ({SYCLperform}(_a~i~_)) executes the
@@ -559,10 +559,10 @@ image::{images}/three-cg-three-queue.svg[align="center",opts="{imageopts}"]
 ==== Controlling execution order with events
 
 Submitting an action for execution returns an [code]#event# object. Programmers
-may use these events to explicitly synchronize programs. Host code can wait for an
-event to complete, which will block execution on the host until the action represented
-by the event has completed.  The [code]#event# class is described in greater detail
-in <<sec:interface.event>>.
+may use these events to explicitly coordinate host and device execution. Host
+code can wait for an event to complete, which will block execution on the host
+until the action represented by the event has completed.  The [code]#event#
+class is described in greater detail in <<sec:interface.event>>.
 
 Events may also be used to explicitly order the execution of kernels. Host code may
 wait for the completion of specific event, which blocks execution on the host until
@@ -624,7 +624,7 @@ Each work-item in the ND-range is identified by a value of type
 [code]#nd_item<N>#.  The type [code]#nd_item<N># encapsulates a
 global id, local id and work-group id, all of type [code]#id<N>#
 (the iteration space offset also of type [code]#id<N>#, but this is deprecated in SYCL 2020), as well as
-global and local ranges and synchronization operations necessary to
+global and local ranges and coordination mechanisms necessary to
 make work-groups useful. Work-groups are assigned ids using a similar
 approach to that used for work-item global ids.  Work-items are
 assigned to a work-group and given a local id with components in the
@@ -869,7 +869,7 @@ execute concurrently with the barrier.
 Finally, _CG~c~_ will be enqueued after _H(b1~RW~)_ is finished,
 but still has to wait for _CG~b~_ to conclude for all its requirements to
 be satisfied.
-See <<sec:synchronization>> for details on synchronization rules.
+See <<sec:coordination>> for details on host-device coordination.
 
 // Formerly in host_acc.tex
 // Uses same definitions for cga, cgb, cgc, and hostA in code and picture,
@@ -1008,11 +1008,11 @@ made consistent across the work-items in that set through the use of
 <<mem-fence>> and atomic operations.
 
 Memory consistency between the host and SYCL device(s), or different SYCL
-devices in the same context, can be guaranteed through synchronization in
-the host application as defined in <<sec:synchronization>>.  On SYCL devices
-supporting concurrent atomic accesses to USM allocations and acquire-release or sequentially
-consistent memory orderings, cross-device memory consistency can be
-enforced through the use of <<mem-fence>> and atomic operations.
+devices in the same context, can be guaranteed through library calls in the
+host application, as defined in <<sec:coordination>>. On SYCL devices
+supporting concurrent atomic accesses to USM allocations and acquire-release or
+sequentially consistent memory orderings, cross-device memory consistency can
+be enforced through the use of <<mem-fence>> and atomic operations.
 
 [[sec:memory-ordering]]
 ==== Memory ordering
@@ -1204,10 +1204,10 @@ sections of this specification.
 === Basic data parallel kernels
 
 Data-parallel <<kernel,kernels>> that execute as
-multiple <<work-item,work-items>> and where no local synchronization is required are enqueued
-with the [code]#sycl::parallel_for# function parameterized by a
-[code]#sycl::range# parameter. These kernels will execute the kernel
-function body once for each work-item in the specified <<range>>.
+multiple <<work-item,work-items>> and where no work-group-local coordination is
+required are enqueued with the [code]#sycl::parallel_for# function
+parameterized by a [code]#sycl::range# parameter. These kernels will execute
+the kernel function body once for each work-item in the specified <<range>>.
 
 Functionality tied to <<group, groups>> of work-items, including
 <<group-barrier, group barriers>> and <<local-memory>>, must not be used
@@ -1225,10 +1225,10 @@ work-group size as parameters to the [code]#sycl::parallel_for# function with a
 [code]#sycl::nd_range# parameter. In this mode of execution,
 kernels execute over the <<nd-range>> in work-groups of the specified
 size. It is possible to share data among work-items within the same
-work-group in <<local-memory,local>> or <<global-memory>> and to synchronize between
-work-items in the same work-group by calling the
-[code]#group_barrier# function. All work-groups in a given
-[code]#parallel_for# will be the same size, and the global size
+work-group in <<local-memory,local>> or <<global-memory>>, and the
+[code]#group_barrier# function can be used to block a work-item until all
+work-items in the same work-group arrive at the barrier. All work-groups in a
+given [code]#parallel_for# will be the same size, and the global size
 defined in the nd-range must either be a multiple of the work-group size in
 each dimension, or the global size must be zero. When the global size
 is zero, the kernel function is not executed, the local size is ignored, and
@@ -1243,8 +1243,8 @@ work-items in any sub-group in a kernel is based on a combination of the kernel
 and its dispatch dimensions.  The size of any sub-group in the dispatch is
 between 1 and this maximum sub-group size, and the size of an individual
 sub-group is invariant for the duration of a kernel's execution.  Similarly
-to work-groups, the work-items within the same sub-group can be synchronized
-by calling the [code]#group_barrier# function.
+to work-groups, the [code]#group_barrier# function can be used to block a
+work-item until all work-items in the same sub-group arrive at the barrier.
 
 Portable device code must not assume that work-items within a sub-group execute
 in any particular order, that work-groups are subdivided into sub-groups in a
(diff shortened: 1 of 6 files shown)


Pennycook:
A quick note to reviewers: there were also a few places where "synchronize" was used in the glossary.  I was surprised that there were such lengthy descriptions for these terms in the glossary, as they could very easily get out of sync with the main text.  Rather than rewrite these glossary descriptions to match the updated text, I simplified them and pointed back to where the relevant behaviors are specified.

keryell (approved):
Nice *coordination* with ISO C++. ;-)

TApplencourt:
For `c++` spec readers like me, can he have a little definition of "coordination" somewhere in the spec (or at least a link to the C++ spec)? 

Or maybe  something like:
```
Coordination ("synchronization") between the host and any devices can be expressed in the host SYCL
```

adoc/chapters/programming_interface.adoc
```diff
@@ -14062,7 +14063,7 @@ of 10 elements of [code]#int# type, the host pointer must at least have
 A special case is the [code]#update_host# member function.
 This member function only requires an accessor, and instructs the runtime to update
 the internal copy of the data in the host, if any. This is particularly
-useful when users use manual synchronization with host pointers, e.g.
```
TApplencourt, line 14065: why removing `host pointers`?

adoc/chapters/programming_interface.adoc
```diff
@@ -19596,15 +19596,14 @@ template <typename T> const stream& operator<<(const stream& os, const T& rhs)
...
-=== Synchronization
+=== Output
 
-An instance of the SYCL [code]#stream# class is required to synchronize with the host, and must output
-everything that is streamed to it via the [code]#operator<<()# operator before a flush operation (that
-doesn't exceed the [code]#workItemBufferSize# or [code]#totalBufferSize# limits) within a SYCL
-kernel function by the time that the event associated with a command group submission enters the completed
-state. The point at which this synchronization occurs and the member function by which this synchronization is
-performed are implementation-defined. For example it is valid for an implementation to use
-[code]#printf()#.
```
TApplencourt, line 19607: Why remove all the `printf`  and the implementation detail stuff?

Pennycook:
> For `c++` spec readers like me, can he have a little definition of "coordination" somewhere in the spec (or at least a link to the C++ spec)?
> 
> Or maybe something like:
> 
> ```
> Coordination ("synchronization") between the host and any devices can be expressed in the host SYCL
> ```

I don't think we should redefine coordination to mean synchronization.  I'm worried that will confuse people more, because they're not exactly the same.

I think this specific usage of coordination really follows the English meaning, and just means something like "getting the host and device to work together effectively".  I'm not opposed to swapping it out for a different word.  The only usage of "coordinate" that I feel strongly about keeping is the description of barriers as "a coordination mechanism".

adoc/chapters/programming_interface.adoc
Pennycook: I was trying to avoid the phrase "synchronization with host pointers".  I felt it was ambiguous whether it meant "manual synchronization [with host pointers]" or "manual [synchronization with host pointers]", and neither is good.  The former is confusing because host pointers don't provide synchronization, and the latter is potentially confusing because it's quite close to "synchronize with".

I convinced myself that the mention of "host pointers" didn't add anything to the sentence, so the simplest fix was to take it out.  If we're trying to clarify what "manual synchronization" is, I think the most important part to highlight is that we're talking about the buffer constructor which takes a mutex.  If somebody looks at the description of that constructor, they'll see how it relates to host pointers.

adoc/chapters/programming_interface.adoc
Pennycook: The sentence before this one says that:

> An instance of the SYCL `stream` class is required to output everything ... by the time that the event associated with a command group submission enters the completed state.

If we reworded the removed sentence to replace "synchronization" with "output", it would read:

> The point at which this output occurs and the member function by which this output is performed are implementation-defined.

...which just feels like repetition to me.  We've already said that output occurs at some point before the event enters the completed state.

I don't think the reference to `printf` adds anything.  (It's also confusing, because `printf` isn't a member function.  So either `printf` isn't a valid example of what the sentence before it is trying to describe, or the sentence before it is wrong.)

In hindsight, I should probably have separated these sorts of edits from the synchronization re-wording.  But it was too tempting to try and simplify things given that I was editing the sentences already.

TApplencourt:
Oh sorry, I was not clear. I'm highly supportive of following the C++ terminology! 

It's maybe just me, but reading the new `Coordination` section, it was still not clear what `coordination` meant concretely. At least synchronization, even if it's defined loosely, I understand what that means. 
`coordination` is a  CS term I was not familiar with; where `synchronization`, I kind of understand it may means`barrier` and co.

My suggestion was just to add a sentence to try to "ease" the reading to non-CS / C++ experts. But maybe it's not the goal of the specification to do so.  I will let other people chime in.

Regardless of this, It's a nice change and improves the spec clarity which is always good!

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
TApplencourt: >In hindsight, I should probably have separated these sorts of edits from the synchronization re-wording. But it was too tempting to try and simplify things given that I was editing the sentences already.

:D I know the feeling! Your justification makes sense! I just wanted to double those tangential changes :) 

Can we replace it with something like:

>The point at which the flush is performed is implementation-defined

I agree with the removal of `printf` which was unclear to me too. Hence my original residence to removing it. Removing stuff I don't understand seam... wrong)

adoc/chapters/programming_interface.adoc
TApplencourt: Agree that the term `manual synchronization` is confusing, as we don't use the term `synchronization` anymore.
If we use particularly useful, I guess we can go deep into the `particularity` and just wrote:

>This is particularly useful when users use the mutex objects on the [code]#buffer# constructors.

or something around those line

Pennycook: To make things even more confusing, I think we _could_ use "manual synchronization" here, because a mutex _is_ a synchronization primitive. :laughing:

But I'm not opposed to changing it to something like you suggested.  Maybe:

> This is particularly useful when used in conjunction with the [code]#buffer# constructor overloads which accept mutex objects.

What do you think?

Pennycook:
> Oh sorry, I was not clear. I'm highly supportive of following the C++ terminology!

No need to apologize -- you were perfectly clear!  I'm not opposed to explaining what things mean, I was just trying to point out that 'coordination ("synchronization")' might leave a reader with the impression that the terms were equivalent.


> It's maybe just me, but reading the new `Coordination` section, it was still not clear what `coordination` meant concretely. At least synchronization, even if it's defined loosely, I understand what that means. `coordination` is a CS term I was not familiar with; where `synchronization`, I kind of understand it may means`barrier` and co.

You're not alone.  I hadn't seen "coordination" used to describe barriers until I saw it in C++20.

My understanding of how things are defined in C++20 is that there's a sort of hierarchy of concepts.  A barrier provides a way to manage/organize/coordinate a set of threads.  The implementation of the barrier guarantees that all calls to a function like `barrier::arrive` _happen before_ any call unblocks from `barrier::wait`, by employing _synchronization_ operations.  (You can see an example of this [here](http://eel.is/c++draft/thread.barrier#class-14)).

I deliberately went through and tried to change every usage of "synchronization" where I wasn't 100% sure it was matching the ISO C++ definition.  If we can convince ourselves that I've been too heavy-handed here, and that some functions in SYCL _do_ provide synchronization, then I'm more than happy to revisit the text.  I expected we'd have to discuss this PR for quite some time, so please keep the questions coming!


> My suggestion was just to add a sentence to try to "ease" the reading to non-CS / C++ experts. But maybe it's not the goal of the specification to do so. I will let other people chime in.

Where I'd like us to end up eventually is even closer to ISO C++, so each function has its own _Synchronization_: section that says what synchronization operations (if any) are involved.  So, the description of `group_barrier` would have a "_Synchronization_" section saying that all the work-items must arrive at the barrier before any barrier is unblocked.  Only functions with synchronization effects would have such a section.

Would that help?  I meant to write something about this in the PR description, but forgot.

adoc/chapters/programming_interface.adoc
TApplencourt: Perfect! Far clearer. I finally understood the sentence! :D

TApplencourt:
>  Would that help? I meant to write something about this in the PR description, but forgot.

Yes! Then, perfect, we can merge this as-is and eagerly wait for the next PR :).

 I guess my comment boiled down to more of a philosophical point.  I would like to keep the spec readable for mere mortals.  

Of course, clarify / precision / unambiguous terms are the critical things in a spec! No debate about it.  But if we can avoid becoming like some other specs that are totally unreadable by non-implementers, I will appreciate it. I guess I want a "SYCL Spec for Dummies" sentences from time to time :)

adoc/chapters/programming_interface.adoc
Pennycook: New wording in 19fcedc59740213a52d6f5137103463a50c4baa3.

adoc/chapters/programming_interface.adoc
Pennycook: I've restored the note about the implementation-defined behavior in aa30f4da953f806eeee247287b65f09d4fc89084.

adoc/chapters/programming_interface.adoc
```diff
@@ -6253,7 +6254,7 @@ There are several different {cpp} classes that implement accessors:
 
 Accessor objects must always be constructed in host code, either in
 <<command-group-scope>> or in <<application-scope>>.  Whether the constructor
-blocks waiting for data to synchronize depends on the type of accessor.  Those
+blocks until data has been copied back depends on the type of accessor.  Those
```
gmlueck, line 6257:
```suggestion
blocks depends on the type of accessor.  Those
```

Can we just say this here?  I'd rather not say "copied back" because some implementations of accessors will not require memory to be copied (e.g. USM).  Note below that we say it blocks until "the data is available".

Pennycook: I have a weak preference for giving some indication of what condition will unblock the call.  Could we repeat "blocks until data is available" here?

gmlueck: Yes, that sounds good.

adoc/chapters/architecture.adoc
```diff
@@ -1316,55 +1316,54 @@ These kernels are not programmable, hence they are not bound by the SYCL
...
-    <<sycl-runtime>> ensures that extra host-based synchronization is
-    added to enable the SYCL event objects to operate between contexts
-    correctly.
+    objects which can be used to track and specify dependencies. The
+    <<sycl-runtime>> must ensure that these objects can be used to
+    enforce dependencies that span SYCL contexts from different <<backend, SYCL
+    backends>>.
+
+The operation(s) that unblock calls to these functions always happen before
+returns from any and all calls that are unblocked.
```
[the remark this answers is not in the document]
Pennycook: I added something based on your second suggestion in 66488ef.

I made a very minor tweak, so that the first occurrence of "happen before" is not in quotes, because I think we should normally use phrases like "happen before" and "synchronizes with" without quotes.  I left the quotes in the second occurrence, because it helps draw attention to a specific part of the previous sentence.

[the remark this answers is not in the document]
gmlueck: Thanks, I like it!

adoc/chapters/programming_interface.adoc
Pennycook: Fixed in 9bcbefb40be659fa1a8ce05c0551801188e0e0b9.

gmlueck (approved):

adoc/chapters/architecture.adoc
```diff
@@ -1316,55 +1316,56 @@ These kernels are not programmable, hence they are not bound by the SYCL
...
-can be specified by the user by constructing buffers with other kinds of
-references to memory, such as [code]#shared_ptr# and [code]#unique_ptr#.
+More complex forms of buffer destruction can be specified by the user by
+constructing buffers with other kinds of references to memory, such as
+[code]#shared_ptr# and [code]#unique_ptr#.
 --
-  * _Host Accessors_: The constructor for a host accessor waits for all
-    kernels that modify the same buffer (or image) in any queues to complete
+  * _Host Accessors_: The constructor for a host accessor blocks until all
+    kernels that modify the same buffer (or image) in any queues complete
```
AerialMantis, line 1344: This is not related to this PR, but this should say commands rather than kernels, as this should also apply to non-kernel invocation commands such as copies.

adoc/chapters/programming_interface.adoc
```diff
@@ -5961,7 +5962,7 @@ property::image::use_mutex
       application via a [code]#std::mutex# provided to the
       property. The [code]#std::mutex# is locked by
       the runtime whenever the data is in use and unlocked
-      otherwise. Data is synchronized with [code]#hostData#, when
+      otherwise. Data is copied back to [code]#hostData# when
```
AerialMantis, line 5965: The way this is worded suggests that data will always be copied back once the mutex is unlocked, but if the buffer would ordinarily wait on commands accessing that data on destruction it will also coordinate, and if the buffer would ordinarily not wait, then there will be no copy back.

AerialMantis (commented):
Looks good, thanks! Just a few minor comments.

Pennycook:
> Of course, clarify / precision / unambiguous terms are the critical things in a spec! No debate about it. But if we can avoid becoming like some other specs that are totally unreadable by non-implementers, I will appreciate it. I guess I want a "SYCL Spec for Dummies" sentences from time to time :)

After thinking about this some more, I've renamed the section to "Coordination _and Synchronization_", and added a paragraph that attempts to connect the two.

adoc/chapters/programming_interface.adoc
Pennycook: I don't think I understand how to fix this, primarily because I still don't think I understand how all the buffer destruction rules work.  Could you suggest alternative wording to fix this?

gmlueck: I think we could rewrite that last sentence to read:

> The contents of `hostData` are guaranteed to reflect contents of the image when the `std::mutex` is unlocked by the runtime.

Note that this paragraph describes the `use_mutex` property for images.  I think there is a similar paragraph earlier in the spec that describes the `use_mutex` property for buffers.  That should also be changed.

Pennycook: I've made both changes in d341f85357d1c102216bee74cf307dc13fdfe287.

AerialMantis: I think this wording makes sense, though I think we should also clarify that this only occurs if there is a host pointer to copy to.

Pennycook: Is there any reason to use `use_mutex` if there isn't a host pointer?  I've not used it, but my understanding from reading the spec was that it's only useful to synchronize access to `hostData`.

gmlueck:
@Pennycook: we decided in the WG that we can merge this PR once @AerialMantis approves it (indicating he is happy with your resolution of his concerns).  I will hold off merging #481 until this happens.

It looks like there are merged conflicts that need to be resolved, so please do that also @Pennycook.

adoc/chapters/architecture.adoc
```diff
@@ -559,10 +559,10 @@ image::{images}/three-cg-three-queue.svg[align="center",opts="{imageopts}"]
 ==== Controlling execution order with events
 
 Submitting an action for execution returns an [code]#event# object. Programmers
-may use these events to explicitly synchronize programs. Host code can wait for an
-event to complete, which will block execution on the host until the action represented
-by the event has completed.  The [code]#event# class is described in greater detail
-in <<sec:interface.event>>.
+may use these events to explicitly coordinate host and device execution. Host
+code can wait for an event to complete, which will block execution on the host
+until the action(s) represented by the event has completed.  The [code]#event#
```
gmlueck, line 564:
```suggestion
until the action(s) represented by the event have completed.  The [code]#event#
```

I think we normal use plural grammar when using "(s)".

Pennycook: Good catch.

adoc/chapters/programming_interface.adoc
AerialMantis: Not really, though technically it is possible to create a buffer with `use_mutex` and without a host pointer and mutex would be on any host allocation for the memory, you could also use `set_final_data` to set a host pointer.

Saying that I think it's fair to assume that it's implicit that the contents of `hostData` is only reflected if a host pointer is provided, so I think the wording is fine as it is.

AerialMantis (approved):

PR #485, opened 2023-10-31
Fix bugs in SYCL2020 USM section

mkinsner:
1. The templated aligned_alloc_host had the wrong return type, not matching the other templated signatures and clear intent of the API
1. Zero size malloc behavior was only specified for device allocations, when it needs to be specified for host and shared mallocs as well
1. Fix incorrect word use

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10570,7 +10570,7 @@ a@
 [source]
 ----
 template <typename T>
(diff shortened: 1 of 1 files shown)


gmlueck (approved):

illuhad (approved):

gmlueck:
I checked the CTS, and I didn't see any assumption that the return type of `aligned_alloc_host` is `void *`.  Therefore, I think this spec fix does not need any corresponding change to the CTS.

PR #494, opened 2023-11-10
Update contributors

gmlueck:
Update contributors to rev 8 (a little late)
Update the current spec editor


gmlueck:
Merging as editorial

PR #497, opened 2023-11-16
Improve wording around buffer mutexes

mkinsner:


adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,14 @@ if set final data is [code]#nullptr)#.
 [[subsec:mutex]]
 === Synchronization primitives
 
-When the user wants to use the [code]#buffer# simultaneously in the
-<<sycl-runtime>> and their own code (e.g. a multi-threaded mechanism) and wants
+When the user wants to use a [code]#buffer# simultaneously in the
+<<sycl-runtime>> (e.g. in a command running on a device) and their own code (e.g. a multi-threaded mechanism) and wants
```
Pennycook, line 6118: I think this is definitely better, but still not very clear.  I may be splitting hairs here, but developers can't use buffers in commands (they have to use accessors).  I'm also not sure that the multi-threaded example adds anything here, because the mutex is used to protect against race conditions between the runtime and user code rather than different threads in user code.

I'm not 100% if my suggestion below is accurate, but I think it would be clearer to write something like:
```suggestion
To prevent race conditions between accesses to the host allocation owned by a [code]#buffer# in the
<<sycl-runtime>> (e.g., by accessors) and in host code, it is necesssary
```

TApplencourt: Agree with your change! 
Do you think  we maybe use `hostData` in place of `host allocation`? 
For me `host allocation` mean `sycl::malloc_host`.  

But Thank to the `owned by a buffer`, the current sentence  is already unambiguous so not a big deal

gmlueck: Note that internal issue 426 also points out that `use_mutex` is underspecified.  If you resolve that issue with this PR, please close it.

By my reading of the spec, it would be legal for the implementation to lock the mutex when the buffer is constructed and unlock it only when the buffer is destroyed.  Of course, this wouldn't be very useful, though.

I think we either need to define some guarantees on when the mutex is unlocked by the RT, or we need to deprecate / remove this property.

Pennycook: > By my reading of the spec, it would be legal for the implementation to lock the mutex when the buffer is constructed and unlock it only when the buffer is destroyed. Of course, this wouldn't be very useful, though.

I agree with your reading. I don't think we need to define additional guarantees.

I do agree that we may want to consider deprecating/removing the property, though, if nobody understands what the feature is for.

adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,14 @@ if set final data is [code]#nullptr)#.
 [[subsec:mutex]]
 === Synchronization primitives
 
-When the user wants to use the [code]#buffer# simultaneously in the
-<<sycl-runtime>> and their own code (e.g. a multi-threaded mechanism) and wants
+To prevent race conditions between accesses to the host allocation owned by a [code]#buffer# in the
+<<sycl-runtime>> (e.g., by accessors) and in host code, it is necesssary
 to use manual synchronization without using a [code]#host_accessor#, a
```
mkinsner, line 6119:
```suggestion
to use manual synchronization though a [code]#host_accessor#, or by 
```

adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,14 @@ if set final data is [code]#nullptr)#.
...
-[code]#std::mutex# can be passed to the [code]#buffer# constructor via the right
+[code]#std::mutex# can be passed to the [code]#buffer# constructor through a
```
mkinsner, line 6120:
```suggestion
passing a [code]#std::mutex# to the [code]#buffer# constructor through a
```

adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,14 @@ if set final data is [code]#nullptr)#.
...
-[code]#std::mutex# can be passed to the [code]#buffer# constructor via the right
+To prevent race conditions between accesses to the host allocation owned by a [code]#buffer# in the
+<<sycl-runtime>> (e.g., by accessors) and in host code, it is necesssary
+to use manual synchronization though a [code]#host_accessor#, or by 
+passing a [code]#std::mutex# to the [code]#buffer# constructor through a
 [code]#property#.
 
-The runtime promises to lock the mutex whenever the data is in use and unlock it
-when it no longer needs it.
+The SYCL runtime is required to lock the mutex whenever the data is in use by the runtime, and unlock
```
mkinsner, line 6123:
```suggestion
When a [code]#buffer# was constructed with a [code]#std::mutex# property, the SYCL runtime is
required to lock the mutex whenever the data is in use by the runtime, and unlock
```

adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,15 @@ if set final data is [code]#nullptr)#.
 [[subsec:mutex]]
 === Synchronization primitives
 
-When the user wants to use the [code]#buffer# simultaneously in the
-<<sycl-runtime>> and their own code (e.g. a multi-threaded mechanism) and wants
-to use manual synchronization without using a [code]#host_accessor#, a
-[code]#std::mutex# can be passed to the [code]#buffer# constructor via the right
+To prevent race conditions between accesses to the host allocation owned by a
+[code]#buffer# in the <<sycl-runtime>> (e.g., by accessors) and in host code, it
+is necesssary to use manual synchronization though a [code]#host_accessor#, or
```
gmlueck, line 6119:
```suggestion
is necessary to use manual synchronization through a [code]#host_accessor#, or
```

mkinsner: Agh - thanks for noticing!

gmlueck (approved):

keryell (approved):
Thanks.

gmlueck:
This PR needs one more reviewer.  @TApplencourt, I see that you left comments.  If you are happy with this PR, could you approve?

gmlueck:
> I still dislike the `host allocation`. `host allocation` for me should be reserved for memory allocated via `sycl::mallloc_host` and not the regular `malloc` memory.

We should probably use the term "host USM" or "host USM allocation" when we mean memory from `sycl::malloc_host`.  This is a bigger change, though, that is outside of the scope of this PR.

adoc/chapters/programming_interface.adoc
```diff
@@ -6114,14 +6114,15 @@ if set final data is [code]#nullptr)#.
 [[subsec:mutex]]
 === Synchronization primitives
 
-When the user wants to use the [code]#buffer# simultaneously in the
-<<sycl-runtime>> and their own code (e.g. a multi-threaded mechanism) and wants
-to use manual synchronization without using a [code]#host_accessor#, a
-[code]#std::mutex# can be passed to the [code]#buffer# constructor via the right
+To prevent race conditions between accesses to the host allocation owned by a
```
mkinsner, line 6117:
```suggestion
To prevent race conditions between accesses to the host memory owned by a
```

gmlueck (approved):

gmlueck:
We agreed in the WG meeting last week that this PR can be merged once @mkinsner makes the wording change that the WG proposed.  He has now done this in 01e0449b30d8fa7c2a67b71f27de179b79b8bff9, so I think this is ready to merge.

I'll wait another day for comments.  If there are no more comments, I'll merge it.

keryell (approved):
Thanks for the even better clarification.

gmlueck:
Merging as agreed in last WG meeting.

PR #499, opened 2023-11-17
Clarify group synchronization behavior

Pennycook:
These changes build on https://github.com/KhronosGroup/SYCL-Docs/pull/484 to clarify the expected synchronization behavior of group barriers and algorithms.

The wording proposed here deliberately does not specify how a SYCL implementation must satisfy certain guarantees.  Previously, barriers were described in terms of fences and atomics, and we had discussed (offline) the possibility of defining algorithms similarly (or in terms of barriers).  The approach taken by `std::barrier` in ISO C++ is much simpler, giving implementations the freedom to use any combination of synchronization operations that guarantee certain events to happen before other events.
For example, it should be legal to implement `sycl::group_barrier` using any of the following: `std::atomic_thread_fence` + relaxed `std::atomic_ref`, release atomics paired with acquire atomics, a call to `std::barrier`, or any implementation-defined (backend-specific) functions that provide the desired memory ordering semantics.

There's a lot of duplication here now, but there are two reasons for that:

1) I wanted some feedback on the wording, and specifically whether we would like the "Effects" for each algorithm to mention the algorithm by name.

2) I wanted to separate the review of the clarifications from any later formatting changes.  The new table of overloads format that @gmlueck has been working on would allow us to combine some of these definitions.

Closes internal issue 576.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19977,6 +19977,14 @@ include::{header_dir}/groups/broadcast.h[lines=4..-1]
     trivially copyable type.
 +
 --
+_Effects:_ Blocks until all work-items in group [code]#g# have reached this
+synchronization point, then executes the broadcast operation.
+
+_Synchronization:_ The call to this function in each work-item happens before
+the broadcast operation begins execution.
+The completion of the broadcast operation happens before any work-item blocking
+on the same synchronization point is unblocked.
+
 _Returns:_ The value of [code]#x# from the work-item with the smallest linear id
 within group [code]#g#.
 --
@@ -19988,6 +19996,14 @@ within group [code]#g#.
 --
 _Preconditions:_ [code]#local_linear_id# must be the same for all work-items in
 the group and must be in the range [code]#[0, get_local_linear_range())#.
+
+_Effects:_ Blocks until all work-items in group [code]#g# have reached this
+synchronization point, then executes the broadcast operation.
+
+_Synchronization:_ The call to this function in each work-item happens before
+the broadcast operation begins execution.
+The completion of the broadcast operation happens before any work-item blocking
+on the same synchronization point is unblocked.
 
 _Returns:_ The value of [code]#x# from the work-item with the specified linear
 id within group [code]#g#.
@@ -20002,6 +20018,14 @@ _Preconditions:_ [code]#local_id# must be the same for all work-items in the
 group, and its dimensionality must match the dimensionality of the group.
 The value of [code]#local_id# in each dimension must be greater than or equal to
 0 and less than the value of [code]#get_local_range()# in the same dimension.
+
+_Effects:_ Blocks until all work-items in group [code]#g# have reached this
+synchronization point, then executes the broadcast operation.
+
+_Synchronization:_ The call to this function in each work-item happens before
+the broadcast operation begins execution.
+The completion of the broadcast operation happens before any work-item blocking
+on the same synchronization point is unblocked.
 
 _Returns:_ The value of [code]#x# from the work-item with the specified id
 within group [code]#g#.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20072,8 +20072,8 @@ with the same type and state for all work-items in group [code]#g#.
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
@@ -20091,8 +20091,8 @@ and state for all work-items in group [code]#g#.
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
@@ -20107,8 +20107,8 @@ _Returns:_ true if [code]#pred(x)# returns true for any work-item in group
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
@@ -20132,8 +20132,8 @@ with the same type and state for all work-items in group [code]#g#.
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
@@ -20151,8 +20151,8 @@ and state for all work-items in group [code]#g#.
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
@@ -20167,8 +20167,8 @@ _Returns:_ true if [code]#pred(x)# returns true for all work-items in group
 _Effects:_ Blocks until all work-items in group [code]#g# have reached this
 synchronization point, then executes the algorithm.
 
-_Synchronization:_ The call to this function in each work-item happens
-before the algorithm begins execution.
+_Synchronization:_ The call to this function in each work-item happens before
+the algorithm begins execution.
 The completion of the algorithm happens before any work-item blocking on the
 same synchronization point is unblocked.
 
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -20073,6 +20069,14 @@ _Preconditions:_ [code]#first# and [code]#last# must be the same for all
 work-items in group [code]#g#, and [code]#pred# must be an immutable callable
 with the same type and state for all work-items in group [code]#g#.
 
+_Effects:_ Blocks until all work-items in group [code]#g# have reached this
```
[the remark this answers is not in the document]
Pennycook: Good catch.

I think it's actually the barrier sentence that is wrong, so I changed that one (to remove "this work-item") in https://github.com/KhronosGroup/SYCL-Docs/pull/499/commits/d1388096494d14b8555a8a87a0e6e5ef8a9829a3.  C++ doesn't say "blocks this thread" (for example, see [here](https://eel.is/c++draft/thread.barrier.class#20)) so I think "this work-item" is implied.

adoc/chapters/programming_interface.adoc
```diff
@@ -20431,13 +20571,21 @@ the type of [code]#binary_op# must be the same for all work-items in group
...
+synchronization point, then executes the algorithm.
+
+The value written to [code]#result# + _i_ is the exclusive scan of the values
+resulting from dereferencing the first _i_ values in the range [code]#[first,
+last)# and an initial value specified by [code]#init#, using the operator
+[code]#binary_op#.
 The scan is computed using a generalized noncommutative sum as defined in
 standard {cpp}.
 
+_Synchronization:_ The call to this function in each work-item happens before
```
TApplencourt, line 20584: Should we say that we put a "barrier" before and after the execution? Or do you prefer to not rely on previous concepts? 

BTW It makes me wonder why those function doesn't take a scope...

Pennycook: > Should we say that we put a "barrier" before and after the execution? Or do you prefer to not rely on previous concepts?

Relying on previous concepts is fine, but I don't think we want to refer back to barrier here.  I'm concerned that this might be too restrictive.  Implementations shouldn't have to generate barriers when they're not needed (e.g., for back-to-back algorithms) and there may be some clever alternative implementations that don't need barriers for some algorithms (I'm not sure what they are, but I don't want to rule out the possibility).

> BTW It makes me wonder why those function doesn't take a scope...

I think this is primarily because the group algorithms were designed to align with the equivalents from the C++ standard algorithms (which don't take a scope parameter).

I agree that the fact there's no scope is inconsistent, but that might just be something we need to revisit for barriers.  I'm not sure why OpenCL's barrier supports a scope, and I'm not sure I've ever seen anybody use `group_barrier` with a non-default scope argument.

TApplencourt: >  Relying on previous concepts is fine, but I don't think we want to refer back to barrier here. I'm concerned that this might be too restrictive. Implementations shouldn't have to generate barriers when they're not needed (e.g., for back-to-back algorithms) and there may be some clever alternative implementations that don't need barriers for some algorithms (I'm not sure what they are, but I don't want to rule out the possibility).

I see, I was thinking that the phrasing is strictly equivalent to the barrier one. For example for the effect it's said:

>_Effects:_ Blocks until all work-items in group [code]#g# have reached this
synchronization point. then executes the algorithm.

Who can be verbatim be replaced by `_Effet_: Execute a barrier, then executes the algorithm.`

The different for the sync is

> The call to [code]#group_barrier# in each work-item happens before any work-item blocking on the same synchronization point is unblocked
versus
> The call to this function in each work-item happens before the algorithm begins execution.

The difference is that " any work-item blocking on the same synchronization point is unblocked" versus "algorithm begins execution".  It seem weird to me that the "effect are" the same, but the `synchronization` rules are slightly different. 

(And to be honest `The call to this function in each work item happens before the algorithm begins execution.` doesn't seem clear to me. One interpretation of this sentence is that is instruction is "volatile" and the instructions should not be re-ordened / executed out of order. This is of course *not* what is implied. But the semantic difference between this and a barrier and unclear to me) 

Also Implement should also be able to optimie 2 back-to-back calls to barrier.  IMO in the spec when we refer to `function`, we refer to their semantic not to their implementation. Implementers are free to have multiple `barrier` implementations, and implement any barrier optimization is that doesn't break the rules. 


>I agree that the fact there's no scope is inconsistent, but that might just be something we need to revisit for barriers. I'm not sure why OpenCL's barrier supports a scope, and I'm not sure I've ever seen anybody use group_barrier with a non-default scope argument.

Deal! Let remove it :D

Pennycook: > The difference is " any work-item blocking on the same synchronization point is unblocked" versus "algorithm begins execution". It seems weird to me that the effect are the same, but the synchronization rules are slightly different. The semantic difference between this and a barrier and unclear to me

The behavior of a barrier is:
- A work-item calls `group_barrier`, then waits for all the other work-items to arrive
- Once all the work-items have arrived, any work-item that is waiting can continue execution

The behavior of an algorithm is:
- A work-item calls the function, then waits for all the other work-items to arrive
- Once all the work-items have arrived, the algorithm is executed
- Once the algorithm has executed, any work-item that is waiting can continue execution

(How many work-items in the group execute the algorithm is an implementation detail, which is why I'm just saying "the algorithm is executed" instead of detailing exactly how the algorithm is divided across work-items.)

I wasn't really sure how to divide things up into _Effects_, _Synchronization_, etc, but I've tried to follow the ISO C++ definition of `std::barrier`.

`group_barrier` is like `arrive_and_wait`, which is defined [here](https://eel.is/c++draft/thread.barrier.class#23) as being an `arrive` followed by a `wait`.  `arrive` has a _Synchronization_ section [here](https://eel.is/c++draft/thread.barrier.class#14) that says the call happens before the barrier completes.  `wait` has an _Effects_ section [here](https://eel.is/c++draft/thread.barrier.class#20) that says the call blocks.  I tried to mirror that in the `group_barrier` definition -- we can't use the "equivalent to" wording, because SYCL doesn't have split barriers.

Group algorithms are like a `std::barrier` with a completion function (where the completion function is the algorithm).  So, I've tried to follow the ISO C++ definition, but substituting "algorithm" instead of "completion function".

Does that make things clearer?  There might be more from those sections that we need to copy across...

> (And to be honest The call to this function in each work item happens before the algorithm begins execution. doesn't seem clear to me. One interpretation of this sentence is that this instruction is "volatile" and all the instructions should not be re-ordened / executed out of order. This is of course not what is implied here )

The intent here is that if one work-item in the group thinks something "happened before" the algorithm starts executing, then the algorithm execution will be able to observe it.  Consider something like:

```c++
for (i = it.get_local_id(); i < N; i += it.get_local_range()) {
  input[i] = /* something */;
  output[i] = 0;
}
sycl::joint_exclusive_scan(input, input + N, output, sycl::plus<>()); 
```

The specification doesn't make any guarantees about which values of the input and output array will be assigned to which work-items.  But we need to guarantee that once all the work-items have called `joint_exclusive_scan`, whichever work-items actually perform the scan operation can "see" all the input values.  Similarly, we need to guarantee that once all the work-items have been unblocked, all work-items in the group can "see" all the output values.

i.e., all the writes to `input` happen before the algorithm starts executing, and the completion of the algorithm happens before any work-item is unblocked.

TApplencourt: Thanks a lot for the reply! 

I like the "Group algorithms are like a std::barrier with a completion function (where the completion function is the algorithm)." Clear and unambiguous. 

I  found the term " completion of the algorithm " unclear. Completion of one work item? All the work items? All the work items on the group? Completion in the sense that the instructions where issued? Or all the data need to have been "seen".

The C++ spec said, for the completion function: "The end of the completion step strongly happens before the returns from all calls that were unblocked by the completion step[.](https://eel.is/c++draft/thread.barrier.class#3.sentence-2)" 

I like the `all calls that were unblocked by the completion step`, make it super clear. 

Do you think we can do the same? Define a "general barrier" abstract concept that takes a completion function.
 
Then said that our `group_barrier` is a barrier without a completion function and that all the other algorithms are just a "general barrier" with the correct algorithm as a completion function? With that, we can potentially remove all the `_synchronization_` part for the barrier and algorithm

Pennycook: I do think that's the right long-term direction.  We have a proposed extension ([sycl_ext_oneapi_barrier](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_barrier.asciidoc)) that would introduce functions like `group_arrive` and `group_wait` as well as `barrier` objects, which would make it possible to describe `group_barrier` in terms of those operations.

I'm a little concerned about defining a barrier as an abstract concept here (in SYCL 2020) since we know that it will conflict with `std::barrier` (and potentially `sycl::barrier`) if a future version of SYCL decides to adopt C++20 as its base language.

For now, would it be sufficient to:
- Add a section clarifying what it means for an algorithm to complete
- Work in the "all calls that were unblocked by the completion step" wording
- Add a non-normative note drawing a comparison with `std::barrier` as introduced in C++20

I think it's okay to reference `std::barrier` in a non-normative note.  Making the link might help people understand what's going on, and also give some clues as to where we might go.

TApplencourt: Perfect! Thanks. Sorry to be so anal about it. Really appreciate the past PR you did, it really improved the spec IMO :)

TApplencourt: I think that we are good!  Thanks to the `The behavior of every group function is as follows:` paragraph, all my unclearness has been resolved.

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -19889,6 +19889,35 @@ SYCL provides a number of functions that expose functionality tied to groups of
...
 These group functions act as _synchronization points_ and must be encountered in
 converged <<control-flow>> by all work-items in the group.
+
+The behavior of every group function is as follows:
+
+  * Each work-item in the group arrives at the synchronization point associated
+    with the group function, then blocks until any operation(s) specified by the
+    group function have completed.
+
+  * Once all work-items in the group have arrived, an implementation-defined
```
gmlueck, line 19899:
```suggestion
  * Once all work-items in the group have arrived, an unspecified
```

Kind of a nit, but I don't think we need implementations to document this subset of work-items.

Pennycook: Good catch!

gmlueck (approved):

gmlueck:
This needs one more reviewer.  Can someone volunteer?

PR #501, opened 2023-11-22
Improve default allocator wording

mkinsner:
The spec currently implies (by my reading) that the default allocator can never fail (even on e.g. memory exhaustion).  So the spec makes  a stronger guarantee than the named allocator requirement, which probably isn't implementable.

The modified wording weakens the guarantee to allow allocations to fail.


gmlueck (approved):
Changes look good, but the CI error says that you need to run reflow.

TApplencourt (approved):

keryell (approved):
Thanks.

PR #502, opened 2023-11-23
Host allocation sharing via set final data - construction->destruction

mkinsner:
I'm not completely sure of this fix, but I think "destruction" fits better here since `set_final_data` is called post-construction.  Maybe better to say "valid at buffer destruction" instead of "after", but that matters less.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6044,7 +6044,7 @@ For example, the following could be used:
 
 However, optionally the [code]#buffer::set_final_data()# can be set to a
 [code]#std::weak_ptr# to enable copying data back, to another host memory
-address that is going to be valid after buffer construction.
+address that is going to be valid after buffer destruction.
 
 [source,,linenums]
 ----
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6044,7 +6044,7 @@ For example, the following could be used:
 
 However, optionally the [code]#buffer::set_final_data()# can be set to a
 [code]#std::weak_ptr# to enable copying data back, to another host memory
-address that is going to be valid after buffer destruction.
+address that will be valid when the buffer is destroyed.
 
 [source,,linenums]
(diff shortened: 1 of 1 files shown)


mkinsner:
The existing wording seems to reflect one very specific use case of `set_final_data`, where an initial host allocation which was bound to the buffer becomes unavailable and needs to be replaced with an alternate host allocation, during the buffer's lifetime.  There are however other use cases for `set_final_data`, and the pointer set through that API only matters across the buffer destruction boundary AFAIK.

adoc/chapters/programming_interface.adoc
```diff
@@ -6044,7 +6044,7 @@ For example, the following could be used:
 
 However, optionally the [code]#buffer::set_final_data()# can be set to a
 [code]#std::weak_ptr# to enable copying data back, to another host memory
-address that is going to be valid after buffer construction.
+address that is going to be valid after buffer destruction.
```
gmlueck, line 6047: If we want to commit this as an interim improvement, let's change this to something like:

> ... that will be valid when the buffer is destroyed.

Saying "after buffer destruction" is wrong because the pointer must be valid before the buffer destructor is invoked.

mkinsner, line 6047:
```suggestion
address that will be valid when the buffer is destroyed.
```

mkinsner: The latest commit makes this change, so I think the conversation is resolved (but @gmlueck should confirm)

gmlueck (approved):

adoc/chapters/programming_interface.adoc
gmlueck: Thanks, @mkinsner

gmlueck:
This PR needs one more reviewer.

gmlueck:
We could use one more reviewer here before today's WG meeting.

AerialMantis (approved):
Good catch, looks good.

gmlueck:
WG approved

PR #506, opened 2023-11-29
Update reqd_sub_group_size argument name.

bader:
`dim` name is tipically used to denote the dimension (see reqd_work_group_size attribute description). In this case we set the size characterisitic, rather than the dimension.

### DIFF
--- a/adoc/chapters/device_compiler.adoc
+++ b/adoc/chapters/device_compiler.adoc
@@ -569,7 +569,7 @@ a future version of the specification and is not recommended for use in new code
 a@
 [source]
 ----
-reqd_sub_group_size(dim)
+reqd_sub_group_size(size)
 ----
    a@ Indicates that the kernel must be compiled and executed with the specified
       sub-group size. The argument to the attribute must be an integral constant


Pennycook:
I think it was originally written as "dim" to match `reqd_work_group_size`.

Should we change to `reqd_work_group_size(size0, size1, size2)` as well?  Note that what the attribute really sets in that case is the size of the work-group in each dimension, so I think the same logic could apply.

bader:
> I think it was originally written as "dim" to match `reqd_work_group_size`.
> 
> Should we change to `reqd_work_group_size(size0, size1, size2)` as well? Note that what the attribute really sets in that case is the size of the work-group in each dimension, so I think the same logic could apply.

According to my understanding, the `reqd_sub_group_size` doesn't specify the size in any particular dimension. 
E.g. sub-group size `8` can be implemented as group of work-items with dimension sizes `(1, 1, 8)` or `(1, 2, 4)` or `(2, 2, 2)`.
Therefore, using `dim` for `reqd_sub_group_size` doesn't make sense.

I don't have a strong opinion about `reqd_work_group_size`, but `reqd_work_group_size(size0, size1, size2)` looks clearer to me than `reqd_work_group_size(dim0, dim1, dim2)`.

gmlueck (approved):
I agree that it makes sense to change `reqd_sub_group_size` because the parameter is not related to the dimensions of the invocation range.

I could go either way with the naming of `reqd_work_group_size`.  The parameters here describe the size of the work-group in each dimension, so the words "size" and "dim" both make sense.

TApplencourt (approved):

gmlueck:
This needs one more reviewer.  Can someone volunteer?

keryell (approved):
Yes, that looks more uniform. Thanks.

PR #511, opened 2023-11-30
Builtin function precision is backend specific

### LINKED ISSUE #507: Conflicting specification for precision of builtin functions
The SYCL 2020 specification has two conflicting statements about the precision guarantees for the builtin functions in [section 4.17](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sycl:builtins).  The introductory paragraph in that section says:

> The expected precision and any other semantic requirements are defined in the backend specification.

However, [section 4.17.4](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_math_functions) says:

> In SYCL the OpenCL math functions are available in the namespace sycl on host and device with the same precision guarantees as defined in the OpenCL 1.2 specification document [ch. 7](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#opencl12) for host and device. For a SYCL platform the numerical requirements for host need to match the numerical requirements of the OpenCL math built-in functions.

I think this statement in 4.17.4 is an oversight, and we intended to move this to the backend specification for OpenCL.

This would make sense.  Otherwise, we would require all backends to have the same precision guarantees as OpenCL (or better), which would make it quite difficult to implement SYCL on other backends.

There is a similar conflicting statement about precision in [section 4.17.6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_half_precision_math_functions):

> These functions are implemented with a minimum of 10-bits of accuracy i.e. the maximum error is less than or equal to 8192 ulp.

Again, this statement is duplicated from OpenCL, so I think we intended to move this statement to the OpenCL backend specification.

### LINKED ISSUE #509: Are builtin "native" and "half-precision" functions available in host?
While working on a PR for #507, I noticed that sections [4.17.5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_native_precision_math_functions) and [4.17.6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_half_precision_math_functions) do not specify whether the "native precision" or "half precision" math functions are available in both host and device.  This is noteworthy because the other sections describing builtin functions (4.17.4, 4.17.7 - 4.17.10) all explicitly say that their functions are available on both host and device.

My guess is that this is an oversight, and we intend these functions to be available on both host and device.  Although the host probably does not have special implementations for native or half-precisions, it's easy enough for an implementation to just use the normal `float` precision versions of these functions.  Defining these functions on both host and device makes it easier to write generic code that works both inside and outside of kernels.

The DPC++ implementation does already provide these functions on both host and device, so there be existing code that depends on this.

gmlueck:
Resolve conflicting wording about the precision requirements of the builtin math functions.  The introductory paragraph of section 4.17 "SYCL built-in functions for SYCL host and device" says that the precision is defined by the backend.  However, the descriptions in section 4.17.4 "Math functions" and 4.17.6 "Half precision math functions" specified a required precision for all backends.  Resolve this by removing the the wording in 4.17.4 and 4.17.6.  (See issue #507 for more on this.)

Also add a new section to the OpenCL backend specification stating that the precision of these builtin math functions for the OpenCL backend matches the precision of the corresponding functions in the OpenCL specification.

There are a few other related changes in this PR:

* Clarify that the native precision and half precision functions are available in both host and device code (see issue #509).

* Remove a statement that the geometric functions use round to nearest even rounding mode.  This also seems backend specific, and it doesn't even make sense for the OpenCL backend.  See OpenCL issue KhronosGroup/OpenCL-Docs#1024 for more details on this.

* Editorial changes to remove gratuitous references to OpenCL in the introductory paragraphs.

* Editorial changes to make the introductions of sections 4.17.4 - 4.17.10 read consistently.

Closes #507
Closes #509

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -23271,8 +23271,11 @@ corresponding [code]#vec#.
 that are available in the [code]#sycl::native# namespace in both host and device
 code.
 
-The range of valid input values and the maximum error for these functions is
-implementation defined.
+The precision requirements and the set of legal input values for these functions
+are defined in the backend specification.
+The intent is that these functions might make use of native device functionality
+which has better performance than their counterparts in <<sec:math-functions>>,
+but they may sacrifice accuracy or limit the set of legal input values.
 
 [[table.native.math.functions]]
 .Native precision math functions
@@ -23734,8 +23737,9 @@ are available in the [code]#sycl::half_precision# namespace in both host and
 device code.
 
 The precision requirements for these functions are defined in the backend
-specification, but the intent is that these functions have a lower precision
-than their counterparts in <<sec:math-functions>>.
+specification.
+The intent is that these functions have higher performance than their
+counterparts in <<sec:math-functions>>, but they have lower accuracy.
 
 [[table.half.math.functions]]
 .Half precision math functions
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20583,17 +20583,11 @@ determined by their linear id.
 --
 
 
+[[sec:math-functions]]
 === Math functions
 
-In SYCL the OpenCL math functions are available in the namespace [code]#sycl# on
-host and device with the same precision guarantees as defined in the OpenCL 1.2
-specification document <<opencl12, ch. 7>> for host and device.
-For a SYCL platform the numerical requirements for host need to match the
-numerical requirements of the OpenCL math built-in functions.
-
-The built-in functions available for SYCL host and device, with the same
-precision requirements for both host and device, are described in
-<<table.math.functions>>.
+<<table.math.functions>> describes the math functions that are available in the
+[code]#sycl# namespace in both host and device code.
 
 The function descriptions in this section use the term _writeable address space_
 to represent the following address spaces:
@@ -23273,10 +23267,9 @@ corresponding [code]#vec#.
 
 === Native precision math functions
 
-In SYCL the implementation-defined precision math functions are defined in the
-namespace [code]#sycl::native#.
-The functions that are available within this namespace are specified in
-<<table.native.math.functions>>.
+<<table.native.math.functions>> describes the native precision math functions
+that are available in the [code]#sycl::native# namespace in both host and device
+code.
 
 The range of valid input values and the maximum error for these functions is
 implementation defined.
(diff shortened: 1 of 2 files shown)


gmlueck:
Call for reviewers here ...

adoc/chapters/programming_interface.adoc
```diff
@@ -23273,10 +23267,9 @@ corresponding [code]#vec#.
 
 === Native precision math functions
 
-In SYCL the implementation-defined precision math functions are defined in the
-namespace [code]#sycl::native#.
-The functions that are available within this namespace are specified in
-<<table.native.math.functions>>.
+<<table.native.math.functions>> describes the native precision math functions
```
AerialMantis, line 23270: Here we describe the functions in the `native` namespace as native precision math functions, but if we are removing the precision requirement of the regular math functions, this term becomes less meaningful. Maybe we should introduce some wording which states that the regular math functions are intended to be precise, whereas the native math functions are intended to be faster at the cost of precision, even as a non-normative note?

AerialMantis (commented):
Looks good, just one comment about the meaning of the native vs regular math functions, now that we don't specify any precision requirements in the core specification.

adoc/chapters/programming_interface.adoc
gmlueck: Yes, I think that makes sense.  I can add a paragraph similar to the one I added below for the half-precision functions.  I'll have to think how to differentiate the half-precision functions from the native-precision ones, though.  They are both defined by the backend and both trade off precision for performance.

illuhad (approved):

gmlueck:
* Do native and half-precision functions guarantee the same range of input values as the normal functions?  E.g. infinity and NaN.

adoc/chapters/programming_interface.adoc
gmlueck: I addressed this in 56196d2c528cf71c86555f40b4dfd96a502757b4.

gmlueck:
> Do native and half-precision functions guarantee the same range of input values as the normal functions? E.g. infinity and NaN.

I looked at the OpenCL C specification to see how they define the half-precision functions.  The only limitation I saw on input values was for `cos`, `sin`, and `tan`.  These functions all say that the input must be in the range [-2<sup>16</sup>, +2<sup>16</sup>], and that same limitation is in the SYCL specification.  According to OpenCL C, the half-precision functions must support NaN and INF, which is the same for the regular math functions.

I don't think we need to say anything more about legal input values in the SYCL spec.

keryell (approved):
Thanks.

tomdeakin:
WG approved to merge.

PR #513, opened 2023-12-05
Fix an editorial mistake in "vec"

gmlueck:
I'm fairly certain this is an editorial mistake with the definition of `vec::operator~`.  The synopses of the `vec` class shows the correct return type, so this just updates the table to match.


TApplencourt (approved):

gmlueck:
This needs one more reviewer.

keryell (approved):
Good catch.

AerialMantis (approved):
Good catch.

PR #514, opened 2023-12-05
Clarify __swizzled_vec__ class

### LINKED ISSUE #504: Unclear specification for __swizzle_vec__ type
This question arose from a discussion with our DPC++ implementation team.  The following statement in [section 4.14.2.4](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#swizzled-vec-class) "Swizzled vec class" is unclear:

> The `__swizzled_vec__` class template should return `__swizzled_vec__&` for each operator inherited
from the `vec` class template interface which would return `vec<DataT, NumElements>&`.

Is this intended to apply to all the hidden friend functions of `vec` that are operators?  For example, consider this operator:

```
friend vec operator+(const vec& lhs, const vec& rhs);
```

Is the spec saying that `__swizzled_vec__` should define a hidden friend like this:

```
friend __swizzled_vec__ operator+(const __swizzled_vec__& lhs, const __swizzled_vec__& rhs);
```

What about other combinations of arguments, such as the ones below?  Is the spec saying they should be defined also?

```
friend __swizzled_vec__ operator+(const __swizzled_vec__& lhs, const vec<DataT, NumElements>& rhs);
friend __swizzled_vec__ operator+(const vec<DataT, NumElements>& lhs, const __swizzled_vec__& rhs);
```

In addition, it's unclear what the effect of this operator should be.  The `__swizzled_vec__` class is supposed to represent the result of a swizzle operation before the operation is actually performed.  Therefore, does this `operator+` return a `__swizzled_vec__` that also represents the addition without actually performing the addition?

This all seems like a lot of guesswork based on a terse statement in the spec.  If this is really what we intend, it seems like the spec should state this more explicitly.

gmlueck:
Prior to this commit, the `__swizzled_vec__` class was not precisely specified.  We said that it had the same interface as `vec` except for a small set of differences that we enumerated.  However, it became clear after discussing with the original authors that this wasn't a complete list of differences.

We also discovered that the `__swizzled_vec__` class should be split into two classes: `__writeable_swizzle__` and `__const_swizzle__` in order to reflect the different operations that are allowed on a swizzle of a `const vec` vs. a swizzle of a modifiable `vec`.  For example, `v.swizzle<1>()++` should not be allowed if `v` is a `const vec`.

This commit adds a complete list of the members and friend functions for the `__writeable_swizzle__` and `__const_swizzle__` types.  These are collectively called the swizzled vector types.  In many cases, the descriptions simply defer to the `vec` specification, but there are also many cases where the semantics are special for the swizzled vector types.

The updated specification also describes the swizzled vector types as a view over the original `vec` object, which I think is what we originally intended.  I intentionally removed wording like "may not be bound to a l-value or escape the expression it was constructed in."  I think the true restriction is that the lifetime of the swizzled vector type must not outlive the lifetime of the underlying `vec`, which is the case for any view.

I also intentionally removed the statement:

> Both kinds of swizzle member functions must not perform the swizzle
> operation themselves, instead the swizzle operation must be performed
> by the returned instance of `__swizzled_vec__` when used within an
> expression, meaning if the returned `__swizzled_vec__` is never used
> in an expression no swizzle operation is performed.

This wording made is sound like the point of `__swizzled_vec__` was to delay the swizzle operation as a sort of optimization.  However, our original intent was only to specify `__swizzled_vec__` as a view.

This commit also uses our updated style to specify the swizzled vector types, using a format that gives more horizontal space for the synopses and descriptions.

Closes #504


adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+
+_Effects:_ Assigns elements from the [code]#rhs# [code]#+__swizzled_vec__+#
+view to elements of "this" [code]#+__swizzled_vec__+# view.
+The value from the [code]#rhs# corresponding to the first element of the
+[code]#rhs# swizzle operation is assigned to the element of the underlying
+[code]#vec# object that corresponds to the first element of the "this" swizzle
+operation, etc.
+
+The order in which the elements are assigned is unspecified.
+Therefore, if the "this" view has any repeated elements in the swizzle, those
```
AerialMantis, line 17874: I think this wording may need to be stronger than this, the original wording did not permit swizzles on the lhs of an expression to have repeated indexes.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+
+_Returns:_ A reference to the "this" [code]#+__swizzled_vec__+# view.
+
+a@
+[frame=all,grid=none]
+!====
+a!
+[source]
+----
+__swizzled_vec__& operator=(const vec<DataT, NumElements>& rhs)
```
AerialMantis, line 17902: Could this overload be combined with the previous one with `__swizzled_vec__` like we have for the later operators?

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+
+_Constraints:_ If [code]#OP# is one of the following: [code]#%#, [code]#&#,
+[code]#|#, [code]#^#, [code]#<<#, [code]#>>#; available only when: [code]#DataT
+!= float && DataT != double && DataT != half#.
+
+In addition, overload (1) is available only when the element data type of
+[code]#rhs# is the same as [code]#DataT# and when the number of elements in the
+[code]#rhs# view is equal to [code]#NumElements#.
+
+_Effects:_ These functions behave as though the swizzle operation represented
```
AerialMantis, line 17953: I see two different wordings for describing this behavior, one describes it as performing the operation on the first element, then the second, etc, and the other describes this as evaluating into a temporary vec object and then performing the operator.

Could we use the same wording across all of these, and is it worth refactoring this wording into a separate section that we could reference from the individual member function/operator definitions?

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+
+The [code]#+__swizzled_vec__+# class is a view over a [code]#vec# object which
+reflects a swizzle operation.
+The tables below define the interface to this class, but in general it supports
+the same interface as [code]#vec#.
+Member functions and operators that read elements from the
+[code]#+__swizzled_vec__+# return the values of elements from the underlying
+[code]#vec# as if it had been modified by the swizzle operation.
+Member functions and operators that modify elements of the
+[code]#+__swizzled_vec__+# modify corresponding elements of the underlying
```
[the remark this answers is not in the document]
AerialMantis: We should also specify here that any function or operator that modifies a `_swizzled_vec__` can't have repeated indexes.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+!====
+a!
+[source]
+----
+template</*unspecified*/>                                                                 (1)
+friend vec<DataT, NumElements> operatorOP(const __swizzled_vec__& lhs,
+                                          const __swizzled_vec__</*unspecified*/>& rhs)
+
+friend vec<DataT, NumElements> operatorOP(const vec<DataT, NumElements>& lhs,             (2)
+                                          const __swizzled_vec__& rhs);
```
AerialMantis, line 17933: Some of the `__swizzled_vec__` types are missing the `</*unspecified*/>` part, is that intentional?

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+----
+!====
+Where [code]#OP# is: [code]#pass:[+]#, [code]#-#, [code]#*#, [code]#/#,
+[code]#%#, [code]#&#, [code]#|#, [code]#^#, [code]#<<#, [code]#>>#.
+
+_Constraints:_ If [code]#OP# is one of the following: [code]#%#, [code]#&#,
+[code]#|#, [code]#^#, [code]#<<#, [code]#>>#; available only when: [code]#DataT
+!= float && DataT != double && DataT != half#.
+
+In addition, overload (1) is available only when the element data type of
```
AerialMantis, line 17949: Maybe this wording could be generalized with something like:

`std::is_same_v<decltype(lhs)::element_type, decltype(rhs)::element_type> == true`

and 

`decltype(lhs)::size() == decltype(rhs)::size()`

AerialMantis (commented):
This is a really good improvement, I like using the read/modified wording as a replacement to the lhs/rhs wording, and having the `__swizzled_vec__` interface defined allows us to be more specific about some of the more nuanced behavior. I've left some suggestions. Thanks!

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
-    escape the expression it was constructed in.
-    For example [code]#auto x = f4.x()# would not be valid.
-  * The [code]#+__swizzled_vec__+# class template should return
-    [code]#+__swizzled_vec__&+# for each operator inherited from the [code]#vec#
-    class template interface which would return [code]#vec<DataT,
-    NumElements>&#.
+==== The [code]#+__swizzled_vec__+# class
+
+The [code]#+__swizzled_vec__+# class is a view over a [code]#vec# object which
+reflects a swizzle operation.
```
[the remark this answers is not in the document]
gmlueck: I addressed this and also some of your other comments on the introduction in 0271b4e1c1dad3348f526d441ff34ea1a5877aa3.

Here, I used the wording "captures the effects of a swizzle operation without actually performing that operation" instead of "lazy".  This wording also helps address your comment about clarifying the behavior when the `__swizzled_vec__` is neither read nor written.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+The tables below define the interface to this class, but in general it supports
+the same interface as [code]#vec#.
+Member functions and operators that read elements from the
+[code]#+__swizzled_vec__+# return the values of elements from the underlying
+[code]#vec# as if it had been modified by the swizzle operation.
```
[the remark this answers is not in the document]
gmlueck: I also addressed this in 0271b4e1c1dad3348f526d441ff34ea1a5877aa3.  I first introduced the concept of "capturing" a swizzle operation, then I refer to this as the "captured swizzle", which seems clearer than "swizzle operation prior".

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,500 @@ member function template.
...
+Member functions and operators that modify elements of the
+[code]#+__swizzled_vec__+# modify corresponding elements of the underlying
+[code]#vec# as translated by the swizzle operation.
```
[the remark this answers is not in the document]
gmlueck: > I think we should also add a sentence that describes the behavior of swizzle functions, where the swizzle is composed together returning a new view and no swizzle operation is performed.

This is described later as part of the description of the member functions that perform swizzles.

> And perhaps a sentence that describes the behavior of a `__swizzled_vec__` which is not read or modified, that it will fizzle and not perform any swizzle operation, this could be inferred form the other wording so maybe this could be a non-normative note.

I addressed this in 0271b4e1c1dad3348f526d441ff34ea1a5877aa3.  The first sentence says that the `__swizzled_vec__` does not perform the swizzle operation.  The next sentences say that the swizzle operation is performed only when the `__swizzled_vec__` is read or written.  Finally, I added an example showing a swizzle that is neither read nor written,  along with a comment saying that no operation is performed in this case.

adoc/chapters/programming_interface.adoc
gmlueck: I think this is a general decision that we should make about the spec.  Should constraints always use C++ syntax, or can we use words to describe the constraints when that is clear and less verbose?

I think our current direction has been to use words in situations when it is less verbose.  For example, the other constraint for these functions reads:

> available only when: `DataT != float && DataT != double && DataT != half`

which is not real C++ syntax.

adoc/chapters/programming_interface.adoc
gmlueck: I think it would end up being more verbose that way.  We can't use the simple "behaves as though the swizzle operation was first evaluated into a temporary" because these are assignment operators (see my other response).  Therefore, I think we would essentially end up with two descriptions anyways.

Since the third assignment operator (`operator=(const DataT&)`) also has its own description, it seemed better to me to have separate table entries for all three.

adoc/chapters/programming_interface.adoc
AerialMantis: So the intent of the original wording was to restrict the interface such that a swizzle with repeated elements could not be used on the lhs of an expression. So an expected implementation of this would be that the relevant operators for a swizzle with repeated elements are not available, resulting in a compile-time error for the user. I agree, this should have been more clearly specified in the wording of the operators and member functions.

The rationale for this approach was that the `vec` class was based on [Clang ext_vector_type ](https://clang.llvm.org/docs/LanguageExtensions.html#vectors-and-extended-vectors) and for this using repeated elements on the lhs of an expression results in a compile-time error (https://godbolt.org/z/dGz84s6Ge). I think we could leave the behavior unspecified, but I think it would be difficult for this to be supported consistently, one implementation could specify a priority so the result could be reasoned about and avoid UB, but other implementations may throw a compile-time error. I would prefer to not provide the interface, and avoid the user doing this by mistake.

gmlueck: OK, that makes sense.  I think we could add this as a constraint to the function specification.  It seems like this could be implemented as a compiler error using normal SFINAE techniques, and that would work for any compiler.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
AerialMantis: The new wording sounds good, thanks.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
AerialMantis: I agree it makes sense to have this defined in the description of the assignment operators.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
AerialMantis: Looks good, thanks.

adoc/chapters/programming_interface.adoc
AerialMantis: Yeah that makes sense, I agree we can use the same wording in all cases, the assignment operators do need to behave differently.

Though I wonder if we could have generic wording which applies to whether the parameters or return type is a `vec` or a `__swizzled_vec__`. I believe all binary operators where there there is an assignment, so the lhs parameter should not be treated as if being converted to a temporary `vec`, and the lhs parameter is a `__swizzled_vec__` then the return type is a `__swizzled_vec__` rather than a `vec`.

So I was thinking we could have something to the effect of:
* Any rhs parameter that is a `vec` is read element by element.
* Any rhs parameter that is a `__swizzled_vec__` is treated as if being converted to a temporary `vec`.
* If the return type is `vec` (non-assignment):
  * Any lhs parameter is treated as the rhs parameters are.
* If the return type is `__swizzled_vec__` (assignment):
  * Any lhs parameter that is a `vec` is assigned to element by element.
  * Any lhs parameter that is a `__swizzled_vec__` is assigned to element-by-element according to the swizzle index transformation.

This may be a little more complex, but we could potentially combine the wording of all binary operators.

adoc/chapters/programming_interface.adoc
AerialMantis: Yeah I think this would be a good way to do it.

adoc/chapters/programming_interface.adoc
AerialMantis: That's a very good point, I think over time I have gradually shifted towards preferring C++ syntax to describe constraints since these can be very easily translated to an SFINAE condition, and therefore reduces chances of misinterpretation. I agree, we should have a discussion about this and decide what the convention should be, it's a very good point that the constrain above is not C++ syntax, so if we were to favour that, this should really use `is_same_v` to compare types.

adoc/chapters/programming_interface.adoc
AerialMantis: That's a very good point, this is definitely the right way to do this, I missed that having an unspecified template parameter pack in multiple places could allow for those to have different template parameters which is not what we want.

adoc/chapters/programming_interface.adoc
gmlueck: Done in 9585110c44a52a9fb2f81b40652f56c4b05ecac8.

adoc/chapters/programming_interface.adoc
gmlueck: I'd like to keep the wording the way it is for now, using English words when the meaning seems clear and is less verbose.  I looked at the real C++ specification, and it seems like they also use English words for clarity.  For example, this is from C++ 27.10.8 "Exclusive scan":

> Mandates: All of
> * binary_op(init, init),
> * binary_op(init, *first), and
> * binary_op(*first, *first)
>
> are convertible to T

Notice how it uses the English words "are convertible to T" instead of using C++ syntax like `std::is_convertible`.

adoc/chapters/programming_interface.adoc
gmlueck: I think your goal is to reduce the size of the spec by combining all of the non-assignment "operatorOP" functions into a single synopsis with a common description.  Although I agree that it's good to have a shorter spec, I think this particular change would not be an improvement.  I see two problems:

* The "operatorOP" functions don't all have the same signatures.  I've already combined them as much as possible in the cases where they do have a common signature.  In order to combine them further, we'd have to somehow indicate which overloads applied to which OPs.  I think that would end up making the spec less clear.

* I'm concerned that the using general wording like what you propose above will either end up being less precise or it will be very long and complicated.  Either way, I think this will not improve the clarity of the spec.

Are you opposed to keeping the current structure?

adoc/chapters/programming_interface.adoc
AerialMantis: Thanks for adding the clarification.

adoc/chapters/programming_interface.adoc
AerialMantis: I agree, best to leave it as it is, especially if this is typically how the C++ specification does it.

adoc/chapters/programming_interface.adoc
AerialMantis: I think that's a fair concern, I was hoping that we could avoid having too much repeated wording, but I agree the alternative is to have a very long and complex general wording, so we may be exchanging verbosity for complexity, which now that I think about it some more is not necessarily helping as much as I thought.

We definitely can't combine all of the operators because as you described some have different signatures, so there would need to be some repeated wording either way.

I'm happy with keeping the wording structure as it is :)

keryell (approved):
Quite an achievement!

gmlueck:
I have removed the "draft" status from this PR.  We have this working on a branch for DPC++ now, so we are confident that the API is implementable.

adoc/chapters/programming_interface.adoc
```diff
@@ -20853,6 +21520,13 @@ to represent the following address spaces:
 * [code]#access::address_space::private_space#
 * [code]#access::address_space::generic_space#
 
+The descriptions in this section use the type name [code]#+__swizzle__+# to
```
AerialMantis, line 21523: In the case that these functions take a `__swizzle__` I don't think we specify whether the return type should also be a `__swizzle__` or if it should be a `vec`, I think in some cases the described operation may result in a `vec` but I'm not sure if that will always be the case.

If these functions are permitted to return a `__swizzle__`, should it be `__writeable_swizzle__` or `__const_swizzle__`?

adoc/chapters/programming_interface.adoc
```diff
@@ -17647,40 +17688,666 @@ member function template.
...
+
+[[sec:constructors.swizzled-vec]]
+===== Constructors for the swizzled vector class templates
+
+[frame=all,grid=none,separator="@"]
+|====
+a@
+[source]
+----
+__writeable_swizzle__() = delete
```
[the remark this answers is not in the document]
gmlueck: There is no move constructor or move-assignment operator, so the rule of 3 would apply, not the rule of 5.  The spec does define the copy constructor and copy-assignment operator, so it's only the destructor that's left out.  In general, the SYCL spec does not describe the destructor unless there is some special behavior that it has (like with the buffer destructor).  I'm not sure we want to treat the swizzle classes differently from the other classes in this regard.

adoc/chapters/programming_interface.adoc
```diff
@@ -17647,40 +17688,666 @@ member function template.
...
-    For example [code]#auto x = f4.x()# would not be valid.
-  * The [code]#+__swizzled_vec__+# class template should return
-    [code]#+__swizzled_vec__&+# for each operator inherited from the [code]#vec#
-    class template interface which would return [code]#vec<DataT,
-    NumElements>&#.
+==== The swizzled vector classes
+
+The [code]#+__writeable_swizzle__+# and [code]#+__const_swizzle__+# classes are
+each views over a [code]#vec# object which captures the effects of a swizzle
+operation without actually performing that operation.
```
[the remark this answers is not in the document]
gmlueck: There are a lot of occurrences of the phrase "the swizzle operation", and I don't think it will make the spec more clear if we change them all to be plural.

Here's my rationale for keep the singular form ... if you apply a swizzle operation to a swizzled vec object, the second swizzle operation is combined with the first swizzle operation, creating a new swizzle operation.  The new swizzled vec object represents this new swizzle operation, but it's still just one operation.

adoc/chapters/programming_interface.adoc
```diff
@@ -17625,18 +17667,17 @@ Note that the simple swizzle functions are only available for up to 4 element
...
+[code]#+__writeable_swizzle__+# or the [code]#+__const_swizzle__+# class
+template.
+These classes have an implementation-defined name, and they represent a view of
+the original [code]#vec# object with the swizzle operation applied.
+The [code]#+__writeable_swizzle__+# class represents a writeable view of the
+[code]#vec# object, while the [code]#+__const_swizzle__+# class represents a
+read-only view.
+Since the swizzle operation may result in a different number of elements, these
+views may represent a different number of elements than the original [code]#vec#
+object.
```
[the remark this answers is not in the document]
gmlueck: This is a good catch.  I think we can just add a constraint to the `vec::swizzle` function, which creates the swizzled vec object.  See fa17042f02d4b26d14796e94ec5ea71ad1dbb107.

adoc/chapters/programming_interface.adoc
gmlueck: None of these builtin functions ever return a swizzled vec object.  The normal wording for the return type is usually:

>The return type is `NonScalar` unless `NonScalar` is the `__swizzle__` type, in which case the return type is
the corresponding `vec`.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+__swizzled_vec__</*unspecified*/> odd() const
+__swizzled_vec__</*unspecified*/> even() const
+----
+!====
+_Constraints:_ These functions have the same constraints as the equivalent
+member functions of the [code]#vec# class.
+
+_Returns:_ A new [code]#+__swizzled_vec__+# view of the underlying [code]#vec#
+object, where the view represents the combination of two swizzle operations.
+The first is the swizzle operation represented by the [code]#+__swizzled_vec__+#
```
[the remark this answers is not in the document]
gmlueck: They refer to the "two swizzle operations" in the previous sentence.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+__swizzled_vec__</*unspecified*/> hi() const
+__swizzled_vec__</*unspecified*/> odd() const
+__swizzled_vec__</*unspecified*/> even() const
+----
+!====
+_Constraints:_ These functions have the same constraints as the equivalent
+member functions of the [code]#vec# class.
+
+_Returns:_ A new [code]#+__swizzled_vec__+# view of the underlying [code]#vec#
+object, where the view represents the combination of two swizzle operations.
```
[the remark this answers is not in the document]
gmlueck: Yes, I agree.  Done in ebfda4015d0f076cbfe9ef9c8176f7c612aff046.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+!====
+a!
+[source]
+----
+operator vec<DataT, NumElements>() const
+----
+!====
+_Constraints:_ Available only when [code]#NumElements > 1#.
+
+_Returns:_ A new [code]#vec# object that represents the elements of the
```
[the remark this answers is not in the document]
gmlueck: Yes, that's better.  Done in 4e94a2c8b1077a72afa8d88c14d3954e6ece4396.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+!====
+a!
+[source]
+----
+template <access::address_space AddressSpace, access::decorated IsDecorated>
+void load(size_t offset, multi_ptr<const DataT, AddressSpace, IsDecorated> ptr)
+----
+!====
+_Effects:_ Loads values from memory into elements of the underlying [code]#vec#
+object.
```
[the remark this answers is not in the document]
gmlueck: I agree this is shorter, but I think it's less clear.  I fear that if we just say "indexes are translated" then readers won't know what that translation means exactly.  I think the longer wording I have states this more clearly.

I have the same response for the next three suggestions below also.  They all shorten the wording by saying that "indexes are translated", but I think this may not be clear to all readers.

TApplencourt (approved):

AerialMantis (approved):

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
gmlueck: Destructors specified in 3514fc2ff9a0405842b9eaea526eea1727ecef3b.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
gmlueck: I'd like to keep this the way it is.  I think it would be less clear if we describe the translation in general terms.  As it is now, we can tailor the description to each operation, which I think is clearer.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+!====
+a!
+[source]
+----
+DataT& operator[](int index)
+const DataT& operator[](int index) const
+----
+!====
+_Returns:_ A reference to the element of the underlying [code]#vec# object that
+corresponds to the position [code]#index# of the swizzle operation.
```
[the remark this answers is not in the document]
AerialMantis: Resolved as above.

adoc/chapters/programming_interface.adoc
```diff
@@ -17640,40 +17635,511 @@ member function template.
...
+
+* The element data type of [code]#rhs# is the same as [code]#DataT#;
+* The number of elements in the [code]#rhs# view is equal to
+  [code]#NumElements#; and
+* The [code]#+__swizzled_vec__+# view (i.e. the left hand side of the
+  assignment) does not contain any repeated elements.
+
+_Effects:_ Assigns elements from the [code]#rhs# [code]#+__swizzled_vec__+#
+view to elements of the left hand side [code]#+__swizzled_vec__+# view.
+The value corresponding to the first element of the [code]#rhs# swizzle
```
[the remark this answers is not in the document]
AerialMantis: Resolved as above.

PR #515, opened 2023-12-07
Clarify that capture by ref host_task requires caution

hdelan:
C++ lambdas if capturing by ref need to be aware of the lifetime of the referred-to objects. This adds this information here as well. 

Motivation:

https://github.com/oneapi-src/oneDNN/pull/1767

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -14622,6 +14622,15 @@ interoperability member functions provided by the [code]#interop_handle# class.
 
 Local <<accessor,accessors>> cannot be used within a <<host-task>>.
 
+[NOTE]
+====
+If a C++ lambda is passed to a <<host-task>>, the lambda may capture by
+reference or by value.
+Since the <<host-task>> callable executes asynchronously, care must be taken to
+ensure that lifetimes of objects captured by reference by a <<host-task>> lambda
+last at least until the <<host-task>> completes.
+====
+
 // TODO: access mode/target resolution rules
 
 [source,,linenums]
--- a/adoc/chapters/acknowledgements.adoc
+++ b/adoc/chapters/acknowledgements.adoc
@@ -31,6 +31,7 @@
   * Verena Beckham, Codeplay
   * Aidan Belton, Codeplay
   * Gordon Brown, Codeplay
+  * Hugh Delaney, Codeplay
   * Morris Hafner, Codeplay
   * Alexander Johnston, Codeplay
   * Marios Katsigiannis, Codeplay


adoc/chapters/programming_interface.adoc
```diff
@@ -14622,6 +14622,13 @@ interoperability member functions provided by the [code]#interop_handle# class.
 
 Local <<accessor,accessors>> cannot be used within a <<host-task>>.
 
+If a C++ lambda is passed to a <<host-task>>, the lambda may capture by
+reference or by value. If a <<host-task>> lambda captures by reference, the user
+must ensure that referred-to user defined classes use appropriate reference
+counting. Since the <<host_task>> callable may execute at an unspecified time
+between a command submission and a synchronization point, its execution may
+outlive the referred-to objects if reference counting is not used.
```
gmlueck, line 14630: This seems too specific to a particular bug that you had, and the recommendation about reference counting is only one possible solution.  I don't think we need to add any new formal requirements to the spec here.  We already say that a host task's execution is deferred and that it executes asynchronously.  A logical outcome of this is that you need to take care when using pointers or references inside the host task.

If we want to add something to the spec, I'd suggest a non-normative note like this:

```
[NOTE]
====
Care must be taken when using pointers or references inside a host task when the pointer
or reference refers to an object that is defined outside of the host task.  Because the host
task is executed asynchronously, applications should ensure that the referenced object's
lifetime lasts at least until the host task completes.
====
```

hdelan: Thanks for comments. Have rephrased to advise caution

hdelan:
@keryell I agree. I think it is worth reiterating here, but can close this PR if you think it doesn't belong in the spec

AerialMantis (approved):

nliber (approved):

hdelan:
> Hi @hdelan, we have some new formatting requirements for the *.adoc files, and these are checked by our CI system. One of these checks failed for this PR. Can you do the following:
> 
>     * Run: `./adoc/scripts/reflow.py -overwrite adoc/chapters/programming_interface.adoc`
> 
>     * This will make some formatting changes to "programming_interface.adoc".  Check that nothing unexpected happened.
> 
>     * Commit the new changes and push the commit to this PR.
> 
> 
> Thanks.

Thanks @gmlueck have done that

gmlueck (approved):

hdelan:
@gmlueck should I add my name to `acknowledgements.adoc`?

gmlueck:
> @gmlueck should I add my name to `acknowledgements.adoc`?

Yes, if you like.

hdelan:
@gmlueck thanks, done

adoc/chapters/programming_interface.adoc
```diff
@@ -14622,6 +14622,15 @@ interoperability member functions provided by the [code]#interop_handle# class.
 
 Local <<accessor,accessors>> cannot be used within a <<host-task>>.
 
+[NOTE]
+====
+If a C++ lambda is passed to a <<host-task>>, the lambda may capture by
+reference or by value.
+Since the <<host-task>> callable executes asynchronously, care must be taken to
+ensure that lifetimes of objects captured by reference by a <<host-task>> lambda
```
keryell, line 14630: Even for objects captured by copy you still need to unsure that it works under the hood with the program logic.
The trivial example is a `std::ref` or a pointer captured by copy. ;-)

gmlueck:
WG approved

PR #517, opened 2023-12-09
Fix comments for nd_item interface.

bader:
This is an editorial change.


gmlueck:
Merging as editorial.

PR #519, opened 2023-12-12
Use fixed-width type in nan()

gmlueck:
We noticed during implementation of the `nan` builtin function that the specification did not ensure the `nancode` parameter had enough bits to cover the entire range of the FP significand.  In particular, `unsigned long` is not guaranteed to be wide enough for the 52 bits of significand that are present in `double`.

It seems better to use the fixed-width integer types in this API, which is what this commit proposes.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -22680,9 +22680,9 @@ a@
 a!
 [source]
 ----
-float nan(unsigned int nancode)         (1)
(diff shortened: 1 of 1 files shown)


AerialMantis (approved):
Good catch, I think this was missed when translating the builtins from OpenCL, where `unsigned short`, `unsigned int` and `unsigned long` were fixed-sized types.

TApplencourt (approved):

keryell (approved):
Great!

PR #520, opened 2023-12-18
Add clarification to event wait wording

AerialMantis:
**Motivation**

It was discovered that the current wording for `event::wait_and_throw` was ambiguous.

_Table 34_
> Wait for the event and the command associated with it to complete.
> 
> Any unconsumed async-error from any context that the event was waiting on executions from will be passed to the async_handler associated with the context. If no user defined async_handler is associated with the context, then an implementation-defined default async_handler is called to handle any errors, as described in [Section 4.13.1.2].

The above wording suggests that only errors produced by commands associated with this event are waiting on are passed to the async handler, however, the use of the phrase "any context that the event was waiting on executions from" suggests that the intention was to pass errors produced by commands of this event and any dependent events.

**Summary**

In this patch, I have modified the wording of the `event::wait_and_throw()` member functions to clarify that they will throw asynchronous errors for this event and all dependent events to their respective async handlers.

While changing this I have also modified the wording around the waiting behavior for both the `event::wait` and `event::wait_and_throw` member functions to match that which is used in the ISO C++ specification; to state that the function blocks until all commands have been completed. This wording could be extended to also include a synchronization clause, however, since these functions are still using the original format, I've just kept this to the effects clause for now.

Finally, I have modified the wording of the static `event::wait()` and `event::wait_and_throw()` functions to be based on the definitions in member functions to avoid duplication of wording.

**Target**

This patch is targeting SYCL 2020 as a clarification.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3838,7 +3838,7 @@ void wait_and_throw()
 Any uncomsumed <<async-error,asynchronous errors>> held by queues (or their
 associated contexts) which were used to enqueue commands associated with this
 event and any dependent events, are passed to the approriate <<async-handler>>
-as described in <<async.handler.priorities>>.
+as described in <<subsubsec:async.handler.priorities>>.
 
 [NOTE]
 ====
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3835,13 +3835,16 @@ void wait_and_throw()
    a@ Blocks until all commands associated with this event and any dependent
       events have completed.
 
-Any unconsumed <<async-error,asynchronous errors>> which were produced by the
-execution of commands associated with this event or any dependent events, are
-passed to the <<async-handler>> associated with the respective context.
+Any uncomsumed <<async-error,asynchronous errors>> held by queues (or their
+associated contexts) which were used to enqueue commands associated with this
+event and any dependent events, are passed to the approriate <<async-handler>>
+as described in <<async.handler.priorities>>.
 
-If no user defined <<async-handler>> is associated with the context, then an
-implementation-defined default <<async-handler>> is invoked to handle errors,
-as described in <<subsubsec:exception.nohandler>>.
+[NOTE]
+====
+This behaviour is equivalent to calling [code]#queue::throw_asynchronous()# on
+the queue associated with this event and any dependent events.
+====
 
 a@
 [source]
@@ -16471,6 +16474,7 @@ defined in <<subsubsec:exception.async>>.
 The default <<async-handler>> must in some way report all errors passed to it,
 when possible, and must then invoke [code]#std::terminate# or equivalent.
 
+[[subsubsec:async.handler.priorities]]
 ==== Priorities of async handlers
 
 If the SYCL runtime can associate an <<async-error>> with a specific queue,
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3835,10 +3835,10 @@ void wait_and_throw()
    a@ Blocks until all commands associated with this event and any dependent
       events have completed.
 
-Any uncomsumed <<async-error,asynchronous errors>> held by queues (or their
-associated contexts) which were used to enqueue commands associated with this
-event and any dependent events, are passed to the approriate <<async-handler>>
-as described in <<subsubsec:async.handler.priorities>>.
+At least all uncomsumed <<async-error,asynchronous errors>> held by queues (or
+their associated contexts) which were used to enqueue commands associated with
+this event and any dependent events, are passed to the approriate
+<<async-handler>> as described in <<subsubsec:async.handler.priorities>>.
 
 [NOTE]
 ====
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3824,46 +3824,40 @@ a@
 ----
 void wait()
 ----
-   a@ Wait for the event and the command associated with
-      it to complete.
+   a@ Blocks until all commands associated with this event and any dependent
+      events have completed.
 
 a@
 [source]
 ----
 void wait_and_throw()
 ----
-   a@ Wait for the event and the command associated with
-      it to complete.
+   a@ Blocks until all commands associated with this event and any dependent
+      events have completed.
 
-Any unconsumed <<async-error,asynchronous errors>> from any context that the event
-was waiting on executions from will be passed to the
-<<async-handler>> associated with the context.
-If no user defined [code]#async_handler# is associated with
-the context, then an implementation-defined
-default <<async-handler>> is called to handle any errors, as
-described in <<subsubsec:exception.nohandler>>.
+Any unconsumed <<async-error,asynchronous errors>> which were produced by the
+execution of commands associated with this event or any dependent events, are
+passed to the <<async-handler>> associated with the respective context.
+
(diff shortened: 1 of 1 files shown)


TApplencourt:
Look good to me! 

I don't especially like the addition of `dependent events`. It seems "obvious/implied " to me, so I don't see why we put it here. But if it's what ISO C++ is doing then it's a good argument to do the same :)

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -3824,46 +3824,40 @@ a@
...
-the context, then an implementation-defined
-default <<async-handler>> is called to handle any errors, as
-described in <<subsubsec:exception.nohandler>>.
+Any unconsumed <<async-error,asynchronous errors>> which were produced by the
+execution of commands associated with this event or any dependent events, are
+passed to the <<async-handler>> associated with the respective context.
+
+If no user defined <<async-handler>> is associated with the context, then an
+implementation-defined default <<async-handler>> is invoked to handle errors,
+as described in <<subsubsec:exception.nohandler>>.
```
gmlueck, line 3844: I think this change is an improvement, but there are a couple other things that we might want to improve also.

The wording about "associated with the respective context" is not quite accurate because the async errors could also be reported to the handler that is associated with the queue.  In fact, [section 4.13.1.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_priorities_of_async_handlers) states that the handler associated with the queue has precedence over the handler associated with the context.  Perhaps we could say something like this instead:

> ... are passed to the application's async-handler as described in section 4.13.1.3.

I also wonder if the part about "which were produced by the execution of commands associated with this event or any dependent events" is accurate.  [Section 4.13.1.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsubsec:exception.async) says only that "any unconsumed asynchronous errors associated with the queue or context" are reported.  It's not clear to me what precisely we intended for a situation like this:

```
sycl::queue q{async_handler};
auto e1 = q.submit(/*....*/);
auto e2 = q.submit(/*...*/);  // Does *not* depend on e1
q.wait();  // Does not report async errors

e2.wait_and_throw();
```

Let's suppose the command associated with `e1` generated an async error.  Will this error be reported when `e2.wait_and_throw` is called?  I think there are three reasonable answers:

1. The spec requires that the error is *not* reported because the error is not generated by the command associated with `e2` or by any of its dependents.
2. The spec allows the error to be reported, but does not require it.
3. The spec requires that the error *is* reported because the call to `q.wait` ensures that the async error was generated, so it becomes an "unconsumed error" on the queue, and the call to `e2.wait_and_throw` is guaranteed to report any unconsumed errors.

gmlueck: In the WG meeting today, the WG seems to be in favor of option (2).  We think the spec wording should be clarified to indicate this.

keryell (approved):
I would prefer the wording `wait_or_throw()`. :-)

tomdeakin:
Waiting for changes according to comment https://github.com/KhronosGroup/SYCL-Docs/pull/520#discussion_r1442020695

etomzak:
What surprises me is that `event::wait_and_throw()` exists at all. An event is a synchronization point. It makes sense to talk about a queue or a context owning some asynchronous exceptions; I don't think it makes sense to talk about an event owning asynchronous exceptions. IMO `event::wait_and_throw()` awkwardly mixes two different abstractions.

Options 1, 2, and 3 all look problematic to me, because they assume that async exceptions can be traced on the level of granularity of individual commands. One of the consequences of an asynchronous API (which, in some cases, is also allowed to reorder commands) is that that level of tracing is not possible. This doesn't have to be a bad thing. I'd argue that the minimum sensible level of granularity of async exception tracing is the queue.

So I'll suggest two more possible solutions:

4. Deprecate `event::wait_and_throw()`
5. Make `event::wait_and_throw()` equivalent to `queue::wait_and_throw()` for all queue(s) that the event is associated with

AerialMantis:
> What surprises me is that `event::wait_and_throw()` exists at all. An event is a synchronization point. It makes sense to talk about a queue or a context owning some asynchronous exceptions; I don't think it makes sense to talk about an event owning asynchronous exceptions. IMO `event::wait_and_throw()` awkwardly mixes two different abstractions.
> 
> Options 1, 2, and 3 all look problematic to me, because they assume that async exceptions can be traced on the level of granularity of individual commands. One of the consequences of an asynchronous API (which, in some cases, is also allowed to reorder commands) is that that level of tracing is not possible. This doesn't have to be a bad thing. I'd argue that the minimum sensible level of granularity of async exception tracing is the queue.
> 
> So I'll suggest two more possible solutions:
> 
> 4. Deprecate `event::wait_and_throw()`
> 5. Make `event::wait_and_throw()` equivalent to `queue::wait_and_throw()` for all queue(s) that the event is associated with

@etomzak I have responded to this comment in reply to the discussion above.

AerialMantis:
> Look good to me!
> 
> I don't especially like the addition of `dependent events` for the definition of `wait` and `wait and throw`. It seems "obvious/implied " to me, so I don't see why we put it here. But if it's what ISO C++ is doing then it's a good argument to do the same :)

I agree, this was the original intent so it was implied, but it was raised by users that this was ambiguous so we thought we should clarify the wording to make sure it's clear.

tomdeakin:
Still waiting on @AerialMantis to update.

adoc/chapters/programming_interface.adoc
AerialMantis: These clarifications have been addressed in https://github.com/KhronosGroup/SYCL-Docs/pull/520/commits/3df6485b1248220252663a2a775517b247169147.

AerialMantis:
I have updated this PR to reflect the clarifications we discussed (https://github.com/KhronosGroup/SYCL-Docs/pull/520/commits/3df6485b1248220252663a2a775517b247169147).

gmlueck (commented):
The new wording looks good to me, but there is an error with the cross reference which causes CI to fail.

gmlueck:
Summarizing a long WG discussion:

* This PR should loosen the language to say that other async errors could also be reported from `wait_and_throw`, even if those errors are not directly related to the event or queue being waited upon.
* Otherwise, an implementation must track which command is associated with each async error.
* In the future, we need a longer discussion on async error handling.  We do not have good agreement on what the current APIs mean, or what the use case is.
* Adding loose language now that allows an implementation to report extra errors during `wait_and_throw` does not limit us later.  A future spec release can tighten the langue and limit which errors can be returned.  This will not be an API break.

adoc/chapters/programming_interface.adoc
```diff
@@ -3824,46 +3824,43 @@ a@
...
-was waiting on executions from will be passed to the
-<<async-handler>> associated with the context.
-If no user defined [code]#async_handler# is associated with
-the context, then an implementation-defined
-default <<async-handler>> is called to handle any errors, as
-described in <<subsubsec:exception.nohandler>>.
+Any uncomsumed <<async-error,asynchronous errors>> held by queues (or their
+associated contexts) which were used to enqueue commands associated with this
+event and any dependent events, are passed to the approriate <<async-handler>>
+as described in <<async.handler.priorities>>.
```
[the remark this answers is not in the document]
AerialMantis: Good catch, thanks.

gmlueck (approved):

AerialMantis:
I've updated the wording to address the feedback from the recent WG discussion, `event::wait_and_throw` now allows an implementation to also throw async errors from queues/contexts not used to enqueue the commands the events is waiting on.

PR #522, opened 2023-12-20
Editorial fix in vector section

mkinsner:
This fix makes the wording better, but I'm actually not sure what it means to have compatible support on the host when device support isn't possible.  Should the later part of this sentence be dropped totally, since I don't think there is alternate host emulation?  There is independently compatible support for the vector types on the host, unrelated to what the device supports, AFAICT.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -16936,9 +16936,6 @@ This type allows sharing of vectors between the host and its SYCL devices.
 The vector supports member functions that allow construction of a new vector
 from a swizzled set of component elements.
 
-[code]#vec<typename _DataT_, int _NumElements_># is a vector type that compiles
-down to a <<backend>> built-in vector type on SYCL devices, where possible, and
-provides compatible support on the host when not possible.
 The [code]#vec# class is templated on its number of elements and its element
 type.
 The number of elements parameter, _NumElements_, can be one of: 1, 2, 3, 4, 8 or 16.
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -16937,8 +16937,8 @@ The vector supports member functions that allow construction of a new vector
 from a swizzled set of component elements.
 
 [code]#vec<typename _DataT_, int _NumElements_># is a vector type that compiles
-down to a <<backend>> built-in vector types on SYCL devices, where possible, and
-provides compatible support on the host or when it is not possible.
+down to a <<backend>> built-in vector type on SYCL devices, where possible, and
+provides compatible support on the host when not possible.
```
gmlueck, line 16941: This entire sentence seems like an implementation detail to me.  How could you write a test for this sentence?  It just says that it might or might not use a built-in vector type in device code.

The part about "provides compatible support on the host when not possible" seems not quite accurate.  The `vec` type works on the host regardless of how `vec` is implemented in device code.  So, I think this part of the sentence is just saying that `vec` is also available in host code.  We say that already in the first sentence of the first paragraph in this section.

I think we should just delete the sentence, rather than trying to fix it.

gmlueck: Thanks.  Can you update the description of this PR also?

gmlueck (approved):

gmlueck:
It would be nice to have a couple other company reviewers on this small PR.

nliber (approved):

illuhad (approved):

tomdeakin:
WG approves merge

PR #534, opened 2024-02-02
Remove unusable asciidoc attribute

etomzak:
I *think* the intent here is to provide a plural alias for the `work-item` glossary term. Since this attribute is defined at the very end of the spec, though, it's not used anywhere and not usable anywhere.

If this alias is needed, then it needs to be defined somewhere like `adoc/config/attribs.adoc`.


keryell (approved):
Thanks.
But as you suggest, perhaps a solution is to advertise more `adoc/config/attribs.adoc` and its neighborhood and put this definition there instead of being at the beginning of `adoc/syclbase.adoc`?

gmlueck (approved):

gmlueck:
Merging this as editorial.

PR #537, opened 2024-03-04
Clarify access::target::host_task

hdelan:
An accessor with `access::target::host_task` is accessible on the host from within the host task lambda. 

Specify that `access::target::host_task` means not only 'accessible within host_task' but 'accessible within host_task from the host'. This is necessary as `host_task` defines a scope, but it doesn't define where the memory is accessible from.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6511,15 +6511,15 @@ access an element produces undefined behaviour.
 
 ==== Buffer accessor for commands
 
-The [code]#accessor# class provides access to data in a [code]#buffer# from
-within a <<sycl-kernel-function>> or from within a <<host-task>>.
-When used in a <<sycl-kernel-function>>, it accesses the contents of the buffer
-via the device's <<global-memory>>.
-These two forms of the accessor are distinguished by the [code]#AccessTarget#
-template parameter as shown in <<table.accessors.command.buffer.capabilities>>.
-Both forms support the following values for the [code]#AccessMode# template
-parameter: [code]#access_mode::read#, [code]#access_mode::write# and
-[code]#access_mode::read_write#.
+The [code]#accessor# class provides access to data in a [code]#buffer# in three
+different ways.
+It can be used to access the buffer's data from within a
+<<sycl-kernel-function>> via the device's <<global-memory>>.
+It can also be used to access the buffer's data on host from within a host task.
+Finally, it can be used to get a native backend handle to the buffer from within
+a host task.
+The AccessTarget template parameter helps distinguish these three cases as shown
+in <<table.accessors.command.buffer.capabilities>>.
 
 [[table.accessors.command.buffer.capabilities]]
 .Description of access targets for buffer accessors
@@ -6528,18 +6528,27 @@ parameter: [code]#access_mode::read#, [code]#access_mode::write# and
 | Access target | Meaning
 | [code]#target::device#
     | Access a buffer from a <<sycl-kernel-function>> via device global memory.
+      Also used to get a native backend handle to the buffer from within a host
+      task.
 | [code]#target::host_task#
-    | Access a buffer from a <<host-task>>.
+    | Access a buffer's data on host from within a <<host-task>>.
 |====
 
-Programs which specify the access target as [code]#target::device# and then
-capture the [code]#accessor# in a <<host-task>> can only use the accessor for
-interoperability through the [code]#interop_handle#, any other uses result in
-undefined behavior.
+When an [code]#accessor# is used from within a <<sycl-kernel-function>>, the
+access target must be [code]#target::device#, [code]#target::constant_buffer#,
+or [code]#target::local#; otherwise the behavior is undefined.
+See <<sec:accessor.deprecated.target.constant.buffer>> and
+<<sec:accessor.deprecated.target.local>> for a description of the deprecated
+[code]#target::constant_buffer# and [code]#target::local# targets.
 
-Programs which specify the access target as [code]#target::host_task# and then
-use the [code]#accessor# from a <<sycl-kernel-function>> result in undefined
-behavior.
+When an [code]#accessor# is used from within a host task, the use of the
+accessor must correspond to the access target, otherwise the behavior is
+undefined.
+If the access target is [code]#target::host_task#, the accessor may only be used
+to access the buffer's data on host, from within the host task function.
+If the access target is [code]#target::device#, the accessor may only be used to
+get a native backend handle for the buffer as described in
+<<subsec:interfaces.hosttasks.interophandle>>.
 
 The dimensionality of the accessor must match the underlying buffer, however,
 there is a special case if the buffer is one-dimensional.
@@ -7119,6 +7128,7 @@ size_t get_count() const
 |====
 
 
+[[sec:accessor.deprecated.target.constant.buffer]]
 ====== Accessor specialization with [code]#target::constant_buffer#
 
 The [code]#accessor# class may be specialized with target
@@ -7489,6 +7499,7 @@ std::add_pointer_t<value_type> get_pointer() const noexcept
 |====
 
 
+[[sec:accessor.deprecated.target.local]]
 ====== Accessor specialization with [code]#target::local#
 
 The [code]#accessor# class may be specialized with target [code]#target::local#,


keryell (approved):
Thanks for the clarification.

adoc/chapters/programming_interface.adoc
```diff
@@ -6537,6 +6537,8 @@ capture the [code]#accessor# in a <<host-task>> can only use the accessor for
 interoperability through the [code]#interop_handle#, any other uses result in
 undefined behavior.
 
+An accessor with access target [code]#target::host_task# is accessible on the
+host, within the host task function for which the accessor is declared.
```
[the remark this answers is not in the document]
hdelan: An accessor in a HT can be used to access device memory (if constructed with `access::target::device`), or it can be used to access host side memory (if constructed with `access::target::host_task`).

But `access::target::host_task` doesn't tell the user where the memory is accessible from, just that it is accessible within the scope of the HT function. Since we can have both device and host accessors within HTs we need to be explicit that the `access::target::host_task` accessor should only be used on host, and not device through some interop method.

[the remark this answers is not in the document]
gmlueck: I see.  I think we can convey this more clearly with a larger change.  I think the problem is that we initially wrote this section without thinking about the interop case, and then we added that in as an afterthought.  It would probably have been better to introduce all three uses up front in the introductory paragraph of this section like so:

> The `accessor` class can be used to access a `buffer` in three different ways.  It can be used to access the buffer's data from within a SYCL kernel function via the device's global memory.  It can also be used to access the buffer's data from within a host task.  Finally, it can be used to get a native backend handle to the buffer from within a host task.  The `AccessTarget` template parameter helps distinguish these three cases as shown in Table 54.
>
> [Table 54]
> `target::device`: Access a buffer's data from within a SYCL kernel function.  Also used to get a native backend handle to the buffer from within a host task.
> `target::host_task`: Access a buffer's data from within a host task.
>
> When an `accessor` is used from within a kernel function, the access target must be `target::device`, otherwise the behavior is undefined.
>
> When an `accessor` is used from within a host task, the use of the accessor must correspond to the access target, otherwise the behavior is undefined.  If the access target is `target::host_task`, the accessor may only be used to access the buffer's data.  If the access target is `target::device`, the accessor may only be used to get a native backend handle for the buffer as described in Section 4.10.2.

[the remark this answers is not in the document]
hdelan:
```quote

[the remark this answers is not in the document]
Comment (hdelan) [reply adoc/chapters/programming_interface.adoc:6541]: Sounds good but I think worth also adding "on host" in a few places

> ... It can also be used to access the buffer's data ***on host*** from within a host task.  Finally, it can be used to get a native backend handle to the buffer from within a host task.  The `AccessTarget` template parameter helps distinguish these three cases as shown in Table 54.
> [Table 54]
> `target::device`: Access a buffer's data from within a SYCL kernel function.  Also used to get a native backend handle to the buffer from within a host task.
> `target::host_task`: Access a buffer's data ***on host*** from within a host task.

> ... If the access target is `target::host_task`, the accessor may only be used to access the buffer's data ***on host***.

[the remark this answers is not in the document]
Comment (gmlueck) [reply adoc/chapters/programming_interface.adoc:6541]: Yes, that seems like a good improvement.  Thanks.

File: adoc/chapters/programming_interface.adoc [hunk 2]
```diff
@@ -6511,15 +6511,15 @@ access an element produces undefined behaviour.
...
-Both forms support the following values for the [code]#AccessMode# template
-parameter: [code]#access_mode::read#, [code]#access_mode::write# and
-[code]#access_mode::read_write#.
+The [code]#accessor# class provides access to data in a [code]#buffer# in three
+different ways.
+It can be used to access the buffer's data from within a
+<<sycl-kernel-function>> via the device's <<global-memory>>.
+It can also be used to access the buffer's data on host from within a host task.
+Finally, it can be used to get a native backend handle to the buffer from within
+a host task, to be used on device.
```
gmlueck, line 6520: The phrase "to be used on device" seems confusing to me.  The native backend handle is used on the host, not on the device.  The handle may refer to data that lives on the device, but the handle itself is used on the host.  I think we could just remove this phrase (and also in the table describing `target::device` which has the same phrase).  Later on, you have this text which clarifies exactly how the handle can be used:

> the accessor may only be used to get a native backend handle for the buffer as described in <<subsec:interfaces.hosttasks.interophandle>>

hdelan: Thanks have rectified

adoc/chapters/programming_interface.adoc
```diff
@@ -6528,18 +6528,20 @@ parameter: [code]#access_mode::read#, [code]#access_mode::write# and
...
 
-Programs which specify the access target as [code]#target::device# and then
-capture the [code]#accessor# in a <<host-task>> can only use the accessor for
-interoperability through the [code]#interop_handle#, any other uses result in
-undefined behavior.
-
-Programs which specify the access target as [code]#target::host_task# and then
-use the [code]#accessor# from a <<sycl-kernel-function>> result in undefined
-behavior.
+When an [code]#accessor# is used from within a host task, the use of the
```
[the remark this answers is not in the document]
hdelan: Thanks have added back in.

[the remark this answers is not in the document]
gmlueck: Thanks.  I think it would help to but a paragraph after that first sentence, though.  The first sentence describes limitations when using `accessor` in a kernel function, while the next three sentences describe limitations when using `accessor` in a host task.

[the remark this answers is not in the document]
hdelan: Thanks have changed

gmlueck (approved):

hdelan:
Thanks @gmlueck . Would you be able to merge this?

hdelan:
Ok sure thanks

adoc/chapters/programming_interface.adoc
```diff
@@ -6528,18 +6528,24 @@ parameter: [code]#access_mode::read#, [code]#access_mode::write# and
...
+    | Access a buffer's data on host from within a <<host-task>>.
 |====
 
-Programs which specify the access target as [code]#target::device# and then
-capture the [code]#accessor# in a <<host-task>> can only use the accessor for
-interoperability through the [code]#interop_handle#, any other uses result in
-undefined behavior.
+When an [code]#accessor# is used from within a <<sycl-kernel-function>>, the
+access target must be [code]#target::device#, otherwise the behavior is
+undefined.
```
tomdeakin, line 6539: Should also be defined for `target::constant_buffer`.

gmlueck: @hdelan: The WG discussed this today, and they want to list the deprecated access targets here also.  The WG mentioned `target::constant_buffer`, but there is also `target::local`.  I'd suggest saying something like:

> When an [code]#accessor# is used from within a \<\<sycl-kernel-function\>\>, the
access target must be [code]#target::device#, [code]#target::constant_buffer#, or [code]#target::local#; otherwise the behavior is undefined.  See sections XXX and YYY for a description of the deprecated [code]#target::constant_buffer# and [code]#target::local# targets.

hdelan: Thanks. Have added this in latest commit.

gmlueck: Looks good to me.  Thanks!

hdelan: Great I will squash commits now

gmlueck: I think there's no need.  We generally leave commits as they are in the PR.  When we merge a PR to the repo, we do a "merge commit", not "squash and merge".

keryell: The commits do still exist somewhere, so it is always possible to change your mind. Either by looking at the commit hashes here or running `git reflog` on your side.

keryell (approved):
Thanks.

adoc/chapters/programming_interface.adoc
hdelan: I think this is a pretty self contained patch so I'm happy keeping it as a single commit

TApplencourt (approved):

hdelan:
@gmlueck is this OK to merge now?

gmlueck:
Merge approvals normally happen during a WG call.  Since there are now 2 organization approvals on this PR, I would expect it to be approved in next week's call.

Is there a reason you want this merged sooner?  If so, we could ask for an exception to the process (but no guarantees).

hdelan:
> Merge approvals normally happen during a WG call. Since there are now 2 organization approvals on this PR, I would expect it to be approved in next week's call.

Thanks for info. I didn't know this process. Just out of curiosity - how many organizations need to approve a patch in order for a SYCL spec change to be made?

> Is there a reason you want this merged sooner? If so, we could ask for an exception to the process (but no guarantees).

No reason at all. Happy to wait for more approvals.

gmlueck:
> Just out of curiosity - how many organizations need to approve a patch in order for a SYCL spec change to be made?

For changes to the SYCL spec, two organizations must approve, which are both different from the organization that authored the patch.

For changes to the CTS, we decided that only one organization (different from the author) is required.

hdelan:
Thanks for info

tomdeakin:
WG approved to merge.

PR #542, opened 2024-03-07
Move cl::sycl::memory_order, atomic class and atomic operations to sycl namespace

VerenaBeckham:
Developers can use sycl::memory_order.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -18977,10 +18977,8 @@ T* operator--() const
 
 The atomic types and operations on atomic types provided by SYCL 1.2.1 are
 deprecated in SYCL 2020, and will be removed in a future version of SYCL.
-The types and operations are made available in the [code]#cl::sycl::# namespace
-for backwards compatibility.
 
-The constructors and member functions for the [code]#cl::sycl::atomic# class are
+The constructors and member functions for the [code]#sycl::atomic# class are
 listed in <<table.atomics.constructors>> and <<table.atomics.members>>
 respectively.
 
@@ -18999,7 +18997,7 @@ include::{header_dir}/atomicoperations.h[lines=4..-1]
 
 
 [[table.atomics.constructors]]
-.Constructors of the [code]#cl::sycl::atomic# class template
+.Constructors of the [code]#sycl::atomic# class template
 [width="100%",options="header",separator="@",cols="65%,35%"]
 |====
 @ Constructor @ Description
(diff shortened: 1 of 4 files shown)


gmlueck (approved):

bader:
Should we add a note about this change to https://github.com/KhronosGroup/SYCL-Docs/blob/SYCL-2020/master/adoc/chapters/what_changed.adoc?

gmlueck:
I don't think this PR changes anything relative to SYCL 1.2.1.  We just removed a redundant definition of the `memory_order` enumeration from the spec.  The enumeration is still available in SYCL 2020, and it is still defined elsewhere in the spec.

AerialMantis:
I think this change makes sense, though on it's own I this will be a breaking change for the deprecated `atomic` class as that relies on the definition of the `memory_order` in the `cl::sycl` namespace.

Technically in SYCL 2020 the `sycl` and `cl::sycl` namespaces are synonymous with each other if the `CL/sycl.hpp` header is included, so the intention is that this would be equivalent to having this in the `sycl` namespace, but having an interface still explicitly defined in the `cl::sycl` namespace suggests this is an exception to the rule.

From what I can tell this synopsis, the following synopsis for deprecated atomic free functions and a couple of member functions in the accessor synopsis are the only places where the `cl::sycl` namespace is still explicitly used, I think we should extend this PR to also switch from `cl::sycl` to `sycl` here to avoid any confusion.

gmlueck:
> From what I can tell this synopsis, the following synopsis for deprecated atomic free functions and a couple of member functions in the accessor synopsis are the only places where the `cl::sycl` namespace is still explicitly used, I think we should extend this PR to also switch from `cl::sycl` to `sycl` here to avoid any confusion.

I think this makes sense.  At one point, we discussed the possibility of structuring the header file so that deprecated APIs are only available when `<CL/sycl.hpp>` is included, but that never made it into the spec.  I suspect the fact that the `atomic` interface is documented in the `cl::sycl` namespace is a holdover from that thinking.  At this point, there is no correlation between deprecated APIs and the `cl::sycl` namespace, so I think we should just document `atomic` in the `sycl` namespace.

VerenaBeckham:
I also moved the atomic class and atomic operations into the sycl namespace.

gmlueck:
> I also moved the atomic class and atomic operations into the sycl namespace.

Thanks, this looks good!

keryell (approved):
Thanks!

AerialMantis (approved):
Looks good, thanks!

tomdeakin:
One more review.

TApplencourt (approved):

tomdeakin:
WG approved to merge

PR #548, opened 2024-04-04
Clarify max work-group size

gmlueck:
Clarify the two information descriptors:

* `info::device::max_work_group_size`
* `info::kernel_device_specific::work_group_size`

These queries correspond to the OpenCL queries
`CL_DEVICE_MAX_WORK_GROUP_SIZE` and `CL_KERNEL_WORK_GROUP_SIZE`, and I shamelessly stole much of the description from the OpenCL specification.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1805,9 +1805,10 @@ info::device::max_work_group_size
       The minimum value is 1.
       This value is an upper limit and will not necessarily maximize
       performance.
-      Some kernels may not be able to execute this many work-items in a
-      work-group.
-      See also [code]#info::kernel_device_specific::work_group_size#.
+      The maximum number of work-items in a work-group depends on the kernel and
+      the implementation.
+      Use [code]#info::kernel_device_specific::work_group_size# to query this
+      limit.
 
 a@
 [source]
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1800,8 +1800,8 @@ info::device::max_work_group_size
 ----
 
     @ [.code]#size_t#
-   a@ Returns the maximum number of work-items in a work-group that this device
-      is capable of executing on a single compute unit.
+   a@ Returns the maximum number of work-items that this device
+      is capable of executing in a work-group .
       The minimum value is 1.
       This value is an upper limit and will not necessarily maximize
       performance.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1800,8 +1800,14 @@ info::device::max_work_group_size
 ----
 
     @ [.code]#size_t#
-   a@ Returns the maximum number of work-items that are permitted in a work-group executing a kernel on a single compute unit.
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -1800,8 +1800,14 @@ info::device::max_work_group_size
 ----
 
     @ [.code]#size_t#
-   a@ Returns the maximum number of work-items that are permitted in a work-group executing a kernel on a single compute unit.
+   a@ Returns the maximum number of work-items in a work-group that this device
+      is capable of executing on a single compute unit.
```
Pennycook, line 1804: I think the mention of a "single compute unit" doesn't add value here.  The whole notion of a compute unit is a little fuzzy anyway, and the wording here implies that there might be a different number of work-items if a work-group executed over a different number of compute units.

```suggestion
   a@ Returns the maximum number of work-items that this device
      is capable of executing in a work-group .
```

gmlueck: Sounds good to me.  @bashbaug, I saw that you gave this a thumbs-up.  Note that [OpenCL](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_WORK_GROUP_SIZE) has the same wording.  Should that be changed also?

adoc/chapters/programming_interface.adoc
```diff
@@ -1800,8 +1800,14 @@ info::device::max_work_group_size
 ----
 
     @ [.code]#size_t#
-   a@ Returns the maximum number of work-items that are permitted in a work-group executing a kernel on a single compute unit.
+   a@ Returns the maximum number of work-items that this device is capable of
+      executing in a work-group.
       The minimum value is 1.
+      This value is an upper limit and will not necessarily maximize
+      performance.
+      Some kernels may not be able to execute this many work-items in a
```
AerialMantis, line 1808: Perhaps we should rephrase this to be more instructive as to why some kernels may not be able to execute the maximum number of work-items. Something like:

"The actual maximum number of work-items that a kernel can execute in a work-group will depend on other factors of the kernel which can impact occupancy.

AerialMantis (approved):
Looks good, just a minor suggestion on the wording.

adoc/chapters/programming_interface.adoc
gmlueck: I'm not sure we want the spec to speculate about why a kernel might have a lower limit.  The reasons for the lower limit seem like implementation details.

AerialMantis: I agree, though I think we can provide a bit more context without being too specific, perhaps we could say that the actual maximum number of work-items per work-group for a compiled kernel is implementation defined, and may not reach the maximum specific here.

gmlueck: Makes sense.  See ed18871275cb6911e52c8100141996e73e76c7f8

nliber (approved):

TApplencourt:
Nevin was faster than me!

illuhad (approved):
LGTM!

PR #549, opened 2024-04-16
Fix typo with "accessor::get_pointer"

gmlueck:
It appears that we introduced a typo with #432 with the return type of `accessor::get_pointer`.  In the synopsis, we specified it as:

```
global_ptr<DataT> get_pointer() const noexcept
```

But in the description, it seems that we made a typo and said:

```
global_ptr<access::decorated::legacy> get_pointer() const noexcept
```

I believe these are both wrong, and it should be:

```
global_ptr<value_type> get_pointer() const noexcept
```

Using `value_type` is better than `DataT` because `value_type` is const when the accessor is read-only.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -6927,7 +6927,7 @@ accessor was constructed.  For other accessors, returns the default constructed
 a@
 [source]
 ----
-global_ptr<access::decorated::legacy> get_pointer() const noexcept
+global_ptr<value_type> get_pointer() const noexcept
 ----
    a@ Available only when [code]#(AccessTarget == target::device)#.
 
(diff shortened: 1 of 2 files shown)


Naghasan (approved):

greglueck (approved):
Checking to see if a non-member can approve a PR.

TApplencourt (approved):

AerialMantis (approved):

keryell (approved):
Thanks.

keryell:
I have the feeling that https://github.com/KhronosGroup/SYCL-Docs/pull/432/files#diff-498c175f1b8454ffa58e5f5d9db2c295f04199f9833a50e48159b9534a4cb496R68 looks weird too.

gmlueck:
> I have the feeling that https://github.com/KhronosGroup/SYCL-Docs/pull/432/files#diff-498c175f1b8454ffa58e5f5d9db2c295f04199f9833a50e48159b9534a4cb496R68 looks weird too.

Actually, that is not wrong because `value_type` and `DataT` are always the same for `local_accessor`.  However, I agree that it can be confusing.  I tried to clarify this in #566.

PR #555, opened 2024-05-21
Allow raw pointers in addition to multi_ptr

gmlueck:
Several functions accepted `multi_ptr` as a parameter type but did not accept raw pointer.  This was confusing to users, especially in cases where the `multi_ptr` was allowed to be in `generic_space`.  Since raw pointers behave like generic space multi-pointers, there's no reason to disallow raw pointers for these functions.

Discussing this within our team, we felt this was an oversight, and not something we intentionally left out.  Therefore, we think it is appropriate as a bug fix to SYCL 2020.

This commit expands the parameter types allowed by these functions to include raw pointers.

I did *not* change `async_work_group_copy` because this function takes `multi_ptr` parameters in a specific address space (which did not include `generic_space`).  Therefore, allowing raw pointers for this function seemed inappropriate.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17237,16 +17237,26 @@ Return an instance of the implementation-defined intermediate class template [co
 a@
 [source]
 ----
-template <access::address_space AddressSpace, access::decorated IsDecorated>
-void load(size_t offset, multi_ptr<const DataT, AddressSpace, IsDecorated> ptr)
+template <access::address_space AddressSpace,
+          access::decorated IsDecorated>
+void load(
+  size_t offset,
+  multi_ptr<const DataT, AddressSpace, IsDecorated> ptr)
+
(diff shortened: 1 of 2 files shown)


TApplencourt:
Agree! I guess we need to update the CTS too

gmlueck:
> Agree! I guess we need to update the CTS too

Yes, I definitely agree!

BTW, this PR is draft state because the CI tests currently fail with the PDF build.  I need to do some format restructuring to fix this, and I plan to do that in a separate PR.

gmlueck:
> BTW, this PR is draft state because the CI tests currently fail with the PDF build. I need to do some format restructuring to fix this, and I plan to do that in a separate PR.

The CI is fixed now due to #556.  This PR is ready to review!

@TApplencourt, since you said you agree in the previous comment, could you also approve?

TApplencourt (approved):

AerialMantis (approved):

illuhad (approved):

PR #556, opened 2024-05-23
Update formatting for builtin functions

gmlueck:
This is a purely editorial change that changes the formatting of the
builtin function sections 4.17.4 "Math function" through 4.17.10
"Relational functions".  The new format does the following:

* Removes the giant table the encloses all of the functions in each
  section.  Instead, we just draw a horizontal line between the
  description of each function.

* Removes the table around each code synopsis.  This table existed only
  to draw a border around the synopsis in the HTML render.  Instead, I
  created a new [source] style called "codebox", and I added some CSS to
  draw a border around these source blocks.

The tables were causing problems in the PDF render because the PDF
rendering tools fail if any single table row is taller than a page.
By removing the tables, we avoid this limitation.

As an aside, I noticed that the [source,linenums] blocks *were*
rendered in HTML with a border.  It turns out this happens because
adding "linenums" causes the block to be rendered as a table in HTML.
(The line numbers are in one column, and each source line is in a
second column.)  This causes the block to get a border just by accident
because the default CSS format for tables draws a border around the
outside of the table.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20853,8 +20853,8 @@ determined by their linear id.
 [[sec:math-functions]]
 === Math functions
 
-<<table.math.functions>> describes the math functions that are available in the
-[code]#sycl# namespace in both host and device code.
+This section describes the math functions that are available in the [code]#sycl#
+namespace in both host and device code.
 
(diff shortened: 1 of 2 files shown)


gmlueck:
Merging as editorial.

PR #557, opened 2024-05-27
Deprecate device selector types.

VerenaBeckham:
Since these types are provided only for 1.2 compatibility they should have been deprecated, since we won't support 1.2 compatibility beyond SYCL 2020.
I also modified some references to the types to be to the new _v objects instead.


gmlueck (approved):
LGTM!

nliber (approved):

PR #559, opened 2024-06-09
Add HTML links for "synopsis" code blocks

gmlueck:
Extend and improve the "codebox" Asciidoc role:

* Rename "codebox" to "synopsis".  My intent is to use this role
  mostly to define the synopsis of an API.

* Create an Asciidoctor extension that enables HTML links for source
  listings that use the "synopsis" role and also define an id like:

  ```
  [source,role=synopsis,id=XXX]
  ```

  In the HTML render, when you hover the mouse over a code synopsis, a
  pound sign ("#") appears.  Clicking on this gives you a link directly
  to that code synopsis.  This makes it easy to get a direct link to
  any API that is defined with the "synopsis" role.  Note that this
  behavior is similar to the way section headers work in the HTML
  render, where you can click on a "section sign" symbol to get a URL
  directly to a section.

Change the synopses of the math builtin function sections to use this
new style.


gmlueck:
Merging as editorial.

PR #560, opened 2024-06-09
Expandable TOC for HTML output

gmlueck:
Add an Asciidoctor extension and CSS styling to implement an expandable table of contents for the HTML render.  This allows the user to click an icon in the TOC to expand or collapse each heading level.

There is also an icon on the title line of the TOC which the user can click to expand or collapse all heading levels.

The initial state is that all TOC levels are collapsed, but this is tunable via the `toclevels-expanded` Asciidoc attribute.


adoc/config/accordion_toc/extension.rb
```diff
@@ -0,0 +1,188 @@
...
+      toc_level = 0
+      lines = output.lines
+      num_lines = lines.length-1
+      for i in 0..num_lines
+        line = lines[i]
+        next_line = (i < num_lines) ? lines[i+1] : ''
+
+        # Keep track of the TOC level by counting the nesting of the <ul> and
+        # </ul> elements that are in the TOC.
+        if TocStart.match(line) then is_in_toc = true end
```
TApplencourt, line 142:
```suggestion
        is_in_toc = True if TocStart.match(line) 
```
And so forth. You can use `https://github.com/rubocop/rubocop`, I think ti will "fix" this kind of ruby "code-smell" idiom. 

But not important, I can always do another PR to make stuff more ruby pretty.

adoc/config/accordion_toc/extension.rb
```diff
@@ -0,0 +1,188 @@
...
+      else
+        toc_levels_expanded = -1
+      end
+      new_output = ''
+      is_in_toc = false
+      hide_next_ul = false
+      toc_level = 0
+      lines = output.lines
+      num_lines = lines.length-1
+      for i in 0..num_lines
```
TApplencourt, line 136: `new_output = outputs.map{ |line| }` is the ruby way (or event better just ` outputs.map{ |line| }` , as the implicit return of the map is what you are doing

adoc/config/accordion_toc/extension.rb
gmlueck: Thanks.  I think I made things better in 2c1a88474922660e4962269a958395c703495a9a, but I suspect it's still not as ruby-ish as it could be.

adoc/config/accordion_toc/extension.rb
gmlueck: Can I really use `map` here?  Note that the algorithm needs a 1-line lookahead.  Can you do that easily with `map`?

TApplencourt: Yes, it's called `each_cons`  so ouputs.map.each_cons(2) { |line, new_line| }

adoc/config/accordion_toc/extension.rb
TApplencourt: Look good, to me!

adoc/config/accordion_toc/extension.rb
gmlueck: Thanks: d513921589b2f96cb05c3fb1731267a9be522f56.

The loop never cares about the last line of input, so we can do without the `+[""]`.

TApplencourt (approved):

gmlueck:
Merging as editorial

PR #561, opened 2024-06-09
Remove "listing" cross references

gmlueck:
In order to further improve synopsis `[source]` blocks, we want to add a title, but we don't want that title to have the "Listing" prefix. In order to do this, we need to delete the `listing-caption` attribute.

However, doing this breaks cross references to listing blocks.  There were only a few such cross references, so restructure the text in those cases to remove them.


gmlueck:
Merging as editorial

PR #562, opened 2024-06-09
Add API synopsis titles and cross references

gmlueck:
Add a title to each `[source]` block that is a synopsis.  The title uses
`[apidef]` if the title is a unique name for a SYCL identifier,
otherwise it uses `[apititle]`.

Also add an extension which creates automatic cross references to the
`[apidef]` synopses.  In order to make a cross reference, use the
Asciidoc role `[api]`.  This is styled like `[code]` but also creates a
cross reference.  For example, `[api]#foo#` creates a cross reference to
the synopsis whose title is `[apidef]#foo#`.

Replace some uses of `[code]` with `[api]` in the sections defining the
math builtins.

### DIFF
--- a/adoc/config/api_xrefs/extension.rb
+++ b/adoc/config/api_xrefs/extension.rb
@@ -1,0 +1,124 @@
+# Copyright (c) 2011-2024 The Khronos Group, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+require 'asciidoctor/extensions' unless RUBY_ENGINE == 'opal'
+
+include ::Asciidoctor
+
+# Make automatic cross references for API names in the HTML render.
+#
+# This extension assumes that APIs are defined in Asciidoc as listing blocks
+# with a particular structure:
+#
+# * The listing block must define an "id" attribute;
+# * The listing block must have a title; and
+# * The title must have one occurrence of the "[apidef]" role, where the text
+#   in that role is the name of the API.
+#
+# Here is an example that defines the API "foobar":
+#
+#   .[apidef]#foobar#
+#   [source,id=api:foobar]
+#   ----
+#   int foobar(void* p)
+#   ----
+#
+# This extension also looks for occurrences of the "[api]" role.  The text in
+# this role must exactly match the text in one of the "[apidef]" roles.  This
+# extension wraps the text in the "[api]" role with a cross reference link
+# to the listing block with the corresponding "[apidef]".  For example, consider
+# Asciidoc like this:
+#
+#   Use the [api]#foobar# function to initialize the system.
+#
+# This extension wraps the text "foobar" with an HTML anchor that links to the
+# listing block whose id is "api::foobar".
+#
+# Note that an "API" can be anything.  This extension only cares that the text
+# in the "[apidef]" role matches the text in the "[api]" role.
+#
+# Typically, projects using this extension also provide some custom CSS styling
+# for the "apidef" and "api" HTML classes.
+
+class AddApiXrefs < Extensions::Postprocessor
+  include Asciidoctor::Logging
+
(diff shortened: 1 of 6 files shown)


adoc/config/api_xrefs/extension.rb
```diff
@@ -0,0 +1,124 @@
...
+      # where the <a> element above may or may not be present, depending on
+      # whether the listing block also uses the "synopsis" extension.
+      #
+      lines = output.lines
+      num_lines = lines.length-1
+      api_to_id = {}
+      match_id_last = nil
+      match_id_last2 = nil
+      for i in 0..num_lines
+        line = lines[i]
```
TApplencourt, line 69: I don't know if you wrote the code or not (and interested on a review or not) but someline line

`output.each { |line|` is more "ruby-ish"

gmlueck: Thanks.  Fixed in e119a2f8272880dadbbabe0a2654f92342ec69ad.

These last few PRs were basically my first Ruby program ever, so I'm sure I could use some pointers on style.

TApplencourt: Welcome to the Better Python :)

TApplencourt: To be fancy and avoid state, in theory, you can do 
```
      api_to_id =  output.lines.cons(3).filter_map { |pre_previous, pre_previous, current|
       continue if no_key
       [api,api_id] }.to_h
      ```
      The good thing with this approach is the check for duplicate keys can be prepared in another block. Make function more "pure" / single-responsibility  (you iterate twice ofc, but 🤷🏽 , I think "readable"  it more important than performance in ruby) 
      ```
            api_to_id_a =  output.lines.cons(3) { |pre_previous, pre_previous, current|  [api,api_id] }
      
           api_to_id_a.map { |key, _|  log("duplicated key #{key}" if key in    api_to_id}
           api_to_id = api_to_id.to_h
      ```

**Comment by TApplencourt** [comment]
For fun, in ruby, we have the in-famous [flip-flop operator](https://en.wikipedia.org/wiki/Flip-flop_(programming)); in theory, it makes dealing with scope easier. In practice, it's unreadable.

File: adoc/config/api_xrefs/extension.rb (hunk 1 above)
Comment (gmlueck) [reply adoc/config/api_xrefs/extension.rb]: Is this what you had in mind: 72e2c30e4c1aa213601405400b95aa794c33f79c ?

Comment (TApplencourt) [reply adoc/config/api_xrefs/extension.rb]: Perfect! Lots of small independent blocks make it easier to review/understand. 

I think you api_id_array can be rewritten like
```
      api_id_array = output.lines.each_cons(3).filter_map do |prev2, prev1, cur|
        api = ApiDefSpan.scan(cur).first
        api_id = ApiDefSpan.scan(prev1).first || ApiDefSpan.scan(prev2).first
        [api, api_id] if api && api_id
      end
      ```
      
This rely on the fast that `scan` return an `empty` array when not finding anything, and that `first` return `nil` (which is false) for `emtpy array`.

gmlueck: Thanks again: 602767cf29ec1671b9c730998f3c158ea3070573

gmlueck: @TApplencourt let me know if I've addressed your concerns.  I'd like to get these editorial PRs merged.

TApplencourt (approved):

adoc/config/api_xrefs/extension.rb
TApplencourt: Perfect! You are now officially a rubyst!

gmlueck:
Merging as editorial.

PR #563, opened 2024-06-10
Add new non-normative note format

gmlueck:
Add a new format for non-normative notes.  This is the same format that the C++ specification uses, and we plan to migrate to this new format as we generally improve the formatting of the SYCL spec.

### DIFF
--- a/adoc/syclbase.adoc
+++ b/adoc/syclbase.adoc
@@ -64,6 +64,17 @@ include::config/attribs.adoc[]
 :cpp20: pass:[C++20]
 // Fortunately, the cpp macro for C++ is defined by default in AsciiDoctor
 
+// Use these to format a non-normative note.  The Asciidoc source:
+//
+//  {note}Caveat emptor.
+//  {endnote}
+//
+// is formatted like:
+//
+//  [Note: Caveat emptor. -- end note]
+:note: pass:q[[_Note:_ ]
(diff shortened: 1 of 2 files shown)


adoc/syclbase.adoc
```diff
@@ -64,6 +64,17 @@ include::config/attribs.adoc[]
...
+// Use these to format a non-normative note.  The Asciidoc source:
+//
+//  {note}Caveat emptor.
+//  {endnote}
+//
+// is formatted like:
+//
+//  [Note: Caveat emptor. -- end note]
+:note: pass:q[[_Note:_ ]
+:endnote: pass:q,a[_&#8212;{nbsp}end{nbsp}note_]]
```
TApplencourt, line 76: Does `&mdash;` work by luck? Just a little more readable

gmlueck: No, `&mdash;` causes the PDF converter to fail.  In fact, the [Asciidoc documentation](https://docs.asciidoctor.org/asciidoc/latest/subs/replacements/) recommends using numeric Unicode values for this reason:

> However, using named character references can cause problems when generating non-HTML output such as PDF because the lookup table needed to resolve these names may not be defined. The recommendation is avoid using named character references, with the exception of the well-known ones defined in XML (i.e., lt, gt, amp, quot, apos). Instead, use numeric character references (e.g., \&#8364;).

TApplencourt (approved):

Issue #459, opened 2023-08-28
Are platforms required to contain at least one device?


psalz:
In section 3.7.1.1 "SYCL backend resources managed by the SYCL application" the spec [currently says](https://github.com/KhronosGroup/SYCL-Docs/blob/0722c71b17b82e096965e04c0e95a638ffee562e/adoc/chapters/architecture.adoc#sycl-backend-resources-managed-by-the-sycl-application) that (emphasis mine)

> 3. Devices: platforms provide **one or more** devices for executing SYCL kernels. In SYCL, a device is accessible through a sycl::device object.

This seems to imply that platforms have to always contain at least one device. However as far as I can tell, this is not stated explicitly anywhere else. If this is intentional, it should be stated more explicitly in sections 3.5 and 4.6.2.

At first glance however it seems to me that there is no real benefit in having this requirement, and I'd be in favor of removing it. It could be useful to enumerate all platforms that are available in theory, but don't have any available devices.

psalz:
The use case I had in mind would be for example a SYCL installation that was compiled with CUDA support, but at runtime `CUDA_VISIBLE_DEVICES=""` is set. By allowing empty platforms one could see that CUDA support is enabled in principle at least, and that the reason for an NVIDIA GPU not showing up must be lying elsewhere.

Of course if that change would have far-fetching consequences as you point out, it might not be worth it. The above use case could in principle also be achieved through backends, although AFAIK there is currently no way of programmatically listing backends (nor their names, for that matter).

TApplencourt:
I tend to agree with Greg, not sure the complexity added is worth it.

keryell:
> I tend to agree with Greg, not sure if the added complexity is worth it.

On the other hand, filtering out platforms without any device requires some complexity instead of just returning the list of known platforms.

psalz:
> I tend to agree with Greg, not sure if the added complexity is worth it.

That's fine with me, but do we agree that it should be stated more explicitly in sections 3.5 and/or 4.6.2?

tomdeakin:
WG discussion Oct/5/23

gmlueck:
Fixed with #468.

PR #567, opened 2024-06-14
Update platform description to new format

gmlueck:
Update the section describing the `platform` class to the new format.


adoc/chapters/programming_interface.adoc
```diff
@@ -963,207 +963,237 @@ are handled by throwing synchronous SYCL exceptions.
...
 ----
-   a@ Constructs a SYCL [code]#platform# instance that is a copy of the
-      [code]#platform# which contains the device returned by the
-      <<device-selector>> parameter.
 
-|====
+_Constraints:_ The [code]#DeviceSelector# must be a type that satisfies the
+requirements of a <<device-selector>> as defined in <<sec:device-selector>>.
 
+_Effects:_ The [code]#selector# is called once for every <<root-device>> as
```
TApplencourt, line 1004: Remove the `once`? Who knows, maybe some implementation wants to do two passes on the list of rood devices?

adoc/chapters/programming_interface.adoc
```diff
@@ -963,207 +963,237 @@ are handled by throwing synchronous SYCL exceptions.
...
-      parameter.
 
-a@
-[source]
+_Constraints:_ The [code]#Param# must be an information descriptor for the
+platform class.
+
+Each information descriptor specifies the return value and may also specify
+preconditions, exceptions that are thrown, etc.
+
```
TApplencourt, line 1036: Do we want to put 
```
The type alias [code]#Param::return_type# must be defined in
      accordance with the info parameters in <<table.platform.info>> to
      facilitate returning the type associated with the [code]#Param#
      parameter.
      ```
      
      
      As a `node`?

File: adoc/chapters/programming_interface.adoc [hunk 3]
```diff
@@ -963,207 +963,237 @@ are handled by throwing synchronous SYCL exceptions.
...
+Returns [code]#false# if this platform does not contain any devices.
+
+'''
+
+.[apidef]#platform::get_devices#
+[source,role=synopsis,id=api:platform-get-devices]
 ----
 std::vector<device>
-get_devices(info::device_type deviceType = info::device_type::all) const
+get_devices(info::device_type type = info::device_type::all) const
```
TApplencourt, line 1093: I guess need to be change in `Static member function` Table 24 too (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_interface). Sorry if this table doesn't exist anymore with those change,

TApplencourt:
Minimal nitpicking questions

adoc/chapters/programming_interface.adoc
gmlueck: I did this in 1b1bde486d535fe9d211b1c82d05ec6e4b5fe80a because I agree that [this section](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-selection) does not currently say that the selector is called exactly once for each device.

However, I wonder if this is an oversight in the spec.  Why wouldn't we want to guarantee that it is called exactly once for each root device?  This could make a difference for a selector that maintains internal state across each  call.

adoc/chapters/programming_interface.adoc
gmlueck: I think we do not need this paragraph anymore.  The updated format specifies the `return_type` type alias for each information descriptor.  See the synopses in the `platform` class "Information descriptors" section.

adoc/chapters/programming_interface.adoc
gmlueck: Makes sense.  Done in 576f2741a0568b45b0543d24125e99e7dd1ec8a0.

adoc/chapters/programming_interface.adoc
TApplencourt: Good point!

adoc/chapters/programming_interface.adoc
TApplencourt: I agree, but that seems more than editorial as, in theory, it may impact a state-full selector, as you said. 

I don't have a strong opinion about this one, but I lean toward adding `once` :) 

It doesn't seem like a big ask for an implementer to call their selector only once per device, and indeed, it may simplify the user's implementation of a selector

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
gmlueck: FWIW, I think these "new format" PRs will be more than pure editorial, so it should be OK to make some small improvements.  I don't mind making this as a separate PR, though, if you prefer.

TApplencourt: Maybe a new PR would give others an opportunity to chime in, and we could merge this PR without waiting for the "once" resolution.

tomdeakin:
Discussion on calling selector more than once - likely in some implementations. Open PR to discuss.

adoc/chapters/programming_interface.adoc
tomdeakin: Do we need to add a hyperlink to that section here to make it easy to find?

adoc/chapters/programming_interface.adoc
```diff
@@ -963,207 +963,237 @@ are handled by throwing synchronous SYCL exceptions.
...
-a@
-[source]
+_Constraints:_ The [code]#Param# must be a backend information descriptor for
+the platform class.
+
+_Throws:_ An [code]#exception# with the [code]#errc::backend_mismatch# error
+code if the backend that corresponds with [code]#Param# is different from the
+backend that is associated with this platform.
+
+Each information descriptor specifies the return value and may also specify
```
tomdeakin, line 1053: Do we want to add a hyperlink to the information descriptors?

adoc/chapters/programming_interface.adoc
```diff
@@ -963,207 +963,237 @@ are handled by throwing synchronous SYCL exceptions.
...
 ----
 static std::vector<platform> get_platforms()
 ----
-   a@ Returns a [code]#std::vector# containing all SYCL
-      [code]#platforms# from all <<backend, SYCL backends>> available in the system.
-
-|====
 
+_Returns:_ A [code]#std::vector# containing all of the platforms from all
+<<backend, backends>> that are known to the SYCL implementation.
```
tomdeakin, line 1112: I'm not sure "known" and "available" mean the same thing here. Would we expect this API to return backends that are currently useable at the time this is called. That is different to an implementation having a backend for devices that are not present in the system. For example, an implementation might provide a CUDA backend, but if the system has no NVIDIA GPUs, I don't think implementations show it.

tomdeakin (changes requested):
LGTM in general, and I like the new style. Two nitpicks about adding hyperlinks for user-friendliness, and one wording change we should think about to make sure we're happy it's equivalent.

keryell (approved):
That looks good!

adoc/chapters/programming_interface.adoc
gmlueck: I restored the wording about "available" platforms in c1780c506780b44f6b85cf3a83d2b589ef2a603d.

However, I wonder if this is really what we want.  The change we made in #468 (issue #459) was done specifically so that `get_platforms` could potentially return an "empty" platform for the case when the implementation supports a backend, but there are no available devices in that backend at runtime.

adoc/chapters/programming_interface.adoc
gmlueck: Yes, that's a good idea.  Done in 6915b81636ef42e0333f24aff19c08dfd62e136e.

gmlueck:
@psalz: Do you know why this is getting a failure in "Open CTS issue for spec changes / create-issue"?  It seems like there are two questions:

* Why is this workflow running at all?  This PR was already open, and it seems like the workflow triggered when I pushed a new commit.  I think it was not your intent to run this workflow on each commit to a PR.

* What causes the workflow error?  The error is "Input required and not supplied: app-id".  It looks like the YML file defines `app-id` as:

  ```
  app-id: ${{ vars.SYCL_ISSUE_BOT_APP_ID }}
  ```

  Is the problem that the variable is not defined somehow?

adoc/chapters/programming_interface.adoc
tomdeakin: I think "available" covers that case. Thanks for the change.

tomdeakin (approved):

psalz:
> * Why is this workflow running at all?  This PR was already open, and it seems like the workflow triggered when I pushed a new commit.  I think it was not your intent to run this workflow on each commit to a PR.

Addressed in https://github.com/KhronosGroup/SYCL-Docs/pull/597.

>  * What causes the workflow error?  [...] Is the problem that the variable is not defined somehow?

I'm not sure what's happening here. Maybe because the PR predates the variable being set it doesn't receive it..? We'll have to keep an eye on this one.

tomdeakin:
WG approves to merge after rev 9 release.
@gmlueck please merge after rev 9 release as editor.

gmlueck:
Rev 9 was released, so merging.

PR #480, opened 2023-10-25
Align description of logical `vec` operator overloads

AlexeySachkov:
The code snippet for `vec` interface does not list any restrictions on `operator&&` and `operator||` and they are not listed in description of overloads `(vec, vec)` and `(vec, scalar)`.

However, they were listed in description of overloads for `(scalar, vec)` and there is no reasons for them to be there.

This commit aligns description of all overloads of logical `vec` operators by removing data types restriction for
`vec::operator&&(scalar, vec)` and `vec::operator||(scalar, vec)`.

PR #570, opened 2024-06-27
Fix Asciidoc error in vec table

gmlueck:
It appears that this error was introduced a while ago in #480, but it was only flagged recently as an error in the build.  The docker image we use to build the spec was updated earlier today, and the new image flags this as an error.  It appears that older docker images did not flag an error, but the build output was corrupt.  Luckily, we did not publish any of the corrupted builds.


gmlueck:
Merging as editorial

PR #574, opened 2024-06-28
Remove unused "OPENCL_VERSION" attribute

gmlueck:
Nothing seems to use this Asciidoc attribute, so remove it.

I compared the HTML rendered output, and removing this attribute had no effect.


gmlueck:
Merging as editorial

PR #575, opened 2024-06-30
Use a specific SHA of the Docker image for running CI

oddhack:
This prevents unexpected behavior when the underlying asciidoctor-spec image is updated on dockerhub.

@gmlueck I built the HTML / PDF outputs and compared against versions built using the older image. They look similar but there are minor visible differences due to the updated asciidoctor versions (mostly that they tweaked CSS and PDF layout options in very small ways). So you should verify the outputs to your satisfaction. There's some stuff about docker in the Makefile which I don't fully understand so have not touched that - as noted in email, my approach in the Vulkan runDocker script was to extract the image name from the CI .yml file to avoid error-prone duplication. Maybe that would be applicable here.

Note this only changes the Github Actions container. In the internal SYCL spec repo there's also a .gitlab-ci.yml for our gitlab that would have to be changed as well to get consistent behavior.

Also, note I have not yet updated the dockerhub 'asciidoctor-spec' image, just the 'asciidoctor-spec.20240630' image (which is just an alias) and am using the image SHA here, as discussed.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,9 @@ jobs:
     runs-on: ubuntu-latest
 
     # Use Khronos container with asciidoctor toolchain preinstalled
-    container: khronosgroup/docker-images:asciidoctor-spec
+    # The SHA refers to a specific version of the
+    # khronosgroup/docker-images:asciidoctor-spec image.
+    container: khronosgroup/docker-images@sha256:bd30a83285a2ea062598f053b5bd8ebc843e16c639c0e4cd88ab4bbb4e63ead3
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it


oddhack:
That works too (in most cases, see below). I did not actually realize you could include both image name and SHA at the same time, that's good to know. Happy to close this in favor of #576, it's your repo.

Re docker-notify, if you were asking to be notified when 'asciidoctor-spec' updates, it has not updated yet, because I did not want to do that before verifying the image worked for SYCL in order not to repeat last week's behavior. But I did need to get it on dockerhub so it could be pulled, thus I pushed only the name with the date tag appended for the moment. If you are OK with updating 'asciidoctor-spec' by that name, I will go ahead with that. Would like to get the 6 GB of build cache off my machine ASAP :-)

gmlueck:
> That works too (in most cases, see below). I did not actually realize you could include both image name and SHA at the same time, that's good to know. Happy to close this in favor of #576, it's your repo.

I did some more testing, and it turns out that docker just ignores the tag if you specify both a tag and a SHA.  You don't get any sort of error if the SHA doesn't match the tag name.  Therefore, it seems better to just specify the SHA, as you have.  At this point, I think #576 is the same as #575, except for the comment.

> Re docker-notify, if you were asking to be notified when 'asciidoctor-spec' updates, it has not updated yet, because I did not want to do that before verifying the image worked for SYCL in order not to repeat last week's behavior.

I signed up for notifications to "khronosgroup/docker-images", so I think I should have been notified.  It seems strange that I was not.

> If you are OK with updating 'asciidoctor-spec' by that name, I will go ahead with that. Would like to get the 6 GB of build cache off my machine ASAP :-)

I did a comparison of the generated HTML as well as a quick visual check of the generated spec.  I think the new docker image is fine, and I plan to merge #576.  Go again and update 'asciidoctor-spec' if you want.  The SYCL-Doc build won't use that tag name any more.

oddhack:
Possibly the docker-notify tool is not working. I filed an internal issue to ask @outofcontrol for any thoughts on doing this, and emailed Enso who hosts docker-notify but have not heard back, yet.

oddhack:
Closing this in favor of #576.

PR #576, opened 2024-07-01
Reference Docker Hub image by SHA

gmlueck:
Download the Docker Hub container image by its SHA, so that we know exactly which version of the Asciidoctor tools we use to build the spec. The version (SHA) should be updated periodically as new container images are published.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,8 +25,11 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
-    # Use Khronos container with asciidoctor toolchain preinstalled
-    container: khronosgroup/docker-images:asciidoctor-spec
+    # Use the Khronos container with the asciidoctor toolchain preinstalled.  We
+    # reference the image by its SHA rather than its tag because they sometimes
+    # overwrite a tag with a different image (which has a different SHA).  This
+    # corresponds to tag 20240627.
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -25,8 +25,11 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
-    # Use Khronos container with asciidoctor toolchain preinstalled
-    container: khronosgroup/docker-images:asciidoctor-spec
+    # Use the Khronos container with the asciidoctor toolchain preinstalled.  We
+    # reference the image its SHA rather than its tag because they sometimes
+    # overwrite a tag with a different image (which has a different SHA).  This
+    # SHA corresponds to tag "asciidoctor-spec.20240630".
+    container: khronosgroup/docker-images@sha256:bd30a83285a2ea062598f053b5bd8ebc843e16c639c0e4cd88ab4bbb4e63ead3
```
oddhack, line 32:
```suggestion
    # SHA corresponds to tag "asciidoctor-spec.20240701".
    container: khronosgroup/docker-images@sha256:1175e55feeaca36d8c53b3628372cd371078473a162a558c3e89ffafdde40676
```

I had to rebuild the image to push by both names (asciidoctor-spec and asciidoctor-spec.20240630), and did not realize that even though the entire build was cached, I'd still get a different SHA out the backend. So this should work identically, but is now current.

gmlueck:
Merging as editorial

PR #581, opened 2024-07-10
Update to latest docker image

gmlueck:
I verified that the generated HTML is unchanged when building with the new image.


TApplencourt (approved):

gmlueck:
Merging as editorial

PR #582, opened 2024-07-12
Relax backend for default-constructed event

gmlueck:
Relax the requirements on the backend for a default-constructed `event` object.  Previously, we required the backend for such an event to be the same as the backend for the default device.  After some implementation experience, we decided that this is not practical.  Requiring interoperation with a particular backend constrains the implementation too much.  It also seems arbitrary to require the default-constructed event to have a particular backend.  To address these concerns, we now say that the backend that is associated with a default-constructed event is implementation defined.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3794,9 +3794,7 @@ event()
       Waiting on this [code]#event# will return immediately and querying
       its status will return [code]#info::event_command_status::complete#.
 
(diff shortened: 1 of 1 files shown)


TApplencourt (approved):

illuhad (approved):
LGTM!

tomdeakin:
Discussed previously and revisit.

keryell (approved):

PR #585, opened 2024-07-17
Disallow repeated swizzle element in assignment

gmlueck:
We previously decided that assignment to a `__writeable_swizzle__` should have a constraint that the lhs does not have any repeated swizzle elements.  Thus, something like this is disallowed:

```
vec<int, 2> x;
myvec.swizzle<1,1>() = x;
```

We did this because the result would depend on the order in which the assignments occurred, and we think that some compilers will not handle this case well.

However, we neglected to add the same constraint to other assignment-like swizzle operations.  This commit adds a similar constraint, which disallows all of the following cases also, where the `__writeable_swizzle__` has repeated elements:

```
myvec.swizzle<1,1>().load(off, p);
myvec.swizzle<1,1>() += x;
myvec.swizzle<1,1>()++;
++(myvec.swizzle<1,1>());
```

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17979,6 +17979,9 @@ void load(size_t offset, multi_ptr<const DataT, AddressSpace, IsDecorated> ptr) 
 |====
 _Availability:_ Available only in [code]#+__writeable_swizzle__+#.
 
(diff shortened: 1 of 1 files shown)


AerialMantis (approved):
Looks good to me.

TApplencourt (approved):

tomdeakin (approved):

PR #586, opened 2024-07-20
Reflow improvements

gmlueck:
* Break line after sentence in parentheses.
* Do not end sentence on "etc."


gmlueck:
Merging as editorial.

PR #590, opened 2024-07-22
Remove "heterogeneous" where it doesn't make sense

TApplencourt:
We agreed that the restriction of "heterogeneous" devices makes little sense.
For example, this means that pure-OpenMP implementations are not "backend"  

I guess we should clarify that cycle provides a portable abstracted discrete device concept that can be used regardless of the physical hardware (discrete, integrated, same ISA, or heterogenous...). But I was not smart enough to put it in the spec, so this pr just removes `heterogeneous` from a few places.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -17,8 +17,8 @@ The terminology used for SYCL inherits historically from OpenCL with some
 SYCL-specific additions.
 However SYCL is a generic {cpp} programming model that can be laid out on top of
 other APIs apart from OpenCL.
-SYCL implementations provide <<backend>>s for various APIs,
-implementing the SYCL general specification on top of them.
+SYCL implementations provide <<backend>>s for various APIs, implementing the
+SYCL general specification on top of them.
 We refer to this API as the <<backend-api>>.
 The SYCL general specification defines the behavior that all SYCL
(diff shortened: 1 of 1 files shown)


adoc/chapters/architecture.adoc
```diff
@@ -16,10 +16,10 @@ can run on either an heterogeneous device or on the <<host>>.
 The terminology used for SYCL inherits historically from OpenCL with some
 SYCL-specific additions.
 However SYCL is a generic {cpp} programming model that can be laid out on top of
-other heterogeneous APIs apart from OpenCL.
-SYCL implementations can provide <<backend>>s for various heterogeneous APIs,
-implementing the SYCL general specification on top of them.
-We refer to this heterogeneous API as the <<backend-api>>.
+other APIs apart from OpenCL.
+SYCL implementations provide <<backend>>s for various APIs, implementing the
```
tomdeakin, line 20: Tidy this, must have 1 backend, not necessarily plural.

TApplencourt:
Put back the `can`  as it opens a new can of worms. We will deal with it in another PR.

tomdeakin (approved):

gmlueck (approved):

tomdeakin:
WG approved to merge.

TApplencourt:
Thanks guys!

PR #593, opened 2024-07-24
Clarify that "sycl::free(nullptr)" does nothing

gmlueck:
The `std::free` function allows its parameter to be null, in which case it has no effect.  I assume we want the same behavior for `sycl::free`.


nliber (approved):

tomdeakin (approved):

keryell (approved):
Let's start 1 year of discussion on this one about C compatibility.. ;-)

AerialMantis (approved):

PR #594, opened 2024-07-25
Reflow and "etc."

gmlueck:
Do not break a line after "etc." when the next word starts with a lower case letter.  However, do break a line after "etc." when the next word starts with an upper case letter.


gmlueck:
Merging as editorial

PR #595, opened 2024-07-28
Update to the latest Docker image

gmlueck:
I verified that the HTML output is identical when building with this image vs. the previous one.


gmlueck:
Merging as editorial

PR #596, opened 2024-07-28
Some editorial cleanup

gmlueck:
* Remove the footer from the HTML build, which contained a confusing date.

* Remove some unused Asciidoc attributes that I think we inherited from another project when we first copied our Asciidoc configuration files.

* Fix some minor bugs in the "syclbase.adoc" file.

I compared the HTML render, and the only change this introduces is due to the first bullet item above.  The HTML output no longer contains the line "Last updated" with a confusing date.


gmlueck:
The error in "Open CTS issue" is unrelated.
Merging as editorial

PR #599, opened 2024-08-04
Generalize synopsis role

gmlueck:
Allow "[role=synopsis]" to be used with paragraphs in addition to "[source]" blocks.  Paragraphs with this role are styled with a clickable "#" that links to the paragraph, similar to the styling for "[source]" blocks with this role.

This lays the groundwork for some future reformatting and does not change any formatting in the current render of the specification.


gmlueck:
Merging as editorial

PR #600, opened 2024-08-04
Enable additional API cross-references

gmlueck:
Enhance the api-xrefs Asciidoc extension to accept an attribute `api-xrefs`, which contains additional mappings between API names and IDs. This allows us to define `[code]#api#` cross references to things that do not have a matching `[apidef]`.

This lays the groundwork for some future reformatting and does not change any formatting in the current render of the specification.


gmlueck:
Merging as editorial

PR #601, opened 2024-08-04
Update device description to new format

gmlueck:
Update the section describing the `device` class to the new format.


adoc/chapters/programming_interface.adoc
```diff
@@ -1422,1390 +1422,1969 @@ info::context::atomic_fence_scope_capabilities
...
-info::device::native_vector_width_int
-info::device::native_vector_width_long
-info::device::native_vector_width_float
-info::device::native_vector_width_double
-info::device::native_vector_width_half
-----
+_Returns:_ The maximum number of work-items that this device is capable of
+executing in a work-group.
+The minimum value is 1.
+This value is an upper limit and will not necessarily maximize performance.
```
TApplencourt, line 1854: Should it be a `note`?

adoc/chapters/programming_interface.adoc
```diff
@@ -1422,1390 +1422,1969 @@ info::context::atomic_fence_scope_capabilities
...
 The [code]#device# class also provides constructors, but constructing a new
 [code]#device# instance merely creates a new object that is a copy of one of the
-objects returned by [code]#device::get_devices()#.
+objects returned by [api]#device::get_devices#.
 
-A SYCL [code]#device# can be partitioned into multiple SYCL devices, by calling
-the [code]#create_sub_devices()# member function template.
-The resulting SYCL [code]#devices# are considered sub devices, and it is valid
+A device can be partitioned into multiple devices, by calling the
+[code]#device::create_sub_devices# member function template.
```
TApplencourt, line 1446: should be `code` or `api`?

adoc/chapters/programming_interface.adoc
```diff
@@ -1422,1390 +1422,1969 @@ info::context::atomic_fence_scope_capabilities
...
+If a device returns one of these, it must return them all.
 
-    @ [.code]#info::partition_property#
-   a@ Returns the partition property of this SYCL [code]#device#. If this SYCL [code]#device# is not a sub device then the return value must be [code]#info::partition_property::no_partition#, otherwise it must be one of the following values:
---
-  * [code]#info::partition_property::partition_equally#
-  * [code]#info::partition_property::partition_by_counts#
-  * [code]#info::partition_property::partition_by_affinity_domain#
---
+At a minimum, each device must support [code]#memory_order::relaxed#.
```
TApplencourt, line 2545: Is this should be a `__remaks_` or ` __postcondition__` ? ( without thinking to much, all the "must" may be in a `condition`)

adoc/chapters/programming_interface.adoc
```diff
@@ -1422,1390 +1422,1969 @@ info::context::atomic_fence_scope_capabilities
...
-a@
-[source]
+_Constraints:_ Available only when [code]#Param# is a backend information
+descriptor for the device class.
+
+_Throws:_ An [code]#exception# with the [code]#errc::backend_mismatch# error
+code if the backend that corresponds with [code]#Param# is different from the
+backend that is associated with this device.
+
+Each information descriptor specifies the return value and may also specify
```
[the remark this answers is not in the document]
gmlueck: This is a *backend* info descriptor query, so the info descriptors and each `return_type` will be defined in the backend interop specifications.  It appears that we do not have any backend-specific info descriptors for the OpenCL backend (Appendix C).  If some other backend interop spec defines any, I think it should use the new format we use in the core spec, which defines the `return_type`.

> And should be that in a `__note__` the section?

I have a weak preference to keep this text normative, but I could change it if you feel strongly.  Note that we have the same language above for `get_info`, so we should change both if we decide to change it.  We also use this normative language for the `platform` versions of these member functions.

adoc/chapters/programming_interface.adoc
gmlueck: This text comes from OpenCL (see [CL_DEVICE_MAX_WORK_GROUP_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_WORK_GROUP_SIZE)), where is is also normative.  I think it makes sense for it to be normative here too.

adoc/chapters/programming_interface.adoc
```diff
@@ -1422,1390 +1422,1969 @@ info::context::atomic_fence_scope_capabilities
...
 [code]#device# instance merely creates a new object that is a copy of one of the
-objects returned by [code]#device::get_devices()#.
+objects returned by [api]#device::get_devices#.
 
-A SYCL [code]#device# can be partitioned into multiple SYCL devices, by calling
-the [code]#create_sub_devices()# member function template.
-The resulting SYCL [code]#devices# are considered sub devices, and it is valid
+A device can be partitioned into multiple devices, by calling the
+[code]#device::create_sub_devices# member function template.
+The resulting [code]#device# objects are considered sub devices, and it is valid
```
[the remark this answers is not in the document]
TApplencourt: Oh forgot about that! Thanks

adoc/chapters/programming_interface.adoc
TApplencourt: >This value is an upper limit and will not necessarily maximize performance.

It seems to be a weird norm to me. But ok

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
TApplencourt: We can keep it normative.

adoc/chapters/programming_interface.adoc
TApplencourt: Oh, I see. Thanks for the explanation!

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -1454,1390 +1454,1969 @@ info::context::atomic_fence_scope_capabilities
...
+==== Constructors
 
-[[table.constructors.device]]
-.Constructors of the SYCL [code]#device# class
-[width="100%",options="header",separator="@",cols="65%,35%"]
-|====
-@ Constructor @ Description
-a@
-[source]
+.[apititle]#Default constructor#
```
keryell, line 1495: Just curious: is this leading `.` on purpose?

gmlueck: Yes.  The `.` Asciidoc syntax adds a title to the `[source]` block below.

keryell: Thanks for the explanation. I have not read the AsciiDoctor documentation for quite a while...

keryell (approved):
What a PR!

tomdeakin:
WG approved to merge.

PR #604, opened 2024-08-16
Undeprecate legacy multi_ptr

steffenlarsen:
The legacy decoration for multi_ptr has been made the default template argument of the decoration in multi_ptr to bridge the gap between SYCL 1.2.1 and SYCL 2020 multi_ptr interfaces. However, the legacy decoration and the specialization of multi_ptr with the decoration is deprecated with SYCL 2020. This is confusing to users as they may easily run into deprecation warnings simply due to not specifying the decoration. This commit proposes that we remove the deprecation of legacy in SYCL 2020.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -9138,7 +9138,6 @@ For interoperability with the <<backend>>, users should rely on types exposed by
 the decorated version.
 If the value of [code]#access::decorated# is [code]#access::decorated::legacy#,
 the 1.2.1 interface is exposed.
-This interface is deprecated.
 
 The template traits [code]#remove_decoration# and type alias
 [code]#remove_decoration_t# retrieve the non-decorated pointer or reference from
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -9846,10 +9846,6 @@ below.
 include::{header_dir}/pointer.h[lines=4..-1]
 ----
 
-Note that using [code]#global_ptr#, [code]#local_ptr#, [code]#constant_ptr# or
-[code]#private_ptr# without specifying the decoration is deprecated.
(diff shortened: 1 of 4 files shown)


steffenlarsen:
Note that `async_work_group_copy` using legacy `multi_ptr` is still deprecated. We can undeprecate them, but their signatures will be slightly different from the yes-no deprecated ones.

nliber (approved):

VerenaBeckham:
How about the[ statement in programming_interface.adoc](https://github.com/KhronosGroup/SYCL-Docs/blob/4e1bf7dc937a0a7c6c4808e7d07a6a115aa83507/adoc/chapters/programming_interface.adoc?plain=1#L9850):

> Note that using [code]#global_ptr#, [code]#local_ptr#, [code]#constant_ptr# or
> [code]#private_ptr# without specifying the decoration is deprecated.
> The default argument is provided for compatibility with 1.2.1.

Should that also be removed as part of this MR?

steffenlarsen:
> How about the[ statement in programming_interface.adoc](https://github.com/KhronosGroup/SYCL-Docs/blob/4e1bf7dc937a0a7c6c4808e7d07a6a115aa83507/adoc/chapters/programming_interface.adoc?plain=1#L9850):
> 
> > Note that using [code]#global_ptr#, [code]#local_ptr#, [code]#constant_ptr# or
> > [code]#private_ptr# without specifying the decoration is deprecated.
> > The default argument is provided for compatibility with 1.2.1.
> 
> Should that also be removed as part of this PR?

Good catch! It has been removed.

AerialMantis (approved):
I think we might need to update the change log as well, but other than that this change LGTM.

gmlueck (approved):

keryell (approved):

gmlueck:
WG approved.

PR #606, opened 2024-08-25
Update queue description to new format

gmlueck:
Update the section describing the `queue` class to the new format.


TApplencourt:
Thanks will review this week!

adoc/headers/queue.h
```diff
@@ -53,22 +53,25 @@ class queue {
 
   bool is_in_order() const;
 
-  template <typename Param> typename Param::return_type get_info() const;
+  template <typename Param>
```
TApplencourt, line 56: It's not important, but AFAIK those files were formatted via clang format.
In the future, we should commit a clang-format-config on we a style we like and stick with it.

adoc/headers/queue.h
```diff
@@ -177,16 +147,48 @@ class queue {
...
+  event mem_advise(void* ptr, size_t numBytes, int advice, event depEvent);
+  event mem_advise(void* ptr, size_t numBytes, int advice,
+                   const std::vector<event>& depEvents);
+
+  /* -- Shortcut functions: update_host -- */
 
   template <typename T, int Dims, access_mode Mode, target Tgt,
             access::placeholder IsPlaceholder>
-  event fill(accessor<T, Dims, Mode, Tgt, IsPlaceholder> dest, const T& src);
+  event update_host(accessor<T, Dim, Mode, Tgt, IsPlaceholder> acc);
```
TApplencourt, line 192: Just realized it was a mistake. Should be `Dims`. `Dim` not defined in the template.

adoc/chapters/programming_interface.adoc
```diff
@@ -2936,845 +2933,841 @@ Any errors from the <<command-group>> after it has been submitted are handled by
 passing <<async-error,asynchronous errors>> at specific times to an
 <<async-handler>>, as described in <<error-handling>>.
 
-A SYCL [code]#queue# can wait for all <<command-group,command groups>> that it
-has submitted by calling [code]#wait# or [code]#wait_and_throw#.
-
-The default constructor of the SYCL [code]#queue# class will construct a queue
```
TApplencourt, line 2942: Why was this removed? because it was redundant with other part of the spec?

adoc/chapters/programming_interface.adoc
```diff
@@ -2936,845 +2933,841 @@ Any errors from the <<command-group>> after it has been submitted are handled by
...
-      accordance with the info parameters in <<table.queue.info>> to
-      facilitate returning the type associated with the [code]#Param#
-      parameter.
 
-a@
-[source]
+_Effects:_ Submit a <<command-group-function-object>> to the queue, in order to
+be scheduled for execution on the device.
+
+_Returns:_ An event which the application can use to determine when the
```
TApplencourt, line 3231:
```
An event which represents the <<command>> that is submitted to the queue.
_note_ The even can be used to determine <<command>> associated with the <<command-group-function-object>> has completed.
```

And I guess we we should use `<<command>>` in al the `event` description.

adoc/chapters/programming_interface.adoc
```diff
@@ -2936,845 +2933,841 @@ Any errors from the <<command-group>> after it has been submitted are handled by
...
-      template parameter [code]#Param#.
-      The type alias [code]#Param::return_type# must be defined in
-      accordance with the info parameters in <<table.queue.info>> to
-      facilitate returning the type associated with the [code]#Param#
-      parameter.
 
-a@
-[source]
+_Effects:_ Submit a <<command-group-function-object>> to the queue, in order to
+be scheduled for execution on the device.
```
TApplencourt, line 3229: Missing the `Synchronous errors are reported through SYCL exceptions.`?

adoc/headers/queue.h
gmlueck: Good catch.  Fixed in 46907d4df3c98edad84fa1ab3e68b6f6b419cc65

adoc/chapters/programming_interface.adoc
gmlueck: Yes, exactly.  The description for each constructor contains this information.  No need to repeat it here.

adoc/chapters/programming_interface.adoc
gmlueck: The text I have here describing the `submit` function is exactly the same as the text prior to this PR.  The only thing I added was the word "Effects".

I do agree that the `submit` function could throw a synchronous exception if there is an error condition, but the sentence you propose was not there before.  Are you suggesting that we should add it as part of this PR?

TApplencourt: Thanks for the clarificaition.
We can do it in a follow-up PR. Cleaner indeed

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
gmlueck: I know you already approved this, but I think your comment here is a good suggestion.  I updated the wording in 3cc87507efa5164df8c3395921b29b06fe40745f.

TApplencourt: I like it thanks!

TApplencourt (approved):

keryell (approved):
Thanks!

tomdeakin:
WG approved to merge.

PR #608, opened 2024-08-26
Remove info::platform::profile from the header

VerenaBeckham:
It is never described in the spec.
Alternatively, we could bring the description in the spec back and mark it deprecated.

PR #609, opened 2024-08-27
CTS issue workflow: Don't assign PR author to issue

psalz:
The workflow may fail when trying to assign a person that is not a member of the @KhronosGroup organization to the CTS issue, as can be seen [in this run](https://github.com/KhronosGroup/SYCL-Docs/actions/runs/10559189740/attempts/1) for #608. Unfortunately there is no simple way of detecting this in advance, so I think we have to instead remove the assignment mechanism altogether.


gmlueck (approved):

tomdeakin:
WG approved to merge

PR #610, opened 2024-08-30
Prepare for next SYCL spec major release

gmlueck:
Update the SYCL version number from "2020" to "NEXT".

Also simplify some of the macros in the Makefile related to SYCL version.  I verified this cleanup has no effect on the generated HTML by comparing the HTML before and after this change.

NOTE: The generated spec output file names are now changed from "sycl-2020.html" to "sycl-next.html" and from "sycl-2020.pdf" to "sycl-next.pdf".

This corresponds to internal MR 726 and 756


gmlueck:
Merging as editorial

PR #614, opened 2024-09-06
Fix typo

hdelan:
Allocation is for device USM, not shared USM.

### DIFF
--- a/adoc/code/usm_device.cpp
+++ b/adoc/code/usm_device.cpp
@@ -9,8 +9,7 @@ int main() {
   // Create a default queue to enqueue work to the default device
   queue myQueue;
 
-  // Allocate device global memory bound to the device and context associated
-  // with the queue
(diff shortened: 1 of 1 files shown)


adoc/code/usm_device.cpp
```diff
@@ -9,8 +9,8 @@ int main() {
   // Create a default queue to enqueue work to the default device
   queue myQueue;
 
-  // Allocate shared memory bound to the device and context associated to the
-  // queue
+  // Allocate device global memory bound to the device and context associated
+  // with the queue
```
gmlueck, line 13:
```suggestion
  // Allocate device USM, using the device and context associated with the queue
```

We should use consistent terminology.  Elsewhere in the spec we call this "device USM".

hdelan: Change made.

gmlueck (approved):

gmlueck:
Merging as editorial

PR #615, opened 2024-09-08
Deprecate fallback/secondary queue.

gmlueck:
At the F2F in Brussels there was general agreement to deprecate the fallback/secondary queue feature. SYCL SC will likely be removing it.

This corresponds to internal MR 754


gmlueck:
Merging as the WG already approved this change when it was an internal MR.

PR #616, opened 2024-09-08
Add framework for unified spec

gmlueck:
* Add a section to the introduction explaining that this document is a
  specification for SYCL 2020 and SYCL-Next and also for the KHR
  extensions.

* Add a (mostly empty) appendix that will hold the future KHR
  extensions.

* Change some references to "SYCL 2020" to more general wording, since
  they apply also to SYCL-Next.

* Remove references to "SYCL Next" and use the Asciidoc SYCL_VERSION
  attribute instead.

This corresponds to internal MR 755


gmlueck:
Merging as the WG already approved this change when it was an internal MR.

PR #625, opened 2024-09-19
Clarify and deprecate "execution_capabilities"

gmlueck:
We had a customer bug report about the behavior of `info::device::execution_capabilities` for non-OpenCL devices.  It turns out that the spec isn't clear about this.  This PR clarifies that an exception should be thrown, which is consistent with our behavior for `info::device::profile` and `info::device::preferred_interop_user_sync`, which are also inherited from OpenCL.  Another option would be to return an empty vector, so we could debate that if people think that's better.

I also think we should deprecate this query, which is consistent with the other queries that are specific to the OpenCL backend.  This PR deprecates the query in SYCL-NEXT, though I would also be open to retroactively deprecating it in SYCL 2020 with the rationale that we deprecated the other OpenCL-specific queries and simply forgot this one.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2748,12 +2748,19 @@ struct execution_capabilities {
 } // namespace sycl::info::device
 ----
 
+Deprecated by SYCL {SYCL_VERSION}.
+
 _Remarks:_ Template parameter to [api]#device::get_info#.
 
-_Returns:_ A [code]#std::vector# of the [api]#info::execution_capability# values
+_Returns:_ Only supported when the backend of this device is OpenCL (see
+<<chapter:opencl-backend>>).
+Returns a [code]#std::vector# of the [api]#info::execution_capability# values
 describing the supported execution capabilities.
-Note that this information is intended for OpenCL interoperability only as SYCL
-only supports [api]#info::execution_capability::exec_kernel#.
+Unless the device type is [api]#info::device_type::custom#, the returned vector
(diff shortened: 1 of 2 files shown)


gmlueck:
**NOTE** This PR is "stacked" on top of #601.  Once #601 is merged, I'll retarget this PR to the "main" branch.

TApplencourt (approved):

AerialMantis (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -2748,12 +2748,19 @@ struct execution_capabilities {
 } // namespace sycl::info::device
 ----
 
+Deprecated by SYCL {SYCL_VERSION}.
```
tomdeakin, line 2751: Deprecated in SYCL 2020

tomdeakin (changes requested):

adoc/chapters/programming_interface.adoc
gmlueck: Fixed in 0e4149691d11786bcfc58464e78600622713d78e and e49d70cd618420a5842d4de16ba8209289427584

tomdeakin (approved):

keryell (approved):

gmlueck:
I think this PR is waiting for Intel to modify the CTS.  To make my life easier, here is the internal Intel tracker for this work: CMPLRLLVM-61990

tomdeakin:
WG approved to merge.

Issue #532, opened 2024-01-18
Can "item" and "nd_item" be shared across work-items?


gmlueck:
This question recently came up from our implementation team.  What is the precise semantics of the member functions on `item` and `nd_item`?  Do they return the index of the *calling* work-item or do they return the index of the work-item that created the object?

The question is relevant only in the weird case where an `item` or `nd_item` object from one work-item is copied into SLM and then referenced by another work-item.  This could be done, for example, by using placement-new to copy the `item` or `nd_item` object.

Note that `id` may have a different semantic because it is user-constructible.  The semantics of `id` seem clear.  The member functions return the information that was used to construct the `id` object.  Therefore, if work-item _A_ copies its `id` object to SLM and then another work-item _B_ calls its member functions, I presume work-item _B_ would get the index of work-item _A_.

AlexeySachkov:
Linking #451 here. `sub_group` is probably different enough to be handled differently compared to `item` and `nd_item`, but we had the same question about it as well.

illuhad:
I don't think these classes are meaningful beyond outside of their own work item, because they generally carry the semantics of "give me the position of this item in the parallel iteration space". Using objects from a different work item should be illegal and/or UB.

Even just within one implementation the behavior might not be consistent. For example, whether AdaptiveCpp stores the position in the iteration space inside `nd_item`/`item` or whether it queries this on demand using builtins changes depending on compilation flow and configuration.

gmlueck:
I think this makes sense also.  I was wondering why we didn't define all the member functions of `nd_item` and `item` as static.  That would make it clear that they only return information about the calling work-item, since static member functions don't have access to any state in the object.  Did we decide not to define them as static to give implementations the freedom to store information about the calling work-item in the object?

In any case, I think we could add text to the introduction for each of these objects something like:

> The implementation constructs an `item` object for each work-item when a kernel is submitted with a range iteration space.  That object (and copies of that object) may be used only by the work-item that "owns" it.  Although it is possible to copy the object into memory that is accessible to other work-items or to the host, calls to member functions of the object from other work-items or from the host produce undefined behavior.

I think we could use similar language for `sub_group` and `group` also.

Pennycook:
[...]

PR #627, opened 2024-09-20
Unify `group` code snippet with other non-user-constructible classes

AlexeySachkov:
`group` is not user-constructible, but explicit information about its constructors was missing from the code snippet.

### DIFF
--- a/adoc/headers/group.h
+++ b/adoc/headers/group.h
@@ -10,6 +10,8 @@ template <int Dimensions = 1> class group {
   static constexpr int dimensions = Dimensions;
   static constexpr memory_scope fence_scope = memory_scope::work_group;
 
+  group() = delete;
+
   /* -- common interface members -- */
 
   id<Dimensions> get_group_id() const;


adoc/headers/group.h
```diff
@@ -10,6 +10,8 @@ template <int Dimensions = 1> class group {
   static constexpr int dimensions = Dimensions;
   static constexpr memory_scope fence_scope = memory_scope::work_group;
 
+  group() = delete;
```
gmlueck, line 13: I think this looks good, but should we do the same also for `sub_group`?

AlexeySachkov: Done in e41eda4b76077507f19e3d2cfcf700e877fd1f9d

gmlueck (approved):

gmlueck:
Adding the editorial label.  The spec was already clear that these types are not user constructible, so this PR just clarifies that in the synopsis.  I'll wait a few days before merging in case others have comments.

keryell:
Just curious about this implementation detail leakage: I have the feeling that "not user-constructible" does not mean it has no default constructor, just it cannot be constructed by the user. For example the constructor could be `private`.

gmlueck:
> Just curious about this implementation detail leakage: I have the feeling that "not user-constructible" does not mean it has no default constructor, just it cannot be constructed by the user. For example the constructor could be `private`.

This situation occurs in several places in the SYCL spec.  We want to document that one of the implicitly-defined member functions is not available to the application.  We currently shows these cases as `= delete` in the code synopsis.  You are correct, though, that an implementation could make these member functions unavailable in other ways, though.  How does the C++ specification handle cases like this?  Is there an example of a standard C++ class whose default constructor is not available?  How does the C++ specification illustrate this?

There is a similar situation with `std::atomic_ref`, where the copy-assignment operator is not available.  The [C++ specification](https://eel.is/c++draft/atomics.ref.generic#general) documents the operator as `= delete`.  Does this mean that an implementation must actually delete the operator, or could an implementation make the operator unavailable in some other way (such as declaring it `private`)?

Pennycook:
I agree with Ronan here.  We need to be precise about what "not user constructible" means, and I think it's different to saying that the default constructor is deleted.

I ran some experiments (see https://godbolt.org/z/d1h4o8Ge3), and the results are a little surprising.  clang doesn't complain about the construction of `b` below in C++17, but does in C++20.

```c++
struct S2 {
    S2() = delete;
};
S2 a; // this is invalid
S2 b{}; // this is (surprisingly) valid, even though the constructor is deleted

struct S3 {
    private:
    S3() {};
};
S3 c; // this is invalid
S3 d{}; // this is also invalid
```

Does anybody understand this behavior?

Regardless, I think that in order to accept `= delete` as the solution here, we need to convince ourselves: 1) that it is sufficient to prevent users from constructing `group` objects; and 2) that we want to forbid implementations from defining private default constructors for `group` objects.

nliber:
@Pennycook There was a core issue on this (I can't check at the moment because open-std.org is down).

The reason it compiles in C++17 is because `S2 b{};` is doing aggregate initialization and ignoring the default constructor.  This got fixed in C++20.

Pennycook:
> @Pennycook There was a core issue on this (I can't check at the moment because open-std.org is down).
> 
> The reason it compiles in C++17 is because `S2 b{};` is doing aggregate initialization and ignoring the default constructor. This got fixed in C++20.

Oof.  Does that mean that we can't use `= delete` to specify this in SYCL 2020, but we could in a future version of SYCL (if it adopted C++20 as the minimum language version)?

nliber:
I think we can use `= delete;`.  None of the real types are actually aggregates, are they?   Specifically, while the definition of aggregate has changed over the years, anything with a non-`public` non-`static` data member has never been an aggregate.

Pennycook:
> None of the real types are actually aggregates, are they?

I don't think they're required to be, but I think they currently _can_ be.

After the discussion in https://github.com/KhronosGroup/SYCL-Docs/issues/532, we reached the conclusion that classes like `nd_item` and `group` could legally be stateless (i.e., they don't need to store the local range, etc, but can query it directly with an intrinsic).  The [DPC++ implementation of `nd_item`](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/nd_item.hpp) is now stateless and has no data members.  I might be wrong, but I think it would be an aggregate if not for the fact that it declares a protected default constructor as an implementation detail.

> We might have to call out that these types are not aggregates.

Would we just say something like "These types must not be aggregates." and leave it up to individual implementations to decide how to enforce that?  I can imagine some implementations will want to rely on private/protected members, while others will want to rely on a private/protected constructor.

keryell:
> After the discussion in #532, we reached the conclusion that classes like `nd_item` and `group` could legally be stateless (i.e., they don't need to store the local range, etc, but can query it directly with an intrinsic). The [DPC++ implementation of `nd_item`](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/nd_item.hpp?rgh-link-date=2024-09-30T06%3A54%3A14Z) is now stateless and has no data members. I might be wrong, but I think it would be an aggregate if not for the fact that it declares a protected default constructor as an implementation detail.

Stateless means empty, which does not seem like an aggregate to me, since in C++ it will have a size of 1 anyway to enforce uniqueness of objects.

Otherwise, I was looking at some types in C++. For example https://en.cppreference.com/w/cpp/types/type_info describes the constructor as deleted but looking more carefully at https://eel.is/c++draft/type.info it is just that at least there is no exposed public constructor.

At the end, this is not super important. We could always have a public constructor `= delete` and have some other private constructor with some tag-dispatch.

nliber:
\> Would we just say something like "These types must not be aggregates." and leave it up to individual implementations to decide how to enforce that? 

I think so, especially since this is really just an issue for those compilers supporting C++17.  Any version of SYCL that requires C++20 or later will likely just delete the default constructor.

AlexeySachkov:
> Just curious about this implementation detail leakage: I have the feeling that "not user-constructible" does not mean it has no default constructor, just it cannot be constructed by the user. For example the constructor could be `private`.

Looking into [16.3.2.4 Detailed specifications](https://eel.is/c++draft/structure.specifications) it says the following:

> Descriptions of class member functions follow the order (as appropriate): [138](https://eel.is/c++draft/structure.specifications#footnote-138)
> ...
> [138)](https://eel.is/c++draft/structure.specifications#footnoteref-138) To save space, items that do not apply to a class are omitted[.](https://eel.is/c++draft/structure.specifications#footnote-138.sentence-1) For example, if a class does not specify any comparison operator functions, there will be no “Comparison operator functions” subclause[.](https://eel.is/c++draft/structure.specifications#footnote-138.sentence-2)

This makes me think that we should actually omit the description of any constructors to say that there aren't any (user-visible) constructors.

keryell:
> There is a similar situation with `std::atomic_ref`, where the copy-assignment operator is not available. The [C++ specification](https://eel.is/c++draft/atomics.ref.generic#general) documents the operator as `= delete`. Does this mean that an implementation must actually delete the operator, or could an implementation make the operator unavailable in some other way (such as declaring it `private`)?

`std::atomic_ref` is very different because it defines a reference wrapper, so a default constructor does not make sense, even internally.

tomdeakin:
WG suggests we change the spec to say not default-constructible and add comment to synopsis that default constructor is not available.

Pennycook:
> WG suggests we change the spec to say not default-constructible and add comment to synopsis that default constructor is not available.

Following up offline, as discussed...

I agree with this, but I still think we should go one step further and _also_ say either:
- `group` is not an aggregate; or
- `group` is not default-initializable.

There are good reasons to allow SYCL implementations to define things like private/protected constructors (because they need to make the object somehow!).  I don't think there is a good reason to allow SYCL implementations to (perhaps accidentally) implement `group` in a way that allows developers to create a `group` without going through one of the standard SYCL functions.

If `group g{}` compiles and works with one SYCL implementation, it is almost certain that somebody somewhere will come to rely on that.  See [Hyrum's Law](https://www.hyrumslaw.com/).  I understand this would likely only cause problems for that specific implementation and may therefore not be an issue with the specification per se, but since many of us are implementers too, I think it makes sense to close as many these gaps as we can.

etomzak:
I'm very interested in the conclusion to the non-user-constructible wording problem.

> I agree with this, but I still think we should go one step further and _also_ say either:
> 
>     * `group` is not an aggregate; or
> 
>     * `group` is not default-initializable.
 
"Not default-initializable" makes more sense to me. "Aggregate" is more like an implementation detail; "initializable" describes a user-facing interface.

> There are good reasons to allow SYCL implementations to define things like private/protected constructors (because they need to make the object somehow!). I don't think there is a good reason to allow SYCL implementations to (perhaps accidentally) implement `group` in a way that allows developers to create a `group` without going through one of the standard SYCL functions.

I'd take it one step further. If the SYCL spec literally contains the code `group() = delete;`, then that strongly suggests that a conformant implementation must also contain that code. IMO `group() = delete;` is a misdirect, because it doesn't prevent all user actions that the spec wants to disallow, but it does prevent some things that the spec wants an implementation to be able to do.
 
> If `group g{}` compiles and works with one SYCL implementation, it is almost certain that somebody somewhere will come to rely on that. See [Hyrum's Law](https://www.hyrumslaw.com/). I understand this would likely only cause problems for that specific implementation and may therefore not be an issue with the specification per se, but since many of us are implementers too, I think it makes sense to close as many these gaps as we can.

I think the key question is, what should happen when a user does `group g{}`? Is it UB? Is it ill-formed? Erroneous? Something gets thrown somewhere? Something else entirely? To avoid Hyrum's Law, ill-formed (with a diagnostic) might be the way to go. Perhaps that's all the SYCL spec needs to say, and it'll be up to the SYCL implementation to figure out how to emit an appropriate diagnostic.

keryell:
> I'm very interested in the conclusion to the non-user-constructible wording problem.
[...]
> I think the key question is, what should happen when a user does `group g{}`? Is it UB? Is it ill-formed? Erroneous? Something gets thrown somewhere? Something else entirely? To avoid Hyrum's Law, ill-formed (with a diagnostic) might be the way to go. Perhaps that's all the SYCL spec needs to say, and it'll be up to the SYCL implementation to figure out how to emit an appropriate diagnostic.

This should not compile at least.
By looking ahead at C++26 which allows
```c++
group() = delete("The SYCL 2032 specification does not allow default construction because we care a lot for your safety");
```
I am in favor of using `= delete()` since the aggregate initialization will be fixed too.
In the mean time...

Pennycook:
> My interpretation of this is that a SYCL implementation is not required to provide a default constructor or to allow `item` to be default initialized. Therefore, applications that do this are not conformant. I think that technically an implementation _could_ provide a default constructor (e.g. as an extension) if it wanted. Therefore, I think this spec wording is _not_ a mandate that an implementation _must_ diagnose an error if an application attempts to default-construct or default-initialize an `item`.

@gmlueck, I agreed with you up until this point.

I think a reasonable C++ developer would expect the phrasing "The `item` class is neither default constructible nor default initializable." to be equivalent to "`std::is_default_constructible_v<T>` is `false` and `T` does not satisfy `std::default_initializable`". Whether a class satisfies these traits/concepts is user-observable, and has observable side-effects: if we define another SYCL feature and say "it only accepts default-initializable types", we want that statement to forbid accepting the `group` type.  Allowing extensions to change statements like this seems like a bad idea, and I think if we add phrasing like this it should be backed up by CTS tests that ensure conformant implementations return the expected value(s) for the related traits.

etomzak:
> Whether a class satisfies these traits/concepts is user-observable, and has observable side-effects: if we define another SYCL feature and say "it only accepts default-initializable types", we want that statement to forbid accepting the `group` type.

Agree with this ^

> Allowing extensions to change statements like this seems like a bad idea, and I think if we add phrasing like this it should be backed up by CTS tests that ensure conformant implementations return the expected value(s) for the related traits.

I'm not sure what to make of this. Are you saying that there are some requirements that the spec places on an implementation that a vendor extension is allowed to override, and there are other requirements that a vendor extension is not allowed to override? How does (can?) the spec capture these two classes of requirements?

Pennycook:
> I'm not sure what to make of this. Are you saying that there are some requirements that the spec places on an implementation that a vendor extension is allowed to override, and there are other requirements that a vendor extension is not allowed to override? How does (can?) the spec capture these two classes of requirements?

To be honest, I haven't thought about this a lot until now, so I'm not sure how to express my opinion precisely. I think my feelings are roughly in line with the [extension guidelines](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#chapter.extensions), but I might not be interpreting them correctly.  They could be clearer.

`sycl::atomic_ref` defines the set of types that _can_ be used with the class template.  An extension that adds a new specialization of `sycl::atomic_ref` is overriding a restriction, but does so without changing "the definition of existing functions defined by the core SYCL specification in a way that changes their specified behavior".

The restrictions on device code are a weaker example.  The specification says that certain things (e.g., virtual functions) are "not permitted", but doesn't specify the behavior of virtual functions. My interpretation of the guidelines allows an extension to add support for virtual functions (overriding a restriction), without changing specified behavior. (If the specification had said that implementations _must_ fail to compile device code containing virtual functions, I think it would be reasonable to require an implementation to pass a CTS test checking that; an implementation would have to disable the extension to be conformant.)

I feel especially strongly when it comes to statements about type traits and concepts, because any change to those changes specified behavior. Whether we're asserting that a class must or must not satisfy a certain trait/concepts, I don't see any point specifying the behavior if extensions can override it.  We shouldn't encourage developers to use these traits/concepts if they're unreliable, because it might break overload resolution in some spectacular ways. If we want to give implementations wiggle room in a case like this, we should either say nothing or explicitly state that certain behavior is "not guaranteed".

gmlueck:
Perhaps this language captures my intent better, though it is more wordy.

> The SYCL specification does not provide a default constructor for `item` nor does it guarantee that `item` can be default initialized.  A SYCL implementation typically creates these objects when invoking a kernel, passing an `item` object as a parameter to the kernel function.
>
> [...]
>
> ```
> template <int Dimensions = 1, bool WithOffset = true> class item {
>  public:
>   // Default constructor not available
>   // ...
> };
> ```

Responding to @keryell: It does seem nice for users to require all implementations to `delete` the default constructor.  However, I thought we rejected this because it would prevent an implementation from defining the default constructor as a private member function.

keryell:
> Responding to @keryell: It does seem nice for users to require all implementations to `delete` the default constructor. However, I thought we rejected this because it would prevent an implementation from defining the default constructor as a private member function.

I thought to this but it did not worry me because you can use alternative solution like have constructors with useful parameter or just with tag-dispatch:
```c++
 class group {
    struct construct_tag {};
    public:
        group() = delete;
    private:
        group(construct_tag) {}
};
```

Pennycook:
> It does seem nice for users to require all implementations to delete the default constructor. However, I thought we rejected this because it would prevent an implementation from defining the default constructor as a private member function.

I think a lot of this came from me, so to clarify: I'm not opposed to deleting the default constructor, I just wanted to ensure that a decision was made with full understanding of the impact to implementations.  I agree with @keryell that existing implementations could be changed, if that's the direction we wanted to take.

gmlueck:
> I think a lot of this came from me, so to clarify: I'm not opposed to deleting the default constructor, I just wanted to ensure that a decision was made with full understanding of the impact to implementations. I agree with @keryell that existing implementations could be changed, if that's the direction we wanted to take.

I have a weak preference for this.  There are two ramifications I see:

* Implementation really would need to define the default constructor as deleted.  @AlexeySachkov, do you see any problem here from the DPC++ perspective?

* Implementations would be forbidden from providing the default constructor as an extension.

etomzak:
> I have a weak preference for this. There are two ramifications I see:
> 
>     * Implementation really would need to define the default constructor as deleted.  @AlexeySachkov, do you see any problem here from the DPC++ perspective?
> 
>     * Implementations would be forbidden from providing the default constructor as an extension.

I'm not opposed to `group() = delete`, but I think there's some cart-before-the-horse going on. There are still open questions:

1. How is *non-user-constructible* defined? (E.g., are implementations and extensions disallowed from exposing other constructors as well?)
2. To what extent is an implementation required to enforce *non-user-constructible*?

`group() = delete` could be the whole solution, or part of the solution, or unnecessary. We can't know without answering these two questions first.

nliber:
We should just delete the default constructor.  We are spending a lot of time and effort on what is ultimately a tiny little corner.

I'm less concerned about _non-user-constructible_.  IMO, it roughly means that there is no public constructor (or aggregate initialization) which doesn't take a reference to another object of the same type or a non-public / exposition only (tag) type.  This covers allowing copy and move constructors.

It is hard to formally define this, because `decltype` can get names to things we don't intend to publicly name.

If implementations/extensions want to make these types user constructible and not do so via a factory function, they can just ask the committee, can't they?  (It's a different story if we are breaking someone now, but I don't think we are.). If it comes up, we can either grant an exception or decide on a more formal definition on what is and isn't allowed.

nliber (approved):

Pennycook:
Following up after last week's WG discussion, here is my attempt at a concrete proposal for wording.

The current wording for `group` is:
> The `group<int Dimensions>` encapsulates all functionality required to represent a particular [work-group](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#work-group) within a parallel execution. It is not user-constructible.

I propose that we accept Alexey's `= delete()` change and modify this paragraph, borrowing some wording from the group interface KHR:
> The `group` class template encapsulates all functionality required to represent a specific [work-group](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#work-group) within a kernel.
>
> The set of work-items represented by an instance of the `group` class template is determined by the implementation, and there is subsequently no way for a user to construct arbitrary instances of the `group` class template. Implementations must ensure that instances of the `group` class template can only be acquired from a call to a standard SYCL function.
>
> [_Note_: If using C++17, implementations must ensure that the `group` class template is not an aggregate -- _end note_]

gmlueck:
> Implementations must ensure that instances of the `group` class template can only be acquired from a call to a standard SYCL function.

How about:

> Applications that attempt to construct a `group` object are ill formed, and the implementation must issue a diagnostic in this case.

We use similar wording in several other places in the spec.

etomzak:
> Applications that attempt to construct a group object are ill formed, and the implementation must issue a diagnostic in this case.

👍 👍 for this wording. It addresses my main concern. Minor nit: the C++ spec uses `ill-formed` (with a hyphen).

> I propose that we accept Alexey's `= delete()` change 

I'm not opposed, but I'm wondering if defining the default constructed as `private`, `// implementation-defined` would be closer to what we intend. It would have the same user-facing effect as `= delete`, but would allow implementations to default-construct if they really wanted to (e.g., with a static class function). This would align with how `handler` does it.

> The set of work-items represented by an instance of the `group` class template is determined by the implementation, and there is subsequently no way for a user to construct arbitrary instances of the `group` class template. 

*Consequently* instead of *subsequently*?

-----

More broadly, I do still think that the spec needs to clearly define *non-user-constructible* in the same way that it defines [§4.5.2. Common reference semantics](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics) and [§4.5.3. Common by-value semantics](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:byval-semantics). I.e., some text explaining what it is, a list of classes that belong to this category, and a code block. This would take us closer to resolving #807 as well. I'm less sure if this is the right PR for making that change.

tomdeakin:
> > I do still think that the spec needs to clearly define _non-user-constructible_
> 
> Good point. Revising my proposed text:
> 
> > Applications that attempt to default construct a group object are ill formed, and the implementation must issue a diagnostic in this case.
> 
> Regarding "ill formed" vs. "ill-formed" ... there are many uses of the term "ill formed" (without the hyphen in the spec already). If we want to adopt the hyphenated version, let's change them all.

I don't see this additional text in the changes here - are they still planned?

adoc/chapters/programming_interface.adoc
```diff
@@ -12728,16 +12728,21 @@ Then return the initial copy of the [code]#id#.
...
+
+The implementation constructs instances of the [code]#item# class template.
+Applications that attempt to default construct an [code]#item# object are ill
+formed, and the implementation must issue a diagnostic in this case.
+
+{note}If using {cpp17}, implementations must ensure that the [code]#item# class
+template is not an aggregate.
+{endnote}
+
+THe [code]#item# class template can optionally carry the offset of the range if
```
VerenaBeckham:
```suggestion
The [code]#item# class template can optionally carry the offset of the range if
```

gmlueck: Thanks for catching this!

PR #633, opened 2024-09-27
Add shorthand aliases for address spaces

Pennycook:
Working with address spaces (e.g., when using address space casts) was previously very verbose. Defining shorthands allows long values like `sycl::access::address_space::global_space` to be replaced by much shorter values like `sycl::global_space`.

### DIFF
--- a/adoc/headers/multipointer.h
+++ b/adoc/headers/multipointer.h
@@ -20,6 +20,12 @@ enum class decorated : /* unspecified */ {
 
 } // namespace access
 
+// Shorthand aliases for address spaces
+constexpr inline access::address_space global_space = access::address_space::global_space;
+constexpr inline access::address_space local_space = access::address_space::local_space;
+constexpr inline access::address_space private_space = access::address_space::private_space;
+constexpr inline access::address_space generic_space = access::address_space::generic_space;
+
 template <typename T> struct remove_decoration {
   using type = /* ... */;
 };
--- a/adoc/chapters/what_changed.adoc
+++ b/adoc/chapters/what_changed.adoc
@@ -9,6 +9,10 @@
   * The overload with a fallback/secondary queue parameter of the
     [code]#submit()# member function of [code]#sycl::queue# was deprecated.
 
+  * Added shorthand aliases for address spaces to reduce verbosity, enabling
+    (for example) [code]#sycl::access::address_space::global_space# to be
+    replaced with [code]#sycl::global_space#.
+
 == What has changed from SYCL 1.2.1 to SYCL 2020
 
 The SYCL runtime moved from namespace [code]#cl::sycl# provided by


Pennycook:
To kickstart the bikeshedding... @AerialMantis had suggested that we adopt different names here, that would be more aligned with names like `memory_order_relaxed` and `memory_scope_group`, something like `address_space_global`.

One way we could split the difference between `sycl::global_space` (19 characters) and `sycl::address_space_global` (27 characters) would be to define a new alias for `access::address_space` as well:

```c++
namespace sycl {
namespace access {
enum class address_space : int {
  global_space,
  local_space,
  private_space,
  generic_space
};
} // namespace access

using addrspace = access::address_space;
static constexpr addrspace addrspace_global = addrspace::global_space;
static constexpr addrspace addrspace_local = addrspace::local_space;
static constexpr addrspace addrspace_private = addrspace::private_space;
static constexpr addrspace addrspace_generic = addrspace::generic_space;

} // namespace sycl
```

...which would give us a shorthand of `sycl::addrspace_global` (23 characters).

Setting a precedent for use of `addrspace` instead of `address_space` might help with naming future features, too.  The proposed `static_address_cast` and `dynamic_address_cast` functions could be renamed `static_addrspace_cast` and `dynamic_addrspace_cast`, maintaining the link to address spaces (and the `enum`) while being only 2 characters longer.

TApplencourt:
I Agree with Gordon, I would like to keep the consistency of our mangling. I personally have no problem with `address_space_global` all fancy IDE will do the auto-complete (the problem was the `::` IMO always painful to navigate namespace).  Also we keep the  3 `_` for regularity with other API as extra bonus :p 

But I have no problem with `addrs` either.

 (even if  I'm still confused between `Q.copy, Q.cpy, Q.memcopy, Q.memcpy` and never know which one exists and the argument orders. End of my rant about alias)

Pennycook:
> (even if I'm still confused between `Q.copy, Q.cpy, Q.memcopy, Q.memcpy` and never know which one exists and the argument orders. End of my rant about alias)

I agree that having too many ways to do the same thing is bad.  The only reason I'm suggesting to make an exception here is simply to avoid breakage between SYCL 2020 and SYCL-Next.

Between SYCL 1.2.1 and SYCL 2020 we renamed `sycl::access::mode` to `sycl::access_mode` and deprecated the old name.  I think keeping the nested namespace `sycl::access::address_space` was a mistake, and if we'd caught this earlier we could have just had `sycl::addrspace` in SYCL 2020. My personal preference would still be to deprecate the old name and move everything to the new one... But I understand that this is an ABI/API concern, and so I could live with an alias. 😄

TApplencourt:
> My personal preference would still be to deprecate the old name and move everything to the new one... But I understand that this is an ABI/API concern, and so I could live with an alias

Agree! Should we deprecate it now, so we can remove it in SYCL Next?

Pennycook:
> Agree! Should we deprecate it now, so we can remove it in SYCL Next?

We should talk about this in a WG meeting.

I don't think we can deprecate things if we go the alias route.  Adding `[[deprecated]]` to `address_space` would cause a deprecation warning to trigger every time somebody used `addrspace` (because it's only an alias).

I believe our options are:

1) Define `addrspace` as an alias of `address_space`, but keep `address_space` around forever for legacy code.
2) Re-define `address_space` as an alias of `addrspace`, and deprecate `address_space` (ABI break)
3) Define `addrspace` as an alias of `address_space` for now, planning to swap them some day far in the future.

gmlueck:
There also a 4th option, which I'll throw out:

4. Define `addrspace` as a separate enumeration and add new APIs for functions that currently take `address_space`.

This would allow us to deprecate and eventually remove `address_space` and any API that uses it without incurring an ABI break.

Many of the uses of `address_space` are from `multi_ptr`, which we may decide to deprecate and replace anyways.  When I noticed this, I was thinking that option (4) would be a good path forward.  However, then I realized that we also use `address_space` extensively in `atomic_ref`, so option (4) seemed less ideal.

keryell:
I am thinking to `local_as`, `constant_as`, `private_as`, `generic_as`.

`as_local` sounds more like a cast in LLVM/MLIR style.

Pennycook:
> I am thinking to `local_as`, `constant_as`, `private_as`, `generic_as`.
> 
> `as_local` sounds more like a cast in LLVM/MLIR style.

I do like that these are much shorter, but I think that "as" might be _too_ short.  Like you pointed out, it's very easy to read it as the English word "as" instead of as an initialism. Also, because we don't have any precedent for this elsewhere in SYCL, a new developer probably won't be able to guess what it means without looking at the specification.

keryell:
> > I am thinking to `local_as`, `constant_as`, `private_as`, `generic_as`.
> > `as_local` sounds more like a cast in LLVM/MLIR style.
> 
> I do like that these are much shorter, but I think that "as" might be _too_ short. Like you pointed out, it's very easy to read it as the English word "as" instead of as an initialism. Also, because we don't have any precedent for this elsewhere in SYCL, a new developer probably won't be able to guess what it means without looking at the specification.

Any reasonable developer not using punch cards will move the mouse on the word and the explanation on what is the meaning of `private_as` will pop up on screen thanks to LSP mode.
Is there any real usage of this with a code sample to see how confusing it would be?

Pennycook:
I've updated the PR to reflect the discussion in today's WG meeting.

TApplencourt (approved):

adoc/headers/multipointer.h
```diff
@@ -21,10 +21,11 @@ enum class decorated : /* unspecified */ {
...
 // Shorthand aliases for address spaces
-constexpr inline access::address_space global_space = access::address_space::global_space;
-constexpr inline access::address_space local_space = access::address_space::local_space;
-constexpr inline access::address_space private_space = access::address_space::private_space;
-constexpr inline access::address_space generic_space = access::address_space::generic_space;
+using addrspace = access::address_space;
+constexpr inline addrspace addrspace_global = addrspace::global_space;
+constexpr inline addrspace addrspace_local = addrspace::local_space;
+constexpr inline addrspace addrspace_private = addrspace::private_space;
+constexpr inline addrspace addrspace_generic = addrspace::generic_space;
```
keryell, line 28: Since this is no longer constrained by namespace hierarchy, why not following English like
`constexpr inline addrspace generic_addrspace = addrspace::generic_space;`
?

Pennycook: @AerialMantis' suggestion was that we should always use `enum_value`, because that's the precedent elsewhere (e.g., `memory_order_relaxed`, `memory_scope_device`).

keryell: But at the end I would prefer
```c++
constexpr inline addrspace global_as = addrspace::global_space;
constexpr inline addrspace local_as = addrspace::local_space;
constexpr inline addrspace private_as = addrspace::private_space;
constexpr inline addrspace generic_as = addrspace::generic_space;
```

gmlueck (approved):

tomdeakin (approved):

adoc/headers/multipointer.h
keryell: > @AerialMantis' suggestion was that we should always use `enum_value`, because that's the precedent elsewhere (e.g., `memory_order_relaxed`, `memory_scope_device`).

Does this impact this PR then?

Pennycook: Yes, this is why we changed the name from `global_space` to `addrspace_global`.

If we were following established precedent very strictly we would have called the shorthand `addrspace_global_space`, but people didn't like the redundant `_space` at the end.  We had previously discussed renaming the `enum` values as well to avoid the `_space` suffix (see Khronos issue 654) but consensus was that renaming the address spaces was a step too far.

tomdeakin:
Awaiting CTS

keryell (approved):
Thanks!

gmlueck:
For my own benefit, this is the Intel tracker for adding CTS tests: CMPLRLLVM-65339

PR #634, opened 2024-09-30
Redefine SYCL_LANGUAGE_VERSION

gmlueck:
Now that we have a unified specification, we need to decouple the `SYCL_LANGUAGE_VERSION` macro from the document revision.  We decided to keep the 6 digit format where the top 4 digits are the year of the SYCL specification.  The lower 2 digits are not formally specified, but we plan to use these to identify the month in which we submit the specification for ratification, which is similar to the C++ macro `__cplusplus`.

Since the SYCL 2020 specification was submitted for ratification in December of 2020, the macro's value is now `202012` for SYCL 2020.

We also decided to add the `L` literal suffix to make the macro into a long integer literal.  This makes sense because an `int` is only guaranteed to be 16 bits, which is not enough to hold these 6-digit values.  This is consistent with `__cplusplus`, which is also a long integer literal.


TApplencourt (approved):

tomdeakin (approved):
LGTM, thanks Greg!

keryell (approved):

gmlueck:
Can this me merged?  A CTS test was added in KhronosGroup/SYCL-CTS#989

tomdeakin:
WG approved to merge

PR #638, opened 2024-10-04
Add sycl_khr_group_interface extension

Pennycook:
This extension introduces an alternative interface for groups of work-items, offering several improvements over the SYCL 2020 interface:

- Shorter names for member functions, dropping the get_ prefix.

- Cleaner separation between properties of a group (e.g., a group id) and properties of the calling work-item (e.g., its id within a group).

- Clearer distinction between "group" concept and "work_group" class.

- New work_item class to represent a single work-item within a specific parent group. This class also satisfies the group concept, modeling a group containing a single work-item.


TApplencourt:
I like the usage of namespace.
```
    khr::work_group<1> g = ndit.get_group();
    khr::work_item it = get_item(g);
 ```
 Few questions (sorry, trying to understand this new khr mechanism), should it be :
 - ` khr::group_interface::work_item`
 - ` khr_group_interface_get_group`?
 - ` khr::group_interface::get_item`?

**Comment by Pennycook** [comment]
> I like the usage of namespace.
> ```
>     khr::work_group<1> g = ndit.get_group();
>     khr::work_item it = get_item(g);
>  ```
>  Few questions (sorry, trying to understand this new khr mechanism), should it be :
>  - ` khr::group_interface::work_item`
>  - `ndit.khr_group_interface_get_group`? (or just `ndit.khr_get_group()`?) 
>  - ` khr::group_interface::get_item`?

I don't think so.  When vendors introduce extensions they use `sycl::ext::vendor::`, not `sycl::ext::vendor::extension_name`.

The way I see it, although putting something in the `khr::` namespace reserves a name, the SYCL WG is in complete control over any names that are introduced there.  Future extensions might have to be more creative with their naming, but there's no risk of surprise conflicts.

**Review by TApplencourt — APPROVED** [review]

**Review by keryell — APPROVED** [review]
That looks good!

**Comment by keryell** [comment]
Adding the 2 rank functions from `mdspan` too? https://gitlab.khronos.org/sycl/Specification/-/issues/630

**Comment by Pennycook** [comment]
> Adding the 2 rank functions from `mdspan` too? https://gitlab.khronos.org/sycl/Specification/-/issues/630

Good idea.  I've added `rank`, `rank_dynamic` and `static_extent` in 008ac23565df212244a692bd01711766f8433000.

File: adoc/extensions/sycl_khr_group_interface.adoc [hunk 1]
```diff
@@ -0,0 +1,479 @@
...
+invariant for the lifetime of the group.
+The leader of the group is guaranteed to be the work-item with index 0 within
+the group.
+
+[[sec:khr-group-interface-work_group]]
+== [code]#work_group# class
+
+The [code]#work_group# class template encapsulates all functionality required to
+represent a particular <<work-group>> within a kernel.
+It is not user-constructible.
```
gmlueck, line 226: It is user constructible from a `group`.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+
+  /* -- common group interface members -- */
+
+};
+
+} // namespace khr
+} // namespace sycl
+----
+
+.[apidef]#work_group constructor#
```
gmlueck, line 265:
```suggestion
.[apititle]#work_group constructor#
```

Use `[apititle]` when the text is not a C++ identifier.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+The SYCL [code]#sub_group# class template provides common by-value semantics
+(see <<sec:byval-semantics>>) and the common group interface (see
+<<sec:khr-group-interface-common>>).
+
+[source,role=synopsis]
+----
+namespace sycl {
+namespace khr {
+
+class sub_group {
```
gmlueck, line 303: What is your naming plan if this KHR is eventually adopted into the core spec?  We already have a type named `sub_group`.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+----
+operator sycl::sub_group() const noexcept;
+----
+
+_Returns_: A [code]#sycl::sub_group# representing the same collection of
+work-items as this [code]#sub_group#.
+
+'''
+
+.[apidef]#max_size#
```
gmlueck, line 354:
```suggestion
.[apidef]#khr::sub_group::max_size#
```

Use the fully qualified name here (without `sycl`).  Otherwise, there will be a conflict with other identifiers named `max_size` in other classes.

Same for the other `[apidef]` in this KHR.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+and less verbose than the interface provided by `sycl::group` and
+`sycl::sub_group` in SYCL 2020.
+
+[[sec:khr-group-interface-dependencies]]
+== Dependencies
+
+This extension has no dependencies on other extensions.
+
+Some features of this extension are only available when an implementation
+provides `<mdspan>`.
```
gmlueck, line 15:
```suggestion
provides `<mdspan>`, which is defined in C++23.
```

Seems like this might be helpful.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+class GroupT {
+
+ public:
+  using id_type = /* ... */;
+  using linear_id_type = /* ... */;
+  using range_type = /* ... */;
+#if defined(__cpp_lib_mdspan)
+  using extents_type = /* ... */;
+  using index_type = typename extents_type::index_type; // exposition only
+  using rank_type = typename extents_type::rank_type; // exposition only
```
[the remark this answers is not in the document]
Pennycook: I declared these as exposition only because we don't want them to be exposed by classes implementing the group interface, but they simplified the synopsis.  The alternative is writing `extents_type::index_type` everywhere it appears; I've made this change in dd2110a6f71f9ca00d688296dc6f7f3dcaf0bfb8.

 I don't want to assume `std::dextents`, because we already know that future group types will use different `extents_type`s, and I'm worried we'll introduce a specification bug by forgetting to revisit this.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+== Common group interface
+
+The `khr::work_group`, `khr::sub_group` and `khr::work_item` objects defined by
+this extension all implement a common set of operations.
+
+[source,role=synopsis]
+----
+
+namespace sycl {
+namespace khr {
```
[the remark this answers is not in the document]
Pennycook: Fixed all occurrences of this in da495737245c552706be1450c5c127bb645ae81d.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+
+'''
+
+.[apidef]#GroupT::rank#
+[source,role=synopsis,id=api:khr-group-interface-common-group-rank]
+----
+static constexpr rank_type rank() noexcept;
+----
+
+_Effects_: Equivalent to [code]#return extents_type::rank();#.
```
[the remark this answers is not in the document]
Pennycook: ISO C++ includes the `return` (e.g., see https://eel.is/c++draft/views.multidim#mdspan.layout.left.obs-9).

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+==== Non-member functions
+
+.[apidef]#get_item#
+[source,role=synopsis,id=api:common-group-get-item]
+----
+template <typename Group>
+work_item<Group> get_item(Group g) noexcept;
+----
+
+_Constraints_: Available only if `Group` is `work_group` or `sub_group`.
```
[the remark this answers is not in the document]
Pennycook: Fixed in 64e45c5c5947649fa66711bf1c3a9c30e6bd0895.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: The plan is to declare equivalents of all the members of `sycl::khr::sub_group` inside of `sycl::sub_group`, and deprecate the existing `sycl::sub_group` interface.

I appreciate this is an awkward way to go about it, but this KHR cannot add the necessary functions directly to `sycl::sub_group`.  Our naming convention dictates that any such functions be called things like `sycl::sub_group::khr_id()` -- in addition to being different from what we're actually trying to propose, it would be different to the interface exposed by `sycl::khr::work_group` and `sycl::khr::work_item`.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,479 @@
...
+=== Member functions
+
+.[apidef]#GroupT::id#
+[source,role=synopsis,id=api:khr-group-interface-common-group-id]
+----
+id_type id() const noexcept;
+----
+
+_Returns_: The index of this group within the index space returned by
+[code]#range()#.
```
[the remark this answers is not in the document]
Pennycook: Oh, thanks for explaining that.  I didn't understand how the `[api]` stuff all fit together.  Fixed in de356628056c3d1034fcde2e2be59f9c4ebf3ecc.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Fixed as part of de356628056c3d1034fcde2e2be59f9c4ebf3ecc.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Fixed as part of de356628056c3d1034fcde2e2be59f9c4ebf3ecc.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: How about something like:

> The `work_group` class template encapsulates all functionality required to represent a specific work-group within a kernel.
> The set of work-items represented by an instance of the `work_group` class template is determined by the implementation, and there is subsequently no way for a user to construct arbitrary instances of the `work_group` class template.
> Instances of the `work_group` class template can only be acquired from a call to a standard SYCL function, or by converting an instance of the `group` class template.

I think this is the intent behind the "not user-constructible" wording in the SYCL 2020 specification, but I agree that intent is lost when we introduce something like a conversion constructor.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,474 @@
...
+and less verbose than the interface provided by `sycl::group` and
+`sycl::sub_group` in SYCL 2020.
+
+[[sec:khr-group-interface-dependencies]]
+== Dependencies
+
+This extension has no dependencies on other extensions.
+
+Some features of this extension are only available when an implementation
+provides `<mdspan>`, which is defined in C++23.
```
gmlueck, line 15: Can you adjust this PR to use the wording we recently agreed to in #667

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Implemented in 3a6e6e2913b96b8b6674dbabb303e38230b27a69, with similar changes for `khr::work_group`, `khr::sub_group` and `khr::work_item`.

gmlueck (approved):

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,495 @@
...
+
+constexpr size_t N = 1024;
+constexpr size_t M = 256;
+
+int main() {
+
+  queue q;
+
+  int* in = malloc_shared<int>(N * M, q);
+  int* out = malloc_shared<int>(N, q);
```
keryell, line 450: Is this code UB because there is no `free`? :thinking:

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,495 @@
...
+== Example
+
+The example below demonstrates the usage of this extension.
+
+[source,,linenums]
+----
+#include <sycl/sycl.hpp>
+#include <iostream>
+#include <numeric>
+#include <algorithm>
```
keryell, line 439: Sort?

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Fixed in f4300546d20b7a65378d91b4e0737187d364d6bf.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Fixed in a68b83210c0783d0fa80fea8ac71ee0db7038a50.

Pennycook:
In offline discussion with @AerialMantis, @gmlueck, and others, we reached the conclusion that using the name `work_item` to represent both the concept of a work-item and a work-item's position inside another group was confusing. 29ec29553bd7bdea20277f4fb526973bd30ffc6e renames the class to `member_item`, in an attempt to disambiguate.

One thing we didn't discuss -- that only became apparent to me after making the change -- is whether people are still happy with the name `get_item`.  We could now alternatively call the `sycl::member_item` constructor (but this wouldn't work with ADL), or could rename the function something like `get_member` or `get_member_item` (but both of these are more verbose).

I don't feel strongly either way, but figured we should probably discuss the names one last time before we try and merge this in.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+
+class __group__ {
+
+ public:
+  using id_type = /* ... */;
+  using linear_id_type = /* ... */;
+  using range_type = /* ... */;
+  using extents_type = /* ... */; // C++23
+  using size_type = /* ... */;
+  static constexpr int dimensions = /* ... */;
```
AerialMantis, line 63: If we're looking to align with ISO C++ should we consider calling this `rank` instead?

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+  static constexpr memory_scope fence_scope = /* ... */;
+
+  id_type id() const noexcept;
+  linear_id_type linear_id() const noexcept;
+
+  range_type range() const noexcept;
```
AerialMantis, line 69: There's no `linear_range()` query, is this replaced with `size()`?

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+
+.[apidef]#+__group__::extents+#
+[source,role=synopsis,id=api:khr-group-interface-common-group-extents]
+----
+constexpr extents_type extents() const noexcept;
+----
+
+_Minimum C++ Version_: {cpp23}
+
+_Returns_: The number of work-items in each dimension of the group.
```
AerialMantis, line 136: Am I correct in thinking that the value returned by `extents()` equivalent to `range()` but as an extents type?

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+=== Non-member functions
+
+.[apidef]#khr::get_item#
+[source,role=synopsis,id=api:common-group-get-item]
+----
+template <typename Group>
+member_item<Group> get_item(Group g) noexcept;
+----
+
+_Constraints_: `Group` is `work_group` or `sub_group`.
```
AerialMantis, line 211: I wonder if we should base this constrain on a type trait so that we can support additional group types with `get_item` in the future without having to modify the definition of this API.

AerialMantis (commented):
Looks great! Just a few minor suggestions.

AerialMantis:
> One thing we didn't discuss -- that only became apparent to me after making the change -- is whether people are still happy with the name `get_item`. We could now alternatively call the `sycl::member_item` constructor (but this wouldn't work with ADL), or could rename the function something like `get_member` or `get_member_item` (but both of these are more verbose).

I was thinking about this when I was reviewing this again just now, we could potentially change this to `get_member_item`, it would avoid any confusion with the current `item` class but it's a bit more wordy. I don't have a strong opinion on this either, but maybe something to discuss with the rest of the working group.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: There is a `static constexpr rank()` function that's available starting with C++23.

`dimensions` is already part of the SYCL 2020 group interface (in the sense that it's defined by `group` and `sub_group`). 
 I assumed we wouldn't want to remove that any time soon, so made it part of this new interface as well.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: No. `range()` returns the range that the `id` is drawn from (i.e., the number of groups), whereas `size()` returns the size of the group (i.e., the number of work-items).

We didn't add `linear_range()` because it's shorter to write `range.size()`.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: No: `extents()` is the (multi-dimensional) size of the group, whereas `range()` is the number of groups.

The way to remember this (and also what was discussed in [this comment](https://github.com/KhronosGroup/SYCL-Docs/pull/638#discussion_r1973906024)) is that the goal of this redesign is to clarify which properties belong to the group.  So:
- `id` is the index _of the group_
- `range` is the range _of the group's indices_ (like how `item.get_range()` returns the range of an item's indices)
- `extents` is the extents _of the group_
- `size` is the size _of the group_

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+
+  extents_type extents() const noexcept; // C++23
+  extents_type::index_type extent(extents_type::rank_type r) const noexcept; // C++23
+
+  static constexpr extents_type::rank_type rank() noexcept; // C++23
+  static constexpr extents_type::rank_type rank_dynamic() noexcept; // C++23
+  static constexpr size_t static_extent(rank_type r) noexcept; // C++23
+
+  size_type size() const noexcept;
+  size_type max_size() const noexcept;
```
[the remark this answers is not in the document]
Pennycook: > Perhaps we could introduce max_size to the generic __group__ interface, though it might be hard to define what this value should return for other group types like work_group. 

This is what we have in SYCL 2020, and feedback from users has been that it's too confusing -- mainly because sub-groups are too confusing. ☹️ 

`sub_group::max_size()` returns the maximum compile-time size of a sub-group, and there might not be any sub-groups with this size.  `group::max_size()` returns the same thing as `size()`.  The functions don't behave the same way and you really can't write general code with them.

Making `max_size()` a sub-group only thing means that people will only try to use it for sub-groups, which hopefully will prevent people from using it without understanding what it means.

> Alternatively, maybe we could have a trait such as `__group__::has_max_size_v` or `__group__::is_sub_group`, and the `max_group()` function could be contained on this value being true.

I don't think we should do this, because it's easy for developers to write these themselves. `has_max_size_v` can be implemented using the detection idiom, and `is_sub_group` can be implemented using `std::is_same`.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,849 @@
...
+'''
+
+.[apidef]#khr::leader_of#
+[source,role=synopsis,id=api:common-group-leader_of]
+----
+template <typename Group>
+bool leader_of(Group g) noexcept;
+----
+
+_Returns_: [code]#true# if the calling work-item is the leader of group
```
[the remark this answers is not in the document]
Pennycook: Good catch! Fixed in 55a2ec402b48c0753ed1def3227a31a050bbafdd.

Pennycook:
> I was thinking about this when I was reviewing this again just now, we could potentially change this to `get_member_item`, it would avoid any confusion with the current `item` class but it's a bit more wordy. I don't have a strong opinion on this either, but maybe something to discuss with the rest of the working group.

I've requested to put this on the agenda for next week's meeting.

tomdeakin:
WG discussed, and converged on `get_member_item`.

Pennycook:
CTS tests are implemented in https://github.com/KhronosGroup/SYCL-CTS/pull/1024. 

The DPC++ implementation that passes the CTS tests is under review in https://github.com/intel/llvm/pull/17595.

gmlueck:
@illuhad : I think we are waiting for your input on this.  Do you agree that this is implementable?

PeterTh:
I've [prototyped this extension in SimSYCL](https://github.com/celerity/SimSYCL/pull/42) and I've noticed some most likely unintended behaviour in GCC -- see this minimal example: https://godbolt.org/z/zosojE1zG

Curiously, Clang and MSVC completely ignore this issue. I wasn't able to find exactly what GCC is basing this decision on in a quick search of the current C++ standard, but on the other hand, I think it's pretty clear that this isn't great (and even if it was, GCC is an important compiler target).

As such, I think it might be better if the relevant parts of the code in the spec were adjusted to nip this problem in the bud. This just requires a pretty simple spelling change of types in 4 lines of code, i.e. `using id_type = sycl::id<Dimensions>;` instead of `using id_type = id<Dimensions>;`, and the same for `range` ([see here](https://github.com/celerity/SimSYCL/blob/95238df3119dc140bc90b6b0df26c1b5521c236f/include/simsycl/extensions/khr_group_interface.hh#L49)).

tomdeakin:
WG approved to merge

adoc/extensions/sycl_khr_group_interface.adoc
AerialMantis: Yeah that's a good point, we don't want to break with the existing interface, though perhaps we could have an additional value rank which is equal to `dimensions`, and then we could look to deprecate `dimensions` in the future.

Though saying that, this is not really an issue, just that it would be nice to align with ISO C++ naming, so it may not be worth it.

adoc/extensions/sycl_khr_group_interface.adoc
AerialMantis: That makes sense thanks.

adoc/extensions/sycl_khr_group_interface.adoc
```diff
@@ -0,0 +1,852 @@
...
+  constexpr static extents_type::rank_type rank() noexcept; // C++23
+  constexpr static extents_type::rank_type rank_dynamic() noexcept; // C++23
+  constexpr static size_t static_extent(rank_type r) noexcept; // C++23
+
+  size_type size() const noexcept;
+
+};
+
+template <typename Group>
+member_item<Group> get_member_item(Group g) noexcept;
```
gmlueck, line 83: Should this instead be:

```suggestion
member_item<Group> get_member_item(const Group& g) noexcept;
```

Likewise for all the other functions in this KHR that take a `Group`, `group`, or `sub_group` parameter by value.

gmlueck (commented):
Sorry for the late comment.  Our recent discussion about passing `queue` by value vs. by reference got me thinking about similar issues with this KHR.

adoc/extensions/sycl_khr_group_interface.adoc
AerialMantis: So I agree we will likely end up requiring multiple traits, so we may not want this to be a generic group trait, as we may introduce a group type in the future which we don't want to support `get_item` for.

Though I'm a little concerned about this approach, as it effectively means introducing new overloads for new group types which would increase the complexity, for example if we introduce `get_item` for `work_group` and `sub_group` now and then later introduce `root_group`, `cluster_group`, etc, we would end up with something like:

```
template <typename Group, typename = std::enable_if_t<std::is_same_v<Group, work_group> || std::is_same_v<Group, sub_group>>::type>
void get_item(Group g) noexcept {}

void get_item(root_group g) noexcept {}

void get_item(cluster_group g) noexcept {}
```

Where as if we had a trait in this extension, something like `is_item_indexable` (not a great name but I'm sure we could think of something better), then we could support additional types by simply making that group type conform to `is_item_indexable`. I feel like this would be cleaner.

```
template <typename Group, typename = std::enable_if_t<is_item_indexable<Group>>::type>
void get_item(Group g) noexcept {}
```

Saying that I don't want to hold up this PR, so perhaps we could consider this as a follow up change.

adoc/extensions/sycl_khr_group_interface.adoc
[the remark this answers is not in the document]
AerialMantis: Yeah my thinking was that we provide a trait so that users can write code which only uses this function if the group is a `sub_group`, but you're right the user could implement this themselves, either via their own type trait or by having separate overloads for different group types.

AerialMantis (approved):
Looks good! I still have a couple of comments open, but neither of them are blockers to merging the PR.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: Exactly.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: We're on the same page.  The original presentation that proposed that we should have a group concept called `indexable_item_group`, so `is_item_indexable` is aligned with that proposal.

I think we should defer this conversation until a later date, though, perhaps even when we discuss the possibility of this feature becoming core.  Whether we'd encode this as a concept or a set of traits will be dependent on whether we can assume C++20 as a minimum version, and that will depend on which version of SYCL this intercepts.

I also want to point out that the specification is deliberately vague about how this should be implemented, saying only:
```
_Constraints_: [code]#Group# is [code]#work_group# or [code]#sub_group#.
```

If we don't like the idea of `root_group` adding a new overload of `get_item`, we could just add a relaxation that says something to the effect of: "If `root_group` is available, the _Constraints_ of `get_member_item` are modified to permit `root_group`".  So I think we can achieve what you want to without exposing a user-facing trait.

adoc/extensions/sycl_khr_group_interface.adoc
Pennycook: I'm not opposed to this direction, but one thing we should give some thought to is if we can also fix the existing group algorithms that take a group by value.

Having overloads for both `Group g` and `const Group& g` defined for the same function would lead to ambiguity, so I think the only way to fix this would be to say that the functions in SYCL 2020 take `Group` and the same functions in SYCL-Next take `const Group&`.  It wouldn't make sense to deprecate these functions in between, because we're not actually removing anything.

I think all implementations of the group algorithms are header-only because they accept user-defined types, so this shouldn't be an ABI break concern.  I also think that the vast majority of user could would continue to work, because `const Group&` is less restrictive than `Group` (but I'm not prepared to rule out that there might be some weird corner cases that require some adjustment).

Could we make this change in SYCL-Next, or are we stuck?

gmlueck: Yes, I think we need to reexamine all the APIs in the core SYCL spec that take objects by value.  It probably does make sense to change them in SYCL-Next even if it does cause some weird corner cases to break.  If there are cases like this, users will need to adjust in order to compile applications in SYCL-Next mode.

I'm much more concerned about retaining ABI compatibility between SYCL 2020 and SYCL-Next.  It would be a major pain if vendors needed to ship two versions of every library that uses SYCL.

gmlueck:
> WG approved to merge

@tomdeakin: I did not merge this because 985a67a08486b7c5aaf4e2101d1da60eb76b93a0 was made after the WG approved.  If you would like this merged prior to the next WG meeting, please add another comment saying so.  Otherwise, we can call for it to be merged again at the next meeting.

PR #645, opened 2024-10-22
Include cp. (compare) in set of non-line-ending abbreviations

etomzak:
Without this change, writing

```
Cp. <<relevant-part-of-spec>>
```

is formatted to

```
Cp.
<<relevant-part-of-spec>>
```


gmlueck:
Do we use "cp." as an abbreviation currently in the SYCL specification?  I wonder if we should just write out "compare" instead.

etomzak:
Consulting the ever-helpful [APA](https://apastyle.apa.org/style-grammar-guidelines/abbreviations/latin), *cf.* would be more correct than *cp.* I can make that change in the PR.

As to the use of an abbreviation meaning *compare*, since the spec already uses *i.e.*, *e.g.,* and *etc.*, for consistency, I think *cf.* should be available for use (and it shouldn't cause surprising newlines to be inserted by `reflow.py` when it's used).

I haven't found any use of *cp.* (or *cf.*), but I can think of a few places where I'd like to use it.

gmlueck (approved):

adoc/scripts/reflow.py
```diff
@@ -53,7 +53,7 @@
 # A single letter followed by a period, typically a middle initial.
 endInitial = re.compile(r'^[A-Z]\.$')
 # An abbreviation, which does not (usually) end a line.
-endAbbrev = re.compile(r'(e\.g|i\.e|c\.f|\bvs\b|\bco\b|\bltd\b|\bch\b)\.$', re.IGNORECASE)
+endAbbrev = re.compile(r'(e\.g|i\.e|c\.f|\bvs\b|\bco\b|\bltd\b|\bch\b|\bcf\b)\.$', re.IGNORECASE)
```
gmlueck, line 56: We already have "c.f." in this list, but it has a period between the "c" and "f".  I think your spelling is the correct one, so can you remove the old "c.f"?  I did a quick search of the spec, and I did not see any occurrences of "c.f.".

gmlueck (commented):
Please also update the PR description -- it no longer matches this PR.

adoc/scripts/reflow.py
etomzak: :facepalm: I double-checked the abbreviations earlier and still missed it. I've removed the version with the extra `.`.

gmlueck (approved):

gmlueck:
Merging as editorial.

PR #648, opened 2024-10-24
Update function objects section to the new format

### LINKED ISSUE #646: Return types of `logical_and` and `logical_or` don't match ISO C++
SYCL defines these function objects as:

```c++
template <typename T = void> struct logical_and {
  T operator()(const T& x, const T& y) const;
};

template <typename T = void> struct logical_or {
  T operator()(const T& x, const T& y) const;
};
```

Whereas [ISO C++](https://eel.is/c++draft/logical.operations) defines them both as returning `bool`.  `bool` makes a lot more sense to me than `T`, because the operators are defined as returning `x && y` and `x || y`.

I suspect this is a bug that arose from copying and pasting other function object definitions, but before writing up the fix I wanted to check whether anybody knew of a reason that things should be defined this way.

Pennycook:
Also fixes a bug in the return type of logical_and and logical_or. Closes #646.

The return types of the `void` specializations of `maximum` and `minimum` may require further discussion.  Prior to this change, the specification didn't say exactly which operator(s) would be used to determine ordering, and so I tried to honor that in this new wording.

When we introduced the prior wording there was some discussion of trying to align with `std::min` and `std::max` (see https://github.com/KhronosGroup/SYCL-Docs/pull/285#issuecomment-1262485753).  It we want to describe the behavior in terms of equivalent statements, I think the fix is to say:
- `minimum` and `maximum` require `T` to be `LessThanComparable`.
- `minimum` returns `(b < a) ? b : a`.
- `maximum` returns `(a < b) ? b : a`.


Pennycook:
In the WG meeting today we decided that we _should_ define the return types of the `minimum` and `maximum` operators.  Because the `decltype()` only describes the return _type_ of the operator, and doesn't actually constrain the implementation of the operator (i.e., it doesn't force an implementation to use a ternary).  I'll make this change soon.

Pennycook:
This should be read to review, now.  Note that the Cpp17LessThanComparable requirement is a _precondition_ and not a _constraint_.  This was surprising to me, but I wanted to follow the ISO C++ definition: https://eel.is/c++draft/alg.min.max

adoc/chapters/programming_interface.adoc
```diff
@@ -21277,142 +21277,273 @@ requirements for both host and device.
...
-@ Member function @ Description
-a@
-[source]
+'''
+
+.[apititle]#call operator#
+[source,role=synopsis,id=api:plus-call-operator]
 ----
-T operator()(const T& x, const T& y) const
+T operator()(const T& x, const T& y) const;
```
TApplencourt, line 21293: Good catch!

TApplencourt:
Not in the PR, but if you don't mind to a sneaky update (don't tell our editor :p ) :

```
float cospi(float x)                (1)
double cospi(double x)              (2)
half cospi(half x)                  (3)

template<typename NonScalar>         (4)
/*return-type*/ cospi(NonScalar x)
```

This guy (4) is not aligned.

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -21277,142 +21277,273 @@ requirements for both host and device.
...
-a@
-[source]
+'''
+
+.[apititle]#transparent call operator#
+[source,role=synopsis,id=api:plus-transparent-call-operator]
 ----
-T operator()(const T& x, const T& y) const
+template <class T, class U> constexpr auto operator()(T&& t, U&& u) const
+  -> decltype(std::forward<T>(t) + std::forward<U>(u));
```
keryell, line 21304: Is there a motivation to have the `std::forward` in the non-evaluated context?
Actually a maniac programmer could define `+` with an l-reference to do division and a `+` with an r-reference to do a multiplication?

Pennycook: I don't know the motivation, but this is the way that the return type is declared for the equivalent function objects in in ISO C+ (e.g., https://eel.is/c++draft/function.objects#arithmetic.operations.plus-2).

I think you're right, though -- if `std::forward<T>(t) + std::forward<U>(u)` is the statement that the operator will run, it's safer to include the `std::forward` in the `decltype` just in case somebody does something really weird.

keryell: Yes, I should remember that if it is weird in the C++ standard it is because some people do weird things in real life at the first place!

keryell (approved):
Thanks!

gmlueck (approved):

gmlueck:
For my own benefit, this is the Intel tracker to add CTS tests: CMPLRLLVM-65338

Pennycook:
The change has been merged into DPC++: https://github.com/intel/llvm/pull/17239.

tomdeakin:
Still waiting CTS I think.

PR #650, opened 2024-10-25
Add sycl_khr_addrspace_cast extension

Pennycook:
This extension splits `sycl::address_space_cast` into two functions:

- `khr::static_addrspace_cast`, which casts with no run-time checks.
- `khr::dynamic_addrspace_cast`, which casts with run-time checks.

### DIFF
--- a/adoc/extensions/index.adoc
+++ b/adoc/extensions/index.adoc
@@ -11,3 +11,5 @@ specification, but their design is subject to change.
 
 // leveloffset=2 allows extensions to be written as standalone documents
 // include::sycl_khr_extension_name.adoc[leveloffset=2]
+
+include::sycl_khr_addrspace_cast.adoc[leveloffset=2]
--- a/adoc/extensions/sycl_khr_addrspace_cast.adoc
+++ b/adoc/extensions/sycl_khr_addrspace_cast.adoc
@@ -1,0 +1,155 @@
+[[sec:khr-addrspace-cast]]
+= SYCL_KHR_ADDRSPACE_CAST
+
+[code]#sycl::address_space_cast# does two things: first, it checks whether a
+given raw pointer can be cast to a specific address space; and second, it
+performs the requested cast.
+In cases where a developer is asserting that a raw pointer points to an
+object in a specific address space, run-time checks are not required and may
+have undesirable performance impact.
+
+This extension splits [code]#sycl::address_space_cast# into two functions:
+
+- [code]#static_addrspace_cast#, which casts with no run-time checks.
+- [code]#dynamic_addrspace_cast#, which casts with run-time checks.
+
+[[sec:khr-addrspace-cast-dependencies]]
+== Dependencies
+
+This extension has no dependencies on other extensions.
+
+[[sec:khr-addrspace-cast-feature-test]]
+== Feature test macro
+
+An implementation supporting this extension must predefine the macro
+[code]#SYCL_KHR_ADDRSPACE_CAST# to one of the values defined in the table
+below.
+
+[%header,cols="1,5"]
+|===
+|Value
+|Description
+
+|1
+|Initial version of this extension.
+|===
+
+[[sec:khr-addrspace-cast-functions]]
+== Address space cast functions
+
+[source,role=synopsis]
+----
+
+namespace sycl {
+namespace khr {
+
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+static_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+static_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+dynamic_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+dynamic_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+
+} // namespace khr
+} // namespace sycl
+----
+
+.[apidef]#static_addrspace_cast#
+[source,role=synopsis,id=api:khr-addrspace-cast-static_addrspace_cast]
+----
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+static_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+static_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+----
+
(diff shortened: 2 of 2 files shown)


Pennycook:
As proposed in last week's WG meeting, I've split this into two extensions: 760627f7f675b08781cf6b4122cc57890b3f85c3

adoc/extensions/sycl_khr_dynamic_addrspace_cast.adoc
```diff
@@ -0,0 +1,104 @@
...
+|===
+
+[[sec:khr-dynamic-addrspace-cast-functions]]
+== Dynamic address space cast functions
+
+[source,role=synopsis]
+----
+
+namespace sycl {
+namespace khr {
```
keryell, line 37:
```suggestion
namespace sycl::khr {
```

adoc/extensions/sycl_khr_dynamic_addrspace_cast.adoc
```diff
@@ -0,0 +1,104 @@
...
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+dynamic_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+dynamic_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+
+} // namespace khr
+} // namespace sycl
```
keryell, line 48:
```suggestion
} // namespace khr::sycl
```

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
```diff
@@ -0,0 +1,99 @@
...
+|===
+
+[[sec:khr-static-addrspace-cast-functions]]
+== Static address space cast functions
+
+[source,role=synopsis]
+----
+
+namespace sycl {
+namespace khr {
```
keryell, line 42:
```suggestion
namespace sycl::khr {
```

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
```diff
@@ -0,0 +1,99 @@
...
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+static_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+static_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+
+} // namespace khr
+} // namespace sycl
```
keryell, line 53:
```suggestion
} // namespace khr::sycl
```

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+template <addrspace Space, typename ElementType>
+multi_ptr<ElementType, Space, access::decorated::no>
+static_addrspace_cast(ElementType* ptr);
+
+template <addrspace Space, typename ElementType, access::decorated DecorateAddress>
+multi_ptr<ElementType, Space, DecorateAddress>
+static_addrspace_cast(multi_ptr<ElementType, addrspace_generic, DecorateAddress> ptr);
+
+} // namespace khr::sycl
```
gmlueck, line 51: I think we do not want to duplicate the synopsis twice like this.  For APIs that are free functions, I think we need only one synopsis that shows the namespace and is followed by the _Preconditions_, etc. descriptions.

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
```diff
@@ -0,0 +1,97 @@
...
+----
+
+.[apidef]#static_addrspace_cast#
+[source,role=synopsis,id=api:khr-static-addrspace-cast-static_addrspace_cast]
+----
+template <addrspace Space, typename ElementType>
```
gmlueck, line 57:
```suggestion
template <access::address_space Space, typename ElementType>
```

Maybe we want this instead?  Note that the `addrspace` alias depends on #633.  Is #633 targeted for SYCL-Next?  If so, I think we want this KHR to work even for SYCL-2020.

adoc/extensions/sycl_khr_dynamic_addrspace_cast.adoc
```diff
@@ -0,0 +1,102 @@
+[[sec:khr-dynamic-addrspace-cast]]
+= SYCL_KHR_DYNAMIC_ADDRSPACE_CAST
+
+This extension renames [code]#sycl::address_space_cast# to
+[code]#dynamic_addrspace_cast# to align with the [code]#static_addrspace_cast#
```
gmlueck, line 5: Kind of pedantic, but I think "renames" is not the right word here because it implies that the old name is going away.  Maybe:

> This extension introduces `dynamic_addrspace_cast` with the same semantics as `sycl::address_space_cast` to align with ...

adoc/extensions/sycl_khr_dynamic_addrspace_cast.adoc
```diff
@@ -0,0 +1,102 @@
...
+----
+#include <sycl/sycl.hpp>
+using namespace sycl; // (optional) avoids need for "sycl::" before SYCL name
+
+// This function accepts raw pointers, but dispatches internally.
+template <typename T>
+void update(T* ptr, int x)
+{
+  // If cast to global returns non-null, call the global version.
+  if (khr::dynamic_addrspace_cast<addrspace_global>(ptr)) {
```
gmlueck, line 93: Will this work?  When I look at the definition of `multi_ptr`, I don't see any conversion to `bool`.  It might work because of:

```
// Conversion to the underlying pointer type
// Deprecated, get() should be used instead.
operator pointer() const;
```

However, it seems bad for this example to rely on a deprecated conversion.  Maybe this example should be rewritten to use `.get()`?  Or, we could consider adding `explicit operator bool()` to `multi_ptr`.

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
Pennycook: Makes sense. Fixed in a12c3ed79e3c60547e3a538e1532d55e03e4f035.

adoc/extensions/sycl_khr_static_addrspace_cast.adoc
Pennycook: You're right; #633 is being targeted for SYCL-Next, so we should use the old enum here. Fixed in c5ccd25458c75c77e52e05dc6c466fe14a009099.

adoc/extensions/sycl_khr_dynamic_addrspace_cast.adoc
Pennycook: I hope it _will_ be going away, but since it's not deprecated yet I've adopted your proposed wording in 49b99c7056b33142c61ec079b47a8c2ed878a4d4.

gmlueck (approved):

keryell (approved):
Thanks!

tomdeakin (approved):

gmlueck:
For my own benefit, this is the Intel tracker for adding CTS tests for this KHR: CMPLRLLVM-65336

PR #651, opened 2024-10-28
Fix error behavior for initialize_to_identity

Pennycook:
Previously, the specification said that a compiler must issue a diagnostic if the `initialize_to_identity` property was used without a known identity value. This requirement cannot be satisfied, because the presence of the `initialize_to_identity` value in a `property_list` cannot be proven until run-time.

This change replaces the compile-time diagnostic with a run-time exception.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -13892,9 +13892,10 @@ property::reduction::initialize_to_identity
       identity value passed to the reduction interface, or to the identity
       value determined by the [code]#known_identity# trait if no identity value
       was specified. If no identity value was specified and an identity value
-      cannot be determined by the [code]#known_identity# trait, the compiler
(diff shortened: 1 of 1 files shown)


TApplencourt:
Good catch. 

I guess we can put this kind of stuff back when we have "compile-time properties".

TApplencourt (approved):

Pennycook:
> I guess we can put this kind of stuff back when we have "compile-time properties".

Agreed.

> BTW, do we have a test on the CTS that we need to adjust? (aka compile time versus runtime error)

I think so.  I'm not very good at reading the CTS tests, but it looks to me like we only test valid uses of `initialize_to_identity`.

gmlueck (approved):

AerialMantis (approved):

tomdeakin (approved):

tomdeakin:
WG approved to merge

PR #657, opened 2024-11-07
Add "squash-commits.sh"

gmlueck:
Add a script that maintains a branch of squashed commits, where each commit represents one pull request that was merged into one of the main branches of this repo.


keryell:
My proposal did not have this concept of starting point.
What is the motivation of having a starting point which is not the beginning of the history?

PR #664, opened 2024-11-26
Add constraint to vec <<, >> operators

gmlueck:
I assume it was simply an oversight that we forgot to add this constraint to these operators.  This synopsis already listed the constraint in its comment, so this change just brings the description into alignment with the synopsis.

C++ does not allow `<<` or `>>` operators where the left-hand-side is a floating point value.  In addition, DPC++ already diagnoses an error if you attempt to use `vec::operator<<` where the left-hand-side is a floating point value.


TApplencourt (approved):

keryell (approved):
Good catch!

tomdeakin:
WG approved to merge

PR #666, opened 2024-12-02
Fix definition of RET for vec operators

gmlueck:
This fixes what I assume was just a typo.  The `vec` comparison and logical operators all return a result type whose vector element is `RET`.  Most of the operators say that `RET` is `int64_t` when the input operands are 64-bit.  However, three of the operators defined `RET` to be `uint64_t` in this case.

The `uint64_t` type is an outlier because `RET` is always signed for the other cases.  I can think of no reason why these three operators should be different from the others.

DPC++ already implements these operators to return a vector of `int64_t`, consistent with this fix.


keryell (approved):
Thanks!

TApplencourt (approved):

AerialMantis (approved):

PR #669, opened 2024-12-03
Add vec explicit conversion

gmlueck:
This is change 2 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change adds an explicit conversion when the `vec` has one element, allowing a conversion to `T` whenever `DataT` can be explicitly converted to `T`.

This enables code like the following, where the element type `DataT` is itself a class:

```
sycl::vec<sycl::half, 1> h1;
int i = static_cast<int>(h1);
if (h1) {}
```

These changes correspond to slides 8 - 9 of the presentation that was discussed in the WG meetings.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17872,6 +17872,21 @@ to allow standard conversion sequence for implicit conversion.
 a@
 [source]
 ----
+template<typename T>
+explicit operator T() const
+----
+a@ _Constraints:_
+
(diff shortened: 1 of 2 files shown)


TApplencourt (approved):

tomdeakin (approved):

tomdeakin:
WG approved to merge for SYCL 2020.

PR #670, opened 2024-12-03
Fix 1-element "vec" ambiguities

gmlueck:
This is change 3 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change fixes an ambiguity that occurs for all of the binary operators when one of the operands is a 1-element `vec` and the other operand is a scalar that is not `DataT` but is convertible to `DataT`. For example:

```
sycl::vec<float, 1> v;
v + 2;
```

Prior to this change, this would result in an ambiguity error from the compiler between:

* Convert 2 -> `float`, call `operator+(const vec&, const DataT&)`
* Convert `vec<float, 1>` -> `float`, call `operator+(float, int)`

After this change, the code snippet above compiles without error, calling `operator+(const vec&, const int&)` that is instantiated from a template.

These changes correspond to slides 10 - 13 of the presentation that was discussed in the WG meetings.


TApplencourt (approved):

tomdeakin:
WG approved to merge

PR #671, opened 2024-12-03
Clarify "vec" ADL behavior

gmlueck:
This is change 4 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change clarifies the ADL behavior of the `vec` class.  Without this clarification, implementations could make additional operators available on the ADL search set, which could lead to ambiguity errors when compiling applications.  For example, we know that adding `vec` to the ADL search set for the swizzled vector operators leads to ambiguity errors.

These changes correspond to slide 14 of the presentation that was discussed in the WG meetings.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -18573,15 +18573,13 @@ behavior:
 
 * When an argument is the [code]#+__writeable_swizzle__+# type, the ADL search
   set must include only the class template definition for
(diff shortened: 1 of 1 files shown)


keryell (approved):
Good catch!

TApplencourt (approved):

tomdeakin:
Wg approved to merge

AerialMantis (approved):

PR #672, opened 2024-12-03
Change return type of named swizzles

gmlueck:
This is change 5 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change aligns the return type of the 1-element "named" swizzles to the behavior of the existing implementations (and the CTS).  This changes the specified behavior of code like this:

```
vec<uint8_t, 4> v4{255};
auto v = v4.x();         // Returns reference to "uint8_t" not 1-element swizzle
int i = v4.x() + 1;      // Result is 256 not 0
```

These changes correspond to slides 19 - 22 of the presentation that was discussed in the WG meetings.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17937,14 +17937,12 @@ The [code]#swizzleIndexes# argument pack specifies the elements in the swizzle.
 a@
 [source]
 ----
-__writeable_swizzle__ XYZW_ACCESS()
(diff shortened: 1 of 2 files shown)


keryell (approved):
Thanks, that makes sense.

TApplencourt (approved):

tomdeakin:
WG approved to merge

PR #673, opened 2024-12-03
Clarify "vec" DataT

gmlueck:
This is change 6 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change clarifies the set of legal data types for a `vec` element and also clarifies what `sycl::byte` is.

These changes correspond to slides 23 - 25 and 28 of the presentation that was discussed in the WG meetings.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17727,8 +17727,8 @@ a@
 ----
 byte
 ----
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks!

TApplencourt (approved):

AerialMantis (approved):

tomdeakin:
WG approved to merge

PR #674, opened 2024-12-03
Allow "std::byte" as "vec" element type

gmlueck:
This is change 7 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change clarifies that `std::byte` is a legal element type for `vec`.  This is one possible interpretation of the previous wording where we said that the element data type must be a "basic scalar type". We think that phrase included `sycl::byte`.  However, `sycl::byte` was deprecated in SYCL 2020 saying that `std::byte` is a replacement.

If we allow `std::byte`, we need to adjust the constraints for the various `vec` operators.  We decided to allow these operators on `vec` of `std::byte`, which mostly follows the C++ rules for operators on plain `std::byte`:

* Comparison: ==, !=, <, <=, >, >=
* Bitwise: &, |, ^, ~, &=, |=, ^=

I decided it was clearer to rephrase the constraints to say which types are allowed rather than listing the types that are not allowed.  For example, this results in phrasing like:

> Available only when DataT is an integral type.

Rather than:

> Available only when: DataT != float && DataT != double &&
> DataT != half && DataT != std::byte.

These changes correspond to slide 26 of the presentation that was discussed in the WG meetings.


keryell (approved):
Thanks for the generalization.

nliber (approved):

gmlueck:
CTS tests added in KhronosGroup/SYCL-CTS#1058

PR #675, opened 2024-12-03
Generalize RET for vec logical and comparison ops

gmlueck:
This is change 8 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change clarifies the return type for the comparison and logical operators.  All of these operators return a vec of integers, where the size of the integer elements depends on the element types of the input operands.  The previous wording only specified the return type for certain `DataT` types, leaving other types unspecified.  Reword this in a general way which applies to all `DataT` types.

These changes correspond to slide 27 of the presentation that was discussed in the WG meetings.


keryell (approved):
Good simplification.

TApplencourt (approved):

tomdeakin:
WG approved to merge.

PR #676, opened 2024-12-03
Remove underspecified vector_t

gmlueck:
This is change 9 of 9 that fix problems with the specification of the `vec` class.  An implementation that follows the existing specification would not accept common code patterns and would not pass the CTS.  None of the existing implementations actually follow the existing specification.

This change removes the `vector_t` type alias from the `vec` class and also removes the conversions to and from this type.  The intention of this type was to provide a conversion between `vec` and a compiler-specific builtin vector type.  However, these conversion also lead to ambiguity errors in common code patterns such as:

```
sycl::vec<sycl::half, 1> h1;
sycl::half h;
h = h1;
```

Prior to this change, the last line could result in an ambiguity error from the compiler between:

* Convert `vec<half,1>` -> `half`, call `half::operator=(const half&)`
* Convert `vec<half,1>` -> `_Float16`, convert `_Float16` -> `float`, call `half::operator=(float)`

What's worse is that the exact type alias of `vector_t` is implementation-defined, so these ambiguity errors could be different for each SYCL implementation.

During the WG meeting, we learned that the two major SYCL implementations (DPC++ and AdaptiveCpp) define `vector_t` differently even though they are both based on clang.  Given this lack of commonality, the WG decided to remove `vector_t` entirely from the spec. Implementations can decide for themselves whether to provide a compiler-specific conversion like this.

These changes correspond to slides 29 - 31 of the presentation that was discussed in the WG meetings.

### DIFF
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -1063,8 +1063,8 @@ OpenCL C functions that are linked with, using either offline or online
 compilation, must be declared as [code]#extern "C"# function declarations.
 The function parameters of these function declarations must be defined as the
 OpenCL C interoperability aliases; [code]#pointer# of the [code]#multi_ptr#
-class template, [code]#vector_t# of the [code]#vec# class template and scalar
-data type aliases described in <<table.types.aliases>>.
+class template, and scalar data type aliases described in
+<<table.types.aliases>>.
 
 // \include{opencl_extensions}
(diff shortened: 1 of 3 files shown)


nliber (approved):

gmlueck:
Note that KhronosGroup/SYCL-CTS#1055 adjusts the CTS to remove tests for `vector_t`.

keryell (approved):
That sounds reasonable.

tomdeakin:
WG approved to merge

PR #678, opened 2024-12-06
Specify copy assignment operator for vec swizzles

gmlueck:
Even though we specify an assignment operator for
`__writeable_swizzle__` like this:

```
template</*unspecified*/>
const __writeable_swizzle__&
operator=(const __writeable_swizzle__</*unspecified*/>& rhs) const
```

The compiler will not select this operator when the left-hand-side has the same type as the right-hand-side.  (Even though the template constraints would allow such a selection.)  Instead, the compiler will generate the default copy assignment operator and call that.

Since the default copy assignment operator would not be correct (performing a shallow copy), the `__writeable_swizzle__` class must provide a user defined copy assignment operator.

The copy assignment operator for `__const_swizzle__` was already deleted in the "constructors" section.  It seemed better to move this to the "member functions" part of the specification, next to the specification for the copy assignment operator for `__writeable_swizzle__`.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -18678,13 +18678,9 @@ __writeable_swizzle__(const __writeable_swizzle__&) = delete
 
 __const_swizzle__() = delete
 __const_swizzle__(const __const_swizzle__&) = delete
-
-__const_swizzle__& operator=(const __const_swizzle__&) = delete
 ----
 |====
 The default constructor and copy constructor are deleted.
-
-The copy assignment operator for [code]#+__const_swizzle__+# is deleted.
 
 ===== Destructors for the swizzled vector class templates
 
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -18860,6 +18856,46 @@ corresponds to the position [code]#index# of the swizzle operation.
 
 '''
 
+[frame=all,grid=none,separator="@"]
+|====
+a@
+[source]
+----
+const __writeable_swizzle__&
+operator=(const __writeable_swizzle__& rhs) const
```
nliber, line 18865: Why is this a `const` member function?

nliber (changes requested):

adoc/chapters/programming_interface.adoc
nliber: It is still logically a non-`const` operation.

There is precedent in C++ for this, such as `ostream_iterator::operator=(const T&)` (as well as some of its other operators).

nliber (approved):

tomdeakin (approved):

tomdeakin:
WG approved to merge.

PR #680, opened 2024-12-11
Change the base C++ version

gmlueck:
Change wording to allow the core C++ language to be any version C++17 or later.

Move sections about normative references and non-normative wording to the introduction, so they are earlier in the document.

My longer term goal is to remove "chapter/references.adoc" and add any references we require to the "Normative references" section.  Some more work needs to be done for this to happen:

* Decide if we need to list the C++ defect reports as normative references.  Currently, the "references" appendix lists only DR2325. Is there a reason to list this one and not others?

* Decide if OpenCL should be a normative reference.  If so, which version(s)?

### DIFF
--- a/adoc/chapters/introduction.adoc
+++ b/adoc/chapters/introduction.adoc
@@ -214,8 +214,8 @@ correct.
 
 A SYCL implementation must conform to {cpp17} or to one of the more recent {cpp}
 versions listed in <<sec:normativerefs>>.
-This version of {cpp} is subsequently referred to as the {cpp} core language in
-this document.
+The version of {cpp} to which a SYCL implementation conforms is referred to as
+the {cpp} core language in this document.
 All APIs in this specification are available regardless of the version of the
 {cpp} core language unless the description specifically states otherwise.
 If an API is documented with the phrase "Minimum C++ Version: __Version__", then
--- a/adoc/chapters/introduction.adoc
+++ b/adoc/chapters/introduction.adoc
@@ -153,6 +153,95 @@ build on top of SYCL as an open platform for system-wide heterogeneous
 processing innovation.
 
 
+[[sec:normativerefs]]
+== Normative references
+
+Normative references are references to external documents or resources to which
+implementers of SYCL must comply with all, or specified portions of, as
+described in this specification.
+
+International Organization for Standardization (ISO). +
+ISO/IEC 14882:2017 - Programming languages - {cpp},
+https://www.iso.org/standard/68564.html +
+(Referred to as {cpp17} in this document.)
+
+International Organization for Standardization (ISO). +
+ISO/IEC 14882:2020 - Programming languages - {cpp},
+https://www.iso.org/standard/79358.html +
+(Referred to as {cpp20} in this document.)
+
+International Organization for Standardization (ISO). +
+ISO/IEC 14882:2024 - Programming languages - {cpp},
+https://www.iso.org/standard/83626.html +
+(Referred to as {cpp23} in this document.)
+
+// Jon: are any of the OpenCL specifications normative? They are also
+// referred to from the SYCL spec, and some of the definitions appear to be
+// required.
+
+
+[[sec:nonnormativerefs]]
+== Non-normative notes and examples
+
+Unless stated otherwise, text within this SYCL specification is normative and
+defines the required behavior of a SYCL implementation.
+Non-normative / informational notes are included within this specification using
+either of two formats.
+One format for non-normative notes is the "`note`" callout of this form:
+
+[NOTE]
+====
+Information within a note callout, such as this text, is for informational
+purposes and does not impose requirements on or specify behavior of a SYCL
+implementation.
+====
+
+The other format for a non-normative note is like this:
+
+{note}This is also a non-normative note.
+{endnote}
+
+Source code examples within the specification are provided to aid with
+understanding, and are non-normative.
+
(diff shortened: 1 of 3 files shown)


adoc/chapters/introduction.adoc
```diff
@@ -153,6 +153,95 @@ build on top of SYCL as an open platform for system-wide heterogeneous
...
+
+A SYCL implementation must conform to {cpp17} or to one of the more recent {cpp}
+versions listed in <<sec:normativerefs>>.
+This version of {cpp} is subsequently referred to as the {cpp} core language in
+this document.
+All APIs in this specification are available regardless of the version of the
+{cpp} core language unless the description specifically states otherwise.
+If an API is documented with the phrase "Minimum C++ Version: __Version__", then
+that API is available only if the version of the implementation's {cpp} core
+language is at least _Version_.
```
Pennycook, line 223: I think the repetition of "C++ core language" here is confusing.  The second sentence in this paragraph defines "C++ core language" to mean "C++17", so these sentences appear to be talking about different versions of C++17.

How about:

```suggestion
All APIs in this specification are available regardless of the {cpp} version
unless the description specifically states otherwise.
If an API is documented with the phrase "Minimum C++ Version: __Version__", then
that API is available only if the SYCL implementation conforms to a {cpp} version of
_Version_ or later.
```

The first sentence in the paragraph already talks about SYCL implementations "conforming to" specific C++ versions, which is why I picked that wording.

gmlueck: The term "C++ core language" is intended to identify whatever C++ version the implementation chooses.  The first sentence says that an implementation must choose one of the versions listed in section `<<sec:normativerefs>>` (which are all the versions from C++17 and greater).

In practice, I expect that most SYCL implementations will choose the C++ version via a command line parameter like `-std=`.

keryell: What happens when ISO C++26 is out but this is not yet mentioned in the SYCL spec as s normative norm? Is it up to the implementation whether a program using C++26 in host part will work? :thinking:

Pennycook: > The term "C++ core language" is intended to identify whatever C++ version the implementation chooses. The first sentence says that an implementation must choose one of the versions listed in section `<<sec:normativerefs>>` (which are all the versions from C++17 and greater).

Oh, I see.  I misunderstood the sentence.  I read "_This_ version of C++" as referring to C++17, not as "the C++ standard to which a SYCL implementation conforms".

If we wanted to make this unambiguous we could say:

> A SYCL implementation must conform to {cpp17} or to one of the more recent {cpp} versions listed in <<sec:normativerefs>>.
The version of {cpp} to which a SYCL implementation conforms is referred to as the {cpp} core language in this document.

or

> A SYCL implementation must conform to a specific version of {cpp}, referred to as the {cpp} core language in this document.
> The {cpp} core language must be {cpp17} or one of the more recent {cpp} versions listed in <<sec:normativerefs>>.

...but if I'm alone in thinking the original wording is confusing, you can ignore me!

gmlueck: > What happens when ISO C++26 is out but this is not yet mentioned in the SYCL spec as s normative norm? Is it up to the implementation whether a program using C++26 in host part will work? 🤔

I don't see this as a practical problem.  There are already C++ compilers that support proposed C++26 features even though there is no ratified C++26 specification yet.  Those compilers can exist -- they just aren't conformant to any spec yet.  I think it could be the same for SYCL.  An implementation can choose to support C++26 with SYCL if it wants.  It just won't be strictly conformant to the SYCL spec.

What would be the alternative?  I guess the SYCL spec could say that the core C++ version could be *any* future version of C++ even though we don't know what language features this would include.  I think that would be more problematic because we wouldn't know how those unknown language features would work in device code.

This reminds me ... we should think through all the new language features in C++20 and C++23 and decide if they are legal in device code.  For those that are not, we need to add them to [section 5.5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:language.restrictions.kernels) "Language restrictions for device functions".

gmlueck: > If we wanted to make this unambiguous we could say:

I updated the wording according to your suggestion in d5d63ed9baf18a4026d67014331b7d0f7f9f5393.

keryell (approved):
Thanks!

etomzak:
How much confidence is there that rebasing SYCL onto newer C++ standards doesn't (drastically) change the meaning of (too many) things in the SYCL spec, as noted [here](https://github.com/KhronosGroup/SYCL-Docs/pull/685#issuecomment-2577704631) for example?

Pennycook:
> How much confidence is there that rebasing SYCL onto newer C++ standards doesn't (drastically) change the meaning of (too many) things in the SYCL spec, as noted [here](https://github.com/KhronosGroup/SYCL-Docs/pull/685#issuecomment-2577704631) for example?

I forget where it was discussed, but I think this is a known risk at this point.

There are some other terms that changed in C++20 (e.g., related to the memory model) and we'll need to make sure that we remain consistent with those changes.  There is a dedicated sub-committee for alignment with ISO C++, so if we encounter issues like this I think we're prepared to do the work to fix things up.

TApplencourt (approved):

Pennycook (approved):

tomdeakin:
WG proved to merge

PR #681, opened 2024-12-12
CTS issue workflow: Only run on main branch

psalz:
Changes that are being cherry-picked into an older version of the spec don't need a separate CTS issue, as the CTS targets all versions from SYCL 2020 onward.

### DIFF
--- a/.github/workflows/open_cts_issue.yml
+++ b/.github/workflows/open_cts_issue.yml
@@ -10,6 +10,9 @@ on:
     types: opened
     paths:
       - 'adoc/**'
+    # We don't want to create issues for cherry-picks into older spec revisions
+    branches:
+      - main
 jobs:
   create-issue:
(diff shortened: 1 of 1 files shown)


psalz:
Now that I think about it, we could also just limit the workflow to run only on the `main` branch - what do you prefer?

gmlueck:
That would probably be easier because we wouldn't need to add the label.  There may be times when we merge a PR directly to the "sycl-2020" branch that is not a cherry pick.  However, it's probably rare enough that we can just remember to open a CTS issue if necessary.

psalz:
> That would probably be easier because we wouldn't need to add the label. There may be times when we merge a PR directly to the "sycl-2020" branch that is not a cherry pick. However, it's probably rare enough that we can just remember to open a CTS issue if necessary.

Ok, done!

gmlueck (approved):

PR #688, opened 2025-01-17
Add pointer-to-object to list of `atomic_ref` types

### LINKED ISSUE #684: Clarification needed on valid types and specializations for `atomic_ref`
While working with SimSYCL we ran into a somewhat confusing aspect of the spec regarding `atomic_ref`.

In the beginning the spec rather clearly and unequivocally states the following:

> The template parameter T must be one of the following types:
> * int
> * unsigned int
> * long
> * unsigned long
> * long long
> * unsigned long long
> * float
> * double

Notably absent are any pointer types.

However, at the same time, later on several functions available specifically on a partial specialization of `atomic_ref<T*>` are described.

This doesn't seem to make much sense to me. One way to read it would be to take it as `T*` for each of the `T`s described previously being allowed, but if a pointer to some types is valid then there's no reason that another pointer to a different type should be invalid (that I can think of).

An easy way to solve it would be to add something like "any pointer type" to the list of allowed types at the start.

Off topic / only tangentially related: if this is touched, it might also be a good idea to use the opportunity to allow `bool`.

(See also https://github.com/celerity/SimSYCL/pull/32)

Pennycook:
The fact that this was missing from the list is a bug. `sycl::atomic_ref` has a specialization for pointer types, and several other parts of the specification talk about atomic operations on pointers.

Use of the term "pointer-to-object types" is aligned with `std::atomic_ref`.

Closes #684.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20033,8 +20033,9 @@ The template parameter [code]#T# must be one of the following types:
 * [code]#unsigned long#,
 * [code]#long long#,
 * [code]#unsigned long long#,
-* [code]#float#, or
-* [code]#double#.
+* [code]#float#,
+* [code]#double#, or
+* Any pointer-to-object type.
 
 In addition, the type [code]#T# must satisfy one of the following conditions:
 


gmlueck (approved):

keryell (approved):
That makes sense.

TApplencourt (approved):

illuhad (approved):

tomdeakin:
WG approved to merge, please merge into SYCL 2020.

PR #691, opened 2025-01-22
Change KHR name to lower case

gmlueck:
@Pennycook and I both think that the name of KHR extensions should be in lower case.  This affects only the documentation when referring to the name of the extension.  The feature-test macro remains in upper case, consistent with other SYCL macro names.


gmlueck:
Merging as editorial

PR #700, opened 2025-01-28
[KHR] add sycl_khr_queue_empty_query

TApplencourt:
Ported: https://gitlab.khronos.org/sycl/Specification/-/merge_requests/727/

Add `queue.size()` and `queue.empty()`. I think we agreed that the other `get_wait_list()`  was hard to implement, so I removed it from here.

I need help naming those APIs. I'm not sure how it works to add a new member function to an existing class... Should I prefix the name of the function with `khr`? 

Thanks in advance!

@Pennycook, I would appreciate it if you could look at my example.  I think/hope / pray that it's free of UB :)


adoc/extensions/sycl_khr_queue_size_queries.adoc
```diff
@@ -0,0 +1,106 @@
...
+__Returns__: [code]#true# if [code]#(size() == 0)#.
+
+'''
+
+'''
+
+.[apidef]#khr::queue::size#
+[source,role=synopsis,id=api:khr-queue-size-queries-size]
+----
+size_t size() const
```
gmlueck, line 59: I think we should not add `queue::size`.  See the discussion in internal issue 636.

Consequently, I think we should rename the extension to something like "sycl_khr_queue_empty".

adoc/extensions/sycl_khr_queue_size_queries.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+This extension adds the following functions to the [code]#sycl::queue# clase,
+which provides information about the size of the queue.
+
+'''
+
+.[apidef]#khr::queue::empty#
+[source,role=synopsis,id=api:khr-queue-size-queries-empty]
+----
+bool empty() const
```
gmlueck, line 47:
```suggestion
bool khr_empty() const
```

We prefix new member functions with "khr_".  See the sycl_khr_default_context KHR.

adoc/extensions/sycl_khr_queue_size_queries.adoc
```diff
@@ -0,0 +1,106 @@
...
+== Queue Functions added to query size information
+
+
+This extension adds the following functions to the [code]#sycl::queue# clase,
+which provides information about the size of the queue.
+
+'''
+
+.[apidef]#khr::queue::empty#
+[source,role=synopsis,id=api:khr-queue-size-queries-empty]
```
gmlueck, line 45:
```suggestion
.[apidef]#queue::khr_empty#
[source,role=synopsis,id=api:queue-khr-empty]
```

adoc/extensions/sycl_khr_queue_size_queries.adoc
```diff
@@ -0,0 +1,106 @@
...
+----
+bool empty() const
+----
+
+__Returns__: [code]#true# if [code]#(size() == 0)#.
```
gmlueck, line 50:
```suggestion
_Returns:_ [code]#true# if [code]#(size() == 0)#.
```

Kind of a nit, but we formatted like this for the other KHRs.

adoc/extensions/sycl_khr_queue_size_queries.adoc
TApplencourt: Oh I see, wrong recollection. I was thinking that remove the event_wait_list, not the `size`. Thanks, will update! 

I started with `empty` and then remenred about size. I guess I should have committed the `empty` version :(

adoc/extensions/sycl_khr_queue_size_queries.adoc
TApplencourt: Thanks!

TApplencourt:
Updated with just `empty`.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,86 @@
...
+'''
+
+.[apidef]#queue::khr_empty#
+[source,role=synopsis,id=api:queue-khr-empty]
+----
+bool khr_empty() const
+----
+
+_Returns:_ [code]#true# if all <<command,commands>> enqueued on this particular
+queue have completed, [code]#false# otherwize.
```
Pennycook, line 44:
```suggestion
queue have completed, [code]#false# otherwise.
```

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,86 @@
...
+This extension adds the following function to the [code]#sycl::queue# class,
+which provides information about the emptiness of the queue.
+
+'''
+
+.[apidef]#queue::khr_empty#
+[source,role=synopsis,id=api:queue-khr-empty]
+----
+bool khr_empty() const
+----
```
Pennycook, line 41: I forget if we already discussed this or not, but should this be called `is_empty()`?

Somebody might call `queue.empty()` expecting it to _empty_ the queue of its contents.  It's more obvious that `queue.is_empty()` tests a condition.

I know that there is an `empty()` in `std::vector` and a `std::empty()` free-function, but those are defined for containers.  `queue` is _not_ a container (even if it has the same name as one), so it may be in our interest to use names that are very clearly different to those in the container interface.

I'd be interested to hear what @nliber thinks about this.

Pennycook (commented):
> @Pennycook, I would appreciate it if you could look at my example. I think/hope / pray that it's free of UB :)

I think this is free of UB *if* (as Greg pointed out) you use an in-order queue.

My reasoning, in case anybody is interested:
- The `std::atomic_bool` is only used by host code.
- If the `host_task` executes eagerly it will run until the atomic value changes.
- If the `host_task` executes lazily it will not run until `e2.wait()` is called.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+_Returns:_ [code]#true# if all <<command,commands>> enqueued on this particular
```
gmlueck, line 43:
```suggestion
_Returns:_ [code]#true# if all <<command,commands>> enqueued on this
```

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,86 @@
...
+      }
+      std::cout << iter << std::endl;
+    });
+  });
+  assert(!q.khr_empty());
+  auto e2 = q.single_task(e1, [=] {});
+  assert(!q.khr_empty());
+  start = true;
+  e2.wait();
+  assert(q.khr_empty());
```
gmlueck, line 84: I think this is not correct.  Because the queue is out-of-order, it's possible that `e2` completes before `e1`.  Therefore, the first host task might not be complete at this time.

More generally ... this code reads more like a CTS test to me than an example.  Can you think of some short example that demonstrates a useful usage of `khr_empty`?

TApplencourt: >I think this is not correct. Because the queue is out-of-order, it's possible that e2 completes before e1. Therefore, the first host task might not be complete at this time.

e2 depend on e1. 

>More generally ... this code reads more like a CTS test to me than an example. Can you think of some short example that demonstrates a useful usage of khr_empty?

Good point. Let me cook something

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Good point, maybe`is_empty()` is a better name to avoid confusion. I will bow to Nevin recommendation :) 

(And to lament, it's sad ct that we cannot use `?` in  c++. In ruby, methods that return a bool have their name ended with a `?`, like `empty?`. Beautiful)

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Modified the example. I think it's more in line with how people use the API in Cuda 
```
  // Pool of queues
  std::vector<sycl::queue> Qs;
  size_t n_queues = 12;
  for (size_t i = 0; i < n_queues; i++)
    Qs.push_back(sycl::queue(sycl::property_list{sycl::property::queue::in_order{}}));

  // Useful receipt for load-balancing
  auto it = std::find_if(Qs.begin(), Qs.end(),
                         [](const sycl::queue &q) { return q.khr_empty(); }
  );
  if (it != Qs.end()) {
    std::cout << "Empty queue Present!" << std::endl;
  } else {
    std::cout << "No empty queue." << std::endl;
  }
```

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,80 @@
...
+#include <sycl/sycl.hpp>
+
+int main() {
+  // Pool of queues
+  std::vector<sycl::queue> Qs;
+  size_t n_queues = 12;
+  for (size_t i = 0; i < n_queues; i++)
+    Qs.push_back(sycl::queue(sycl::property_list{sycl::property::queue::in_order{}}));
+
+  // Useful receipt for load-balancing
```
gmlueck, line 70:
```suggestion
  // Useful recipe for load-balancing
```

The word "receipt" doesn't make sense to me.  Did you mean "recipe"?

adoc/extensions/sycl_khr_queue_empty_query.adoc
gmlueck: > e2 depend on e1

Oh, I missed that.  Sorry.

> Good point. Let me cook something

Much better.  Thanks!

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,80 @@
...
+#include <iostream>
+#include <algorithm>
+#include <sycl/sycl.hpp>
+
+int main() {
+  // Pool of queues
+  std::vector<sycl::queue> Qs;
+  size_t n_queues = 12;
+  for (size_t i = 0; i < n_queues; i++)
+    Qs.push_back(sycl::queue(sycl::property_list{sycl::property::queue::in_order{}}));
```
gmlueck, line 68:
```suggestion
    Qs.push_back(sycl::queue{});
```

Maybe just this?  It seems like the in-order property isn't really central to this example.

TApplencourt: If you use `out of order,` you have concurrency inside each queue, so you should not required to do any kind of load balancing or any kind of queue pools, for that matter.

The example was more to mimic what Cuda people do with their in-order streams.

I can add a comment to explain this.

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Indeed :(

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: What about:
```
// Pool of in-order queues. 
// One needs to load-balance between queues to expose concurrency.
```
But I have no hard feelings about it; I can remove the `in-order` property if it's people's preference.

gmlueck: Maybe you could create each queue on a different device?  Then, the example will load balance across different devices.

TApplencourt: Good idea!

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Tried with `c3330dc1994e`

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Tried with c3330dc1994e

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+[source,,linenums]
+----
+#include <iostream>
+#include <algorithm>
+#include <sycl/sycl.hpp>
+int main() {
+  // Pool of queues, one per device
+  std::vector<sycl::queue> Qs;
+  for (sycl::device &d : sycl::platform{}.get_devices())
```
gmlueck, line 65:
```suggestion
  for (sycl::device& d : sycl::platform{}.get_devices())
```

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+    Qs.push_back(sycl::queue(d));
+
+  // Useful recipe for load-balancing
+  auto it = std::find_if(Qs.begin(), Qs.end(),
+                         [](const sycl::queue &q) { return q.khr_empty(); });
```
gmlueck, line 70:
```suggestion
                         [](const sycl::queue& q) { return q.khr_empty(); });
```

gmlueck (commented):
I like your style better, but this is the style in the rest of the spec, so we should be consistent.

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Did we commit to a clang-format at some point? I don't remember...

TApplencourt: Ok, we have one; I just didn't used it my mistake. I should try to find the script I did to format the `[API,code]` and commit it to the repo / in the CI.
I guess I tried long time ago `https://gitlab.khronos.org/sycl/Specification/-/merge_requests/730#note_387779` and then forgot about it...

TApplencourt: Ok, I am confirming that `https://github.com/KhronosGroup/SYCL-Docs/pull/292#issuecomment-1270531337` did fix the wrong syntax. (and fixed the include order)
Will try to put that as a PR / linter; maybe be helpful for other people.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+[source,,linenums]
+----
+#include <algorithm>
+#include <iostream>
+#include <sycl/sycl.hpp>
+int main() {
+  // Pool of queues, one per device
+  std::vector<sycl::queue> Qs;
+  for (sycl::device& d : sycl::platform {}.get_devices())
```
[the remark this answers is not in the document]
TApplencourt: <3 Totally forgot about this one! Thanks a lot (and not thanks for the new C++ nightmare)

adoc/extensions/sycl_khr_queue_empty_query.adoc
tomdeakin: WG discussed, and had concencous on `empty`

TApplencourt:
Ready to ~review~ accept :)

gmlueck (approved):

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+[source,role=synopsis,id=api:queue-khr-empty]
+----
+bool khr_empty() const
+----
+
+_Returns:_ [code]#true# if all <<command,commands>> enqueued on this queue have
+completed, [code]#false# otherwise.
+
+{note} Since the implementation executes commands asynchronously, the returned
+value is a snapshot in time.
```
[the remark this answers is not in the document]
gmlueck: I think the concern is not that another thread might add something to the queue, it's that the SYCL runtime might asynchronously remove things from the queue.  No amount of careful coding on the application's part can solve this.  If `queue::khr_empty` returns false, the queue might become empty a moment later.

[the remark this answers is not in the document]
TApplencourt: I can add `immediately stale`. Maybe better than `snapshot in time`

You know that if it's `true` and that nobody from your app enqueue stuff in the queue, it will be still `true`.
For `false`, it's outside your app's control. Maybe the tick after it will become `true,` as @gmlueck  said (regardless of what you did in your code).

I don't know how to phrase it better. Sadly, `std::futur` doesn't have an `is_ready` that can inspire us.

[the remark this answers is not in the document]
DuncanMcBain: @gmlueck agreed - I just don't think that's as much of an issue as, say, believing that the queue is empty, then having another thread dump 2500 kernels onto it :D

@TApplencourt I think I stole the "immediately stale" wording from some other API. But I think it communicates the idea.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+'''
+
+.[apidef]#queue::khr_empty#
+[source,role=synopsis,id=api:queue-khr-empty]
+----
+bool khr_empty() const
+----
+
+_Returns:_ [code]#true# if all <<command,commands>> enqueued on this queue have
+completed, [code]#false# otherwise.
```
Pennycook, line 44: The comment about thread safety [here](https://github.com/KhronosGroup/SYCL-Docs/pull/700#discussion_r1937868992) made me think about what the desired synchronization/observability semantics are here.  I'm starting a new comment thread because I think it's a separate issue, and I don't think that comments about thread-safety/staleness can resolve it. 

Before we get into the standardese: should the examples below work, or not?

### Example 1: Host-Device Synchronization

```c++
bool* usm = sycl::malloc_shared<bool>(1, q);

q.single_task([=]() {
  *usm = false;
};

while (not q.empty()) {} // NB: This thread never called wait

// If the queue is empty, we "know" the single_task completed.
// Are its results guaranteed to be visible?
assert(*usm == false);
```

### Example 2: Inter-Thread Synchronization via Device

```c++
// Assume these allocations are visible to both threads.
bool* a = malloc(sizeof(bool));
bool* b = sycl::malloc_shared<bool>(1, q);

// Thread 1
{
  *a = true;
  q.single_task([=]() {
    *b = true;
  };
}

// Thread 2
{
  if (q.empty()) {
    if (*b == true) {
      // If the queue is empty, the single_task might have executed.
      // If b is true, we "know" the single task executed (assuming Example 1 is valid).
      // Are things Thread 1 did before enqueueing the task guaranteed to be visible to Thread 2?
      assert(*a == true); 
    }
  }
}
---

Answering these questions might be something that we want to defer until a larger rework of the execution model, but I wanted to bring it up so we don't lose track of it.

Comment (gmlueck) [reply adoc/extensions/sycl_khr_queue_empty_query.adoc:44]: I think example 1 is not guaranteed to work because a SYCL implementation is allowed to execute work in queues lazily, waiting for the application to call `wait`.  Therefore the loop on `q.empty` could be an infinite loop.

I think your real question is about inter-thread synchronization, though.  It seems like this question exists even without relying on `queue::empty`.  Is the following example well defined?

```
// Assume these allocations are visible to both threads.
bool* a = malloc(sizeof(bool));
bool* b = sycl::malloc_shared<bool>(1, q);
*a = false;
*b = false;

// Thread 1
{
  *a = true;
  q.single_task([=]() {
    *b = true;
  };
}

// Thread 2
{
  q.wait();
  // Are things Thread 1 did before enqueueing the task guaranteed to be visible to Thread 2?
  assert(*a == true); 
}
```

If you think the answer is "yes", what part of the spec guarantees this?

Comment (Pennycook) [reply adoc/extensions/sycl_khr_queue_empty_query.adoc:44]: > I think example 1 is not guaranteed to work because a SYCL implementation is allowed to execute work in queues lazily, waiting for the application to call `wait`. Therefore the loop on `q.empty` could be an infinite loop.

You're right, but I was trying not to overcomplicate the example by bringing eager/lazy into it!

If you ignore that some implementations may go into an infinite loop, the question is still interesting: if the host thread _does_ make it past that loop, should there be any guarantee of the `*usm` value?

> I think your real question is about inter-thread synchronization, though. It seems like this question exists even without relying on `queue::empty`. Is the following example well defined?
> ...
> If you think the answer is "yes", what part of the spec guarantees this?

I agree that we need to clarify the behavior of other APIs, which is why it might make sense just to note that `empty` has this problem and revisit it as part of later execution/memory model clarifications. Querying whether an event is in `complete` status has exactly the same problem.

I think `q.wait()` is different.  Although it's not clearly/formally stated anywhere that your example would work -- and `q.wait()` isn't described formally in terms of synchronization, etc -- I think it's aligned with the intent of `wait`.

I know examples are non-normative, but as a demonstrator of intent: the example in [the USM section](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm) shows that calling `wait` makes the memory available to the thread that called `wait`:
```c++
  myQueue.parallel_for(1024, [=](id<1> idx) {
    // Initialize each buffer element with its own rank number starting at 0
    data[idx] = idx;
  }); // End of the kernel function

  // Explicitly wait for kernel execution since there is no accessor involved
  myQueue.wait();

  // Print result
  for (int i = 0; i < 1024; i++)
    std::cout << "data[" << i << "] = " << data[i] << std::endl;
```

...the implication being that `wait()` is some sort of synchronizing operation, or at least that we can guarantee the end of the kernel function happens-before the thread blocked on `wait` is unblocked.  Everybody is relying on this behavior today whenever they use USM.

I think this makes your example well-defined: `*a = true` is sequenced-before `q.single_task`, which happens-before the start of the kernel (on the device), which is sequenced-before the end of the kernel function (on the device), which happens-before thread 2 is unblocked.

I don't know whether we intended for a query that returns `info::event_command_status::complete` to have the same behavior or not, and I don't know what the intent is with `empty()`.  I suspect we want them both to have some sort of synchronization behavior, but that's not clear from the specification.

gmlueck: > ...the implication being that `wait()` is some sort of synchronizing operation, or at least that we can guarantee the end of the kernel function happens-before the thread blocked on `wait` is unblocked. Everybody is relying on this behavior today whenever they use USM.
> 
> I think this makes your example well-defined: `*a = true` is sequenced-before `q.single_task`, which happens-before the start of the kernel (on the device), which is sequenced-before the end of the kernel function (on the device), which happens-before thread 2 is unblocked.

I agree that it's reasonable to assume that `queue::wait` is a synchronization point that ensures the memory written by kernels in the queue are visible to the calling thread.  I wasn't sure if it also guaranteed that memory written by *another thread* was visible to this thread.  I guess you are saying that normal C++ rules for "synchronizes with" would provide this guarantee because the write to `a` is "sequenced before" the kernel is submitted?

Your point about `info::event_command_status::complete` seems very relevant.  It seems like these two loops should provide similar guarantees (or lack of guarantees):

```
while (not q.empty()) {}

while (e.get_info<info::event::command_execution_status>() != info::event_command_status::complete) {}
```

Pennycook: > I agree that it's reasonable to assume that `queue::wait` is a synchronization point that ensures the memory written by kernels in the queue are visible to the calling thread. I wasn't sure if it also guaranteed that memory written by _another thread_ was visible to this thread. I guess you are saying that normal C++ rules for "synchronizes with" would provide this guarantee because the write to `a` is "sequenced before" the kernel is submitted?

Yes, exactly.  C++ doesn't have multiple devices, of course, so I'm trying to apply the rules as-if the device were just another thread.  My understanding is that "sequenced before" is more or less a fancy way of saying that "within a thread, things happen in program order", and then there are a bunch of transitivity rules (starting [here](https://eel.is/c++draft/intro.races#9)) that I interpret to mean something like "if A synchronizes with B, anything that happened before A must also have happened before B".

> Your point about `info::event_command_status::complete` seems very relevant. It seems like these two loops should provide similar guarantees (or lack of guarantees):

I agree they should behave the same.

gmlueck: > > Your point about `info::event_command_status::complete` seems very relevant. It seems like these two loops should provide similar guarantees (or lack of guarantees):
> 
> I agree they should behave the same.

How do we close on this?  Are we ready to clarify the behavior of the loop over `info::event::command_execution_status`?  Or, do we need to rework the memory model before this is possible?

My initial thought is that these loops are not synchronization points, so there is no guarantee about visibility of memory written by kernels in the queue.  What do others think?

There's another reason it might be good to require applications to call `wait` eventually.  In the DPC++ implementation, I think some internal resources aren't cleaned up until `wait` is called.  Therefore, if an application never called `wait` (and called `queue::empty` instead), these internal resources might never be cleaned up, causing a resource leak.

TApplencourt: Thanks a lot, @Pennycook. You put the correct word on my fuzzy feeling. The core issue is that memory visibility is indeed underspecified.

For the particular "empty" problem, I agree with both you. The description even spells out the `complete` command as "event returned by the event should be in the complete state" (of course, not really in this way, as now/ in the future, we can have command submission that doesn't return an event...) 

So I don't think we should change anything about `empty`. 

I may want to do more digging on how Cuda handles the "memory visibility". I just know that some people do the `while (not q.empty()) {}` to avoid call synchronize  ( I think now it `wait/cuda synchronize` has an API option to change the implementation details -- aka do you want it to spin lock - low latency high CPU usage, or more a "interrupt" implementation.  I remember discussing that with @abagusetty) 


> Are we ready to clarify the behavior of the loop over `info::event::command_execution_status`?

Do we want to say that if one access the `event` info, it will force the command submission?  


I think reworking "the memory model" seems better.

Without thinking too much, I would have thought that "wait" is a "memory-visible barrier" on top of a "synchronization barrier." So, it makes all the memory visible to everybody in the context. 

Inside the device code, you should then do "atomic" with the correct visibility or use a "memory visible barrier" (something like https://spec.oneapi.io/level-zero/1.9.3/core/api.html#zecommandlistappendmemoryrangesbarrier).

TApplencourt: > "For the purposes of Unified Memory, a return value of CUDA_SUCCESS is equivalent to having called cuEventSynchronize()."

So, by similarity, that means accessing the `status event` field should force the command submission. I'm OK with that personally. I'm not sure it's now required by the spec.

Pennycook: > So, by similarity, that means accessing the `status event` field should force the command submission. I'm OK with that personally. I'm not sure it's now required by the spec.

I don't think that's what CUDA is guaranteeing here, but the wording isn't super clear.  My interpretation of what they're saying is that _if_ you observe a value of `CUDA_SUCCESS`, then the effect is the same as if you had called `cuEventSynchronize`.  In other words: if you observe that the kernel is complete, any memory side-effects are visible.

So for SYCL, we can view this as two (very closely related) questions, where one concerns the execution model and one concerns the memory model:
- **Execution Model**: Does querying an event status force the runtime to start executing the event (eventually)?
- **Memory Model**: Does observing a completed event status imply the command's side-effects are visible?

I think the answer to the first question can be "No", even if the answer to the second question is "Yes".

TApplencourt: True. 2 Independent questions indeed.

For 1/  At least, from experience, I know "real" apps that do the spin lock of `cuEventQuerry`, and they never deadlock.   Not sure if it's because cuda is always greedy, or because `cuEventQuerry` force submission.  (I guess (2) it's an implementation details of (1)).

So I think we should say "yes." It will "please" people porting too Cuda or people who are "latency bound," I suppose. 
```
for ...
  Q.submit();  
   Q.empty();
Q.wait()
```
Should be faster than:
```
for ...
  Q.submit();  
   Q.wait();
```


For 2/, definitely yes.  


But my idea is that if checking for an event doesn't submit, it force you always  call `wait` so question 2/ is "useless", or a totology.

You always call wait (because this is the only wait to submit), so the command's side effects are always visible, regardless of if you observe  `completed event `. 

Hope it's kind of clean

Pennycook: > So I think we should say "yes." It will "please" people porting too Cuda or people who are "latency bound," I suppose.

I'm not sure we want to do this. I might be wrong, but my gut says that requiring implementations to start executing kernels when an event is queried would prevent single-threaded implementations (or at least make them a lot more complicated).  The only way to avoid deadlocks would be to implement a mechanism to switch between host and device code _during_ kernel execution, which I think would be difficult in the general case.

> But my idea is that if checking for an event doesn't submit, it force you always call wait so question 2/ is "useless", or a totology.
>
> You always call wait (because this is the only wait to submit), so the command's side effects are always visible, regardless of if you observe `completed event `. And you cannot see an event-completed without calling wait.

`wait` is the only way to _guarantee_ that a kernel is started, but the opposite is not true; there is no guarantee that a kernel will not start until `wait` is called.

So, a spin-loop querying an event _might_ work, it's just implementation-defined.  Such a loop would definitely fail for a single-threaded implementation like SimSYCL (because the kernel will never start), but would probably work when offloading to an accelerator via CUDA/OpenCL/Level Zero (because the kernel will probably start).

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,78 @@
...
+which provides information about the emptiness of the queue.
+
+'''
+
+.[apidef]#queue::khr_empty#
+[source,role=synopsis,id=api:queue-khr-empty]
+----
+bool khr_empty() const
+----
+
```
Pennycook, line 42: Here is my suggested wording for the synchronization behavior:
```suggestion

_Synchronization_:  When this function returns [code]#true#, equivalent to [code]#queue::wait()#.

```

We still need to revisit and clarify the synchronization behavior of `queue::wait`, but I think we're agreed that `empty()` should behave equivalently.

adoc/extensions/sycl_khr_queue_empty_query.adoc
Pennycook: @TApplencourt - I don't think we should say anything about the execution model in this KHR, so I propose that we defer that discussion until later.  I've proposed some wording in https://github.com/KhronosGroup/SYCL-Docs/pull/700/files#r1942646351 to address the memory visibility aspect, though, because I think we are in agreement that the answer to the second (Memory Model) question needs to be "Yes".

TApplencourt: Thanks a lot! Merged.

TApplencourt:
Always reflow, always...

keryell (approved):
Thanks!

tomdeakin:
WG approved to merge.

tomdeakin:
Waiting confirmation on implementations

gmlueck:
This is the Intel internal tracker to implement in DPC++: CMPLRLLVM-65342

TApplencourt:
@illuhad All good? Thumbs up? Green light? :D

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -0,0 +1,81 @@
...
+_Synchronization_: When this function returns [code]#true#, equivalent to
+[code]#queue::wait()#.
+
+_Returns:_ [code]#true# if all <<command,commands>> enqueued on this queue have
```
illuhad, line 46: Should we perhaps add a note to say that repeatedly calling `khr_empty()` does not guarantee progress on its own? From the discussion below I think there is agreement that SYCL implementations might execute commands lazily such that such patterns may not work.

I'm a bit concerned that the equivalence to `queue::wait()` might be misinterpreted by users as providing similar progress guarantees.
I know that we're not saying this here, it's just that I wouldn't be surprised if users make this assumption and feel confirmed when they read what we write.

illuhad (commented):
I've added an implementation in AdaptiveCpp that I *think* is correct. https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1790

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Good idea. 

Yep, this lazy / greepy discussion belong to a new KHR (adding querry / property of the queue or something)

Will work on phrasing! 

Thanks!

Pennycook: We could say something like:
> _Synchronization_: When this function returns `true`, synchronization behavior is equivalent to `queue::wait()`.

...but this seems unnecessarily verbose to me, because the whole paragraph is about synchronization behavior.

If we want to add a note about the lack of progress, I think we should add a more general note to the core specification.  Repeatedly calling `khr_empty()` is the same as repeatedly checking the status of one or more `sycl::event` objects, but I don't think we explicitly say anywhere that querying an event's status does not guarantee progress.

illuhad: > If we want to add a note about the lack of progress, I think we should add a more general note to the core specification. Repeatedly calling khr_empty() is the same as repeatedly checking the status of one or more sycl::event objects, but I don't think we explicitly say anywhere that querying an event's status does not guarantee progress.

I agree that it's the same as repeatedly checking the completion status of an event, and I'm also not aware that we say something there, *but* we also don't draw any comparisons to `queue::wait()` there either. This is the important bit in my opinion.

In any case, this is not a hill I'm willing to die on, just a suggestion to avoid misinterpretation of the spec by users.

TApplencourt: I agree with you both. We should put a note somewhere on the "greedy / lazy -ness"  and that "querry status"  doesn't trigger a submission. 

(like we did for free, and blocking). 

I will need to think more about where it makes sense to put it!  

The `while Q.empy` is the "traditional" CUDA pattern, so it makes sense to put it here. But the problem is more general, so make sense to also put it on the spec :')

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: Ok fixed this one with the `https://github.com/KhronosGroup/SYCL-Docs/pull/808`

adoc/extensions/sycl_khr_queue_empty_query.adoc
TApplencourt: @illuhad I put a simple note to say that it works best with `khr_flush`. So people can read it to get more information.
Not perfect, but it was too hard to write a sentence to explain the greedy/lazy in this KHR for my little brain.

It was a good point, that this this is mostly useless / implementation speficit,  without flush

tomdeakin:
WG approved to merge.

PR #701, opened 2025-01-29
Fix 2 kernel in submit in code example

### LINKED ISSUE #699: `4.12.2. Defining kernels as lambda functions` example is wrong
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interfaces.kernels.as.lambdas

AFAIK cannot have 2 `single_task`  in the same submit.

TApplencourt:
fix #699

### DIFF
--- a/adoc/code/mykernel.cpp
+++ b/adoc/code/mykernel.cpp
@@ -9,10 +9,12 @@ myQueue.submit([&](handler& h) {
   h.single_task<MyKernel>([=] {
     // [kernel code]
   });
+});
 
-  // Explicitly name kernel without forward declaring type at
-  // namespace scope.  Must still be forward declarable at
-  // namespace scope, even if not declared at that scope
+// Explicitly name kernel without forward declaring type at
+// namespace scope.  Must still be forward declarable at
+// namespace scope, even if not declared at that scope
+myQueue.submit([&](handler& h) {
   h.single_task<class MyOtherKernel>([=] {
     // [kernel code]
   });


adoc/code/mykernel.cpp
```diff
@@ -4,16 +4,13 @@
 // Explicit kernel names can be optionally forward declared at namespace scope
 class MyKernel;
 
-myQueue.submit([&](handler& h) {
-  // Explicitly name kernel with previously forward declared type
-  h.single_task<MyKernel>([=] {
+myQueue.single_task<MyKernel>([=] {
```
gmlueck, line 7: You have also removed the calls to `submit` and replaced that with the shortcut function `queue::single_task`.  I don't think we want to do that here.  This section focuses on kernel naming, not the shortcut functions.  In addition, the example code just below this uses the `submit` style, so it would be better to keep a consistent style across the two examples.

TApplencourt: Oh ok. I think the other is verbose for no reason, and by putting verbose code an example, we encourage people to do so.

I will vote to change them all. As you said, we only care about the `naming` of the kernel, not how they are launched (via implicit or explicit cgh)

gmlueck: It seems to me that we mostly use the separate `submit` / `parallel_for` style throughout the specification unless we are specifically illustrating the shortcut functions.  For example our canonical example in ["Example SYCL application"](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_example_sycl_application) uses this style.

TApplencourt: Yes because we use buffer /accessor , so no other choice. We should change that too.

But I agree it's some work for another PR. Will reverse back to the submit  + x

TApplencourt: Done [9d88e6f](https://github.com/KhronosGroup/SYCL-Docs/pull/701/commits/9d88e6f7e3217ff29764dee3d198dafc2a98ae45)

gmlueck (approved):

adoc/code/mykernel.cpp
gmlueck: Thanks!

tomdeakin (approved):
Clearly Editorial, please merge.

gmlueck:
@TApplencourt: GitHub thinks you have not signed the CLA.  I think this is because you used a different email address in your commits.  Can you follow the instructions in the comment above about [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).  If that isn't possible, maybe just squash your commits with the correct email address and then force-push to this PR.

TApplencourt:
Yeah, I guess `applenco@aurora-uan-0012.hostmgmt.cm.aurora.alcf.anl.gov` is not a super nice email address to add. I squashed and force-pushed for my Mac. It should be good now.

PR #712, opened 2025-02-11
[KHR] Add sycl_khr_max_num_work_groups extension

Aympab:
This PR proposes to add to the specification a device descriptor on the maximum number of work-groups that can be submitted to a `range` or `nd_range` parallel for: `max_num_work_groups_nd_range<N>` and `max_num_work_groups_range<N>` for N=1, 2 or 3. The device query returns a tuple `id<N>` with boundaries in each dimension.

## Justification

- In the current revision of the spec, there is no limitation on the maximum iteration size submitted to a parallel for, supposedly meaning that any number should be valid. But when actually running with large sizes, backend-related issues emerge.

- Users rely on these values to check kernel boundaries and often have to hard-code these values, for example Kokkos developers with the SYCL backend (see [this PR](https://github.com/kokkos/kokkos/pull/7724)) or this implementation of [blocking/streaming kernels](https://github.com/Maison-de-la-Simulation/parallel-advection/blob/bkm-paper/src/core/advectors.h#L67-L75).

- The query is already available for all GPU backends:
  - [CUDA](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g9c3e1414f0ad901d3278a4d6645fc266) `CU_DEVICE_ATTRIBUTE_MAX_GRID_DIM_[X,Y,Z]`
  - [hip](https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html#ga7080a145a4239a7276e0dc22062026c1) `hipDeviceAttributeMaxGridDim[X,Y,Z]`
  - [L0](https://spec.oneapi.io/level-zero/1.9.3/core/api.html#ze__api_8h_1ae0e7f575e41e2576ffa0269d886e55f4) `ze_device_properties_t::maxGroupCount[X,Y,Z]`

- DPC++ already implements something similar as [an extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_max_work_group_query.md)

## Notes
- There is a distinction for `nd_range` or `range` for multiple reasons:
  - The semantics of range and nd_range differ which might impact the maximum size of the iteration space
  - Implementers can choose to map higher-level optimization methods for basic range kernels, or directly map on the lower level limitations

- For N=3, the mapping is straightforward, as it directly queries the backend functions with `X, Y, Z`. The mapping for N=1 and N=2 is unclear, implementers could choose the minimum of `X,Y,Z` or compute a product of other dimensions, for example.

- Although this PR is largely inspired by DPC++ extension, they initially proposed `max_global_work_groups` which is not actually queryable with CUDA/HIP which is why it is not proposed here.
  - As a user, the main concern is ensuring that the kernel’s iteration range does not exceed the maximum values for each dimension. The rest should be implementation-defined (e.g., whether the mapping is direct or if blocking/streaming is used).

### DIFF
--- a/adoc/extensions/index.adoc
+++ b/adoc/extensions/index.adoc
@@ -11,3 +11,4 @@ specification, but their design is subject to change.
 // include::sycl_khr_extension_name.adoc[leveloffset=2]
 
 include::sycl_khr_default_context.adoc[leveloffset=2]
+include::sycl_khr_max_num_work_groups.adoc[leveloffset=2]
--- a/adoc/extensions/sycl_khr_max_num_work_groups.adoc
+++ b/adoc/extensions/sycl_khr_max_num_work_groups.adoc
@@ -1,0 +1,92 @@
+[[sec:khr-max-num-work-groups]]
+= sycl_khr_max_num_work_groups
+
+This extension allows developers to query iteration bounds in each dimension for a ND-range or basic range kernel.
+The implementation ensures the execution of the ND-range kernel if its global size is less than of equal to `max_num_work_groups_nd_range<N>` in each dimension. This condition applies to basic range kernels with `max_num_work_groups_range<N>`.
+
+
+[[sec:khr-max-num-work-groups-dependencies]]
+== Dependencies
+
+This extension does not depend on other extensions.
+
+[[sec:khr-max-num-work-groups-feature-test]]
+== Feature test macro
+An implementation supporting this extension must predefine the `SYCL_KHR_MAX_WORK_GROUP_QUERY` macro to one of the values defined in the table below.
+
+[%header,cols="1,5"]
+|===
+|Value
+|Description
+
+|1
+|Initial version of this extension.
+|===
+
+== New device descriptors
+
+[options="header"]
+[cols="1,1,2", options="header"]
+|===
+| Device descriptors                                     | Return type | Description
+
+| `khr::info::device::max_num_work_groups_nd_range<N>`
+| `id<N>`
+| Returns the maximum number of work-groups that can be submitted in each dimension of the `globalSize` of an `nd_range<N>`. The minimum value is `(1, 1, 1)` if the device is not `info::device_type::custom`.
+
+
+| `khr::info::device::max_num_work_groups_range<N>`
+| `id<N>`
+| Returns the maximum number of work-groups that can be submitted in each dimension of a `range<N>` for a basic kernel. The minimum value is `(1, 1, 1)` if the device is not `info::device_type::custom`.
+|===
+
+=== Note
+- N can take the value 1, 2 or 3
+// - nd_range allow barrier, and other scemantic who may impact the maximun size of allocation, hence we split the querry between the range and nd_range?
+
+[[sec:khr-max-num-work-groups-example]]
+== Example
+
+The example below demonstrates the use of this extension for an `nd_range` loop, the same code can be applied to a basic range by using the `max_num_work_group_range` device descriptor.
+
+[source,cpp]
+----
+
+#include <iostream>
+#include <sycl/sycl.hpp>
+
+int
+main(int argc, char *argv[]) {
+    size_t dim1 = std::stoi(argv[1]);
+    size_t dim2 = std::stoi(argv[2]);
+    size_t dim3 = std::stoi(argv[3]);
+    sycl::range<3> globalSize(dim1, dim2, dim3);
+    sycl::range<3> localSize(1, 1, 1);
+
+    sycl::queue queue;
+    sycl::device device = queue.get_device();
+    std::cout << "Running on " << device.get_info<sycl::info::device::name>()
+              << "\n";
+
+    sycl::id<3> nd_limit =
+        gpu.get_info<sycl::khr::info::device::max_num_work_group_nd_range<3>>();
+    std::cout << "Max number groups for ND-Range: x_max: " << nd_limit[2]
+              << " y_max: " << nd_limit[1] << " z_max: " << nd_limit[0]
+              << std::endl;
+
+    // Should aways be satisfied at kernel submission
+    // user's responsibility to check
+    assert(globalSize[2] <= nd_limit[2]
+        && globalSize[1] <= nd_limit[1]
+        && globalSize[0] <= nd_limit[0]);
+
+    // If the assertion if satisfied, the implementation guarantees
+    // the execution of the kernel
+    queue.parallel_for(
+        sycl::nd_range{globalSize, localSize},
+        [=](sycl::id<3> idx) { /*Kernel*/ }).wait();
+
+    return 0;
+}
+
+----


adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,92 @@
+[[sec:khr-max-num-work-groups]]
+= sycl_khr_max_num_work_groups
+
+This extension allows developers to query iteration bounds in each dimension for a ND-range or basic range kernel.
+The implementation ensures the execution of the ND-range kernel if its global size is less than of equal to `max_num_work_groups_nd_range<N>` in each dimension. This condition applies to basic range kernels with `max_num_work_groups_range<N>`.
```
Pennycook, line 5: I don't think it makes sense to expose this query separately for ND-range and basic (range) kernels.

According to [Section 3.7.2, SYCL kernel execution mode](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_kernel_execution_model), work-groups only exist for ND-range kernels.  A basic kernel is launched only with a number of work-items -- there is no way to specify the number of work-groups to use, nor to query the number of work-groups used by the implementation.

This is why the DPC++ extension only defines `max_work_groups` for ND-range kernels.  I think we should align with that design if we're going to pursue this feature as a KHR.

Pennycook (changes requested):

adoc/extensions/sycl_khr_max_num_work_groups.adoc
Aympab: Indeed it doesn't make sense to talk about "work-groups" for a basic kernels.
Still, the user cannot check the bounds of a basic range kernel before submission and I think we should be able to query that so the assertion still remains valid.

Maybe a renaming `max_num_work_groups_range<N>` --> `max_global_size_range<N>` or `max_basic_range<N>`? Or maybe this value already exists in the specification?

Pennycook: There is not a device-level query, but users can already query the maximum range for a specific kernel using `info::kernel_device_specific::global_work_size`, which is defined in [Table 135](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.kernel.devicespecificinfo).

Aympab: I updated the PR by removing `max_num_work_groups_range`. 
I also renamed `max_num_work_groups_nd_range<N>` --> `max_num_work_groups<N>` since it's implicit that it is the maximum for an ND-Range kernel (what about hierarchical kernels, should it be the same?)

Let me know if there should be other changes

TApplencourt: >There is not a device-level query, but users can already query the maximum range for a specific kernel using info::kernel_device_specific::global_work_size

I totally forgot about this query. Sorry, @Aympab! Looking at the table  It's always return `range<3>`. So if I want to submit `range<1>`, I don't know the math to use to linearize it. Should we then add new queries to the 'kernel_device_specific`?

illuhad:
> For N=3, the mapping is straightforward, as it directly queries the backend functions with X, Y, Z. The mapping for N=1 and N=2 is unclear, implementers could choose the minimum of X,Y,Z or compute a product of other dimensions, for example.

I just noticed that this KHR only proposes to add the N=3 case. I'm not sure I understand the problem about N=1 and N=2. Couldn't we just map those to the limits in the corresponding dimensions, given that backends seem to have separate limits for each of the dimensions in the APIs that are listed:

> [CUDA](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g9c3e1414f0ad901d3278a4d6645fc266) CU_DEVICE_ATTRIBUTE_MAX_GRID_DIM_[X,Y,Z]
    [hip](https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html#ga7080a145a4239a7276e0dc22062026c1) hipDeviceAttributeMaxGridDim[X,Y,Z]
    [L0](https://spec.oneapi.io/level-zero/1.9.3/core/api.html#ze__api_8h_1ae0e7f575e41e2576ffa0269d886e55f4) ze_device_properties_t::maxGroupCount[X,Y,Z]

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+== New device descriptors
+
+[options="header"]
+[cols="1,1,2", options="header"]
+|===
+| Device descriptors                                     | Return type | Description
+
+| `khr::info::device::max_num_work_groups<3>`
+| `id<3>`
```
illuhad, line 34: Shouldn't the return type be `range<3>` instead of `id<3>`?

Pennycook:
> I just noticed that this KHR only proposes to add the N=3 case. I'm not sure I understand the problem about N=1 and N=2. Couldn't we just map those to the limits in the corresponding dimensions, given that backends seem to have separate limits for each of the dimensions in the APIs that are listed:
> 
> > [CUDA](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__DEVICE.html#group__CUDA__DEVICE_1g9c3e1414f0ad901d3278a4d6645fc266) CU_DEVICE_ATTRIBUTE_MAX_GRID_DIM_[X,Y,Z]
> > [hip](https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html#ga7080a145a4239a7276e0dc22062026c1) hipDeviceAttributeMaxGridDim[X,Y,Z]
> > [L0](https://spec.oneapi.io/level-zero/1.9.3/core/api.html#ze__api_8h_1ae0e7f575e41e2576ffa0269d886e55f4) ze_device_properties_t::maxGroupCount[X,Y,Z]

I agree with Aksel.  I understand that implementations have flexibility here, because the specification doesn't require any specific mapping of SYCL dimensions to backend dimensions (except for OpenCL), but I don't think that's a specification issue.  Every implementation should be able to provide a number for the 1- and 2-dimensional versions of this query, and it's the implementation's job to ensure that the way the implementation executes work-groups is consistent with the result of these queries.

Aympab:
> it's the implementation's job to ensure that the way the implementation executes work-groups is consistent with the result of these queries.

I agree. The initial idea was to avoid any ambiguity for the implementers and to have a direct mapping, which is 3D (I don't think you can get a 1D or 2D query with the backends?). But we can set N = 1, 2, or 3 if the specification stipulates that the only condition is to guarantee the execution of an N-D kernel, as long as we don't exceed the N-D limits in each dimension, regardless of whether N = 1, 2, or 3. Then, implementers can apply any abstractions as long as they satisfy this condition. 

I updated the PR for N = 1, 2 ,3

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,86 @@
+[[sec:khr-max-num-work-groups]]
+= sycl_khr_max_num_work_groups
+
+This extension allows developers to query iteration bounds in each dimension for a ND-range kernel.
+The implementation ensures the execution of the ND-range kernel of `N` dimension(s) if its number of work-groups is less than of equal to `max_num_work_groups<N>` in each dimension.
```
Aympab, line 5: I am also wondering if we should specify something like:
" The number of work-groups in dimension `n_i` corresponds to `globalSize[n_i]/localSize[n_i]`" since the NDRange doesn't provide an API to directly submit a number of work-groups (as does/did Hierarchical). Or should we just leave it like that? I think it needs to be clear that this limit corresponds to a number of work-groups submitted rather than a global iteration size.

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,91 @@
...
+|===
+| Device descriptors                                     | Return type | Description
+
+| `khr::info::device::max_num_work_groups<N>`
+| `id<N>`
+| Returns the maximum number of work-groups that can be submitted in each dimension of an `nd_range<N>`. The minimum value is 1 in each dimension if the device is not `info::device_type::custom`.
+
+|===
+
+Note: `N` can take the value 1, 2 or 3.
```
tomdeakin, line 39: Is this a normative, or a non-normative note? If non-normative, this should be a constraint.

tomdeakin (changes requested):

adoc/extensions/sycl_khr_max_num_work_groups.adoc
Aympab: I removed the note and now have an array with 3 rows, just like `max_work_item_sizes<N>` in the [device informations descriptors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors)

Aympab:
I've updated the return type of the function to be `range<N>` instead of `id<N>` to match `max_work_item_sizes<N>` which returns a range.

Aympab:
I've just opened a [PR for the CTS](https://github.com/KhronosGroup/SYCL-CTS/pull/1073)

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,100 @@
+[[sec:khr-max-num-work-groups]]
+= sycl_khr_max_num_work_groups
+
+This extension allows developers to query iteration bounds in each dimension for a ND-range kernel.
+The implementation ensures the execution of the ND-range kernel of `N` dimension(s) if its number of work-groups is less than or equal to `max_num_work_groups<N>` in each dimension.
```
gmlueck, line 5:
```suggestion
The implementation ensures the execution of the ND-range kernel of `N` dimension(s) only if its number of work-groups is less than or equal to `max_num_work_groups<N>` in each dimension.
```

Or, maybe even better:

```suggestion
The application must ensure that the number of work-groups in each dimension `N` of an ND-range kernel is
less than or equal to `max_num_work_groups<N>`.
```

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+| `khr::info::device::max_num_work_groups<2>`
+| `range<2>`
+| Returns the maximum number of work-groups that can be submitted in each dimension of an `nd_range<2>`. The minimum value is 1 in each dimension if the device is not `info::device_type::custom`.
+
+| `khr::info::device::max_num_work_groups<3>`
+| `range<3>`
+| Returns the maximum number of work-groups that can be submitted in each dimension of an `nd_range<3>`. The minimum value is 1 in each dimension if the device is not `info::device_type::custom`.
+
+|===
```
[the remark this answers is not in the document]
Aympab: I've changed the "New device descriptors" section to match the new style.
Still, this is a KHR extension so the final namespace has the `::khr` namespace extension to it.

gmlueck (commented):
Thanks for updating the style.  Here are a few more editorial comments.

adoc/extensions/sycl_khr_max_num_work_groups.adoc
```diff
@@ -0,0 +1,100 @@
...
+
```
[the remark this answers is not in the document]
gmlueck: If CUDA does not have a similar limit, wouldn't it be a simple matter for the implementation to return `SIZE_MAX` for that backend?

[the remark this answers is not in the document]
Aympab: I see, in this case wouldn't it be more appropriate to have a separate device descriptor with something like `max_linear_id`?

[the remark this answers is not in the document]
Aympab: Oh okay sorry! I thought you meant to add it in this PR.
I'm fine with `max_total_work_groups`, do you want me to open a new PR for this?

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,115 @@
...
+};
+} // namespace sycl::info::device
+----
+
+_Remarks:_ Template parameter to [api]#device::get_info#.
+
+_Constraints:_ Available only when [code]#Dimensions# is 1, 2 or 3.
+
+_Returns:_ The maximum number of work-groups that can be submitted in each dimension of an [code]#nd_range<Dimensions># kernel. The minimum value is 1 in each dimension if the device is not [code]#info::device_type::custom#.
+
```
gmlueck: Add a horizontal line here:

```
'''
```

(Three single-quote characters with a blank line before and after.)

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,115 @@
...
+struct max_total_work_groups {
+    using return_type = size_t;
+};
+} // namespace sycl::info::device
+----
+
+_Remarks:_ Template parameter to [api]#device::get_info#.
+
+_Returns:_ The total maximum number of work-groups that can be submitted to an ND-range kernel. The minimum value is 1 if the device is not [code]#info::device_type::custom#.
+
```
gmlueck: And put a horizontal, line here too.

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,115 @@
...
+.[apidef]#khr::info::device::max_work_group_sizes#
+[source,role=synopsis,id=api:khr-info-device-max-work-group-sizes]
+----
+namespace sycl::khr::info::device {
+template<int Dimensions = 3>
+struct max_work_group_sizes {
+    using return_type = range<Dimensions>;
+};
+} // namespace sycl::info::device
```
gmlueck:
```suggestion
} // namespace sycl::khr::info::device
```

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,115 @@
...
+_Returns:_ The maximum number of work-groups that can be submitted in each dimension of an [code]#nd_range<Dimensions># kernel. The minimum value is 1 in each dimension if the device is not [code]#info::device_type::custom#.
+
+.[apidef]#khr::info::device::max_total_work_groups#
+[source,role=synopsis,id=api:khr-info-device-max-total-work-groups]
+----
+namespace sycl::khr::info::device {
+struct max_total_work_groups {
+    using return_type = size_t;
+};
+} // namespace sycl::info::device
```
gmlueck:
```suggestion
} // namespace sycl::khr::info::device
```

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+    const auto nb_wg_2 = globalSize[2]/localSize[2];
+    const auto nb_wg_1 = globalSize[1]/localSize[1];
+    const auto nb_wg_0 = globalSize[0]/localSize[0];
+
+    // Should always be satisfied at kernel submission
+    // user's responsibility to check
+    if (nb_wg_0 <= nd_limit[2]
+     && nb_wg_1 <= nd_limit[1]
+     && nb_wg_2 <= nd_limit[0]) {
```
gmlueck: We should also check here that the total number of work-groups does not exceed `max_total_work_groups`.

gmlueck (commented):
> Also, should we rename the name of the file and KHR proposal ? From `sycl_khr_max_work_group_sizes` to something like `sycl_khr_nd_range_limits`?

Actually, the current name is confusing because the query tells you the maximum number of work-groups, not the maximum size of a work group.  Maybe `sycl_khr_max_work_groups` would be better, and it would cover the new query too.

Aympab:
> > Also, should we rename the name of the file and KHR proposal ? From `sycl_khr_max_work_group_sizes` to something like `sycl_khr_nd_range_limits`?
> 
> Actually, the current name is confusing because the query tells you the maximum number of work-groups, not the maximum size of a work group. Maybe `sycl_khr_max_work_groups` would be better, and it would cover the new query too.

I agree the `_sizes` is confusing. It is also for `max_work_item_sizes` [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:info-device-max-work-item-sizes) and `max_work_group_size` [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:info-device-max-work-group-size).

Actually, these two requests are the same matter at a different parallel hierarchy level. I think we should align both these queries in someway. I personally like the `max_total_work_[groups|items]` for the total maximum and `max_work_[groups|items]<N>` for the maximum in each dimension.

Maybe for this PR it is better to align with the existing queries and then make another update for the "renaming part", where we rename all these 4 queries? I have no idea about how this is done.

I will wait for your instructions before updating the PR.

adoc/extensions/sycl_khr_max_work_group_sizes.adoc
```diff
@@ -0,0 +1,122 @@
...
+    const auto nb_wg_2 = globalSize[2]/localSize[2];
+    const auto nb_wg_1 = globalSize[1]/localSize[1];
+    const auto nb_wg_0 = globalSize[0]/localSize[0];
+    const auto nb_wg_total = nb_wg_0*nb_wg_1*nb_wg_2;
+
+    // Should always be satisfied at kernel submission
+    // user's responsibility to check
+    if (nb_wg_0 <= nd_limit[2]
+     && nb_wg_1 <= nd_limit[1]
+     && nb_wg_2 <= nd_limit[0]
```
gmlueck: Sorry, I just noticed this.  Is this correct that you compare `nb_wg_2` with `nd_limit[0]`?  Shouldn't you compare with `nd_limit[2]`?  Are the indices reversed for some reason?

gmlueck:
> I agree the `_sizes` is confusing. It is also for `max_work_item_sizes` [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:info-device-max-work-item-sizes) and `max_work_group_size` [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:info-device-max-work-group-size).

Yes, the naming is a mess!  Actually, the existing `max_work_group_size` is reasonable because it tells the number of work-items in the work-group.  However, the existing `max_work_item_sizes` makes no sense because it tells the number of work-items in a work-group (not the size of a work-item).

In cases like this, I think it's best to plan where we want to be and then figure out how to get there.  For an end state, how about these names:

* `device::max_work_group_sizes` - The maximum number of work-items in each dimension of a work-group.  (This is the existing `device::max_work_item_sizes`).
* `device::max_total_work_group_size` - The maximum total number of work-items in a work-group across all dimensions.  (This is the existing `device::max_work_group_size`).
* `device::max_work_groups` - The maximum number of work-groups in each dimension of an nd-range kernel.  (This is your proposed `device::max_work_group_sizes`).
* `device::max_total_work_groups` - The maximum total number of work-groups in an nd-range kernel across all dimensions.  (This is your proposed `device::max_work_group_size`).
* `kernel_device_specific::max_total_work_group_size` - The maximum total number of work-items in a work-group in all dimensions for a specific kernel running on a specific device.  (This is the existing `kernel_device_specific::work_group_size`).

This would allow us to define this in the future if we want:

* `device::max_total_work_items` - The maximum number of work-items in an nd-range kernel across all dimensions.

@Pennycook: How do these proposed names line up with the names you have in sycl_khr_group_interface?

FWIW, I think we inherited some of these confusing names from OpenCL.  For example, OpenCL has [`CL_DEVICE_MAX_WORK_ITEM_SIZES`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_WORK_ITEM_SIZES), which tells the maximum number of work-items in each dimension of a work-group.  So you can see that the existing SYCL names are aligned with the OpenCL names.  We will need to decide if consistency with the OpenCL names is more important, or if it's better to come up with new names that make more sense.

adoc/extensions/index.adoc
```diff
@@ -12,4 +12,4 @@ specification, but their design is subject to change.
 
 include::sycl_khr_default_context.adoc[leveloffset=2]
 include::sycl_khr_queue_empty_query.adoc[leveloffset=2]
-include::sycl_khr_group_interface.adoc[leveloffset=2]
+include::sycl_khr_max_work_group_sizes.adoc[leveloffset=2]
```
gmlueck: This looks like a merge mistake.  Your new KHR should appear after "sycl_khr_group_interface".  It should not remove that KHR.

Pennycook:
I also agree that the names are confusing.  We've tried to do better in some of our extensions, but the lack of a clear naming scheme makes it hard to argue that we've succeeded 😅 .

> @Pennycook: How do these proposed names line up with the names you have in sycl_khr_group_interface?

I don't think they line up super well.  Using "sizes" as a plural seems wrong to me, too -- in my mental model, a work-group only has one size.  The only other place we use "sizes" that I'm aware of is the `sub_group_sizes` query; using a plural "sizes" in the case of sub-groups makes sense, because we're returning a list of sizes.

The groups KHR tries to align with ISO C++, and uses "size" to mean "the total number of items in a group" while using "extents" to mean "the total number of items in each dimension of a group".  I hadn't thought of it before, but I think you're onto something -- aligning with the KHR here _would_ make some of the names less confusing.  We could also use it as an opportunity to create a more obvious link between the queries and the APIs impacted by them.

Building on what you wrote above, these would make sense to me and seem pretty obvious:

- `device::max_work_group_extents` - The maximum number of work-items in each dimension of a work-group, _and the maximum number that can be returned by `work_group::extents()`_. (This is the existing `device::max_work_item_sizes`).

- `device::max_work_group_size` - The maximum total number of work-items in a work-group across all dimensions, _and the maximum number that can be returned by `work_group::size()`_. (This is the existing `device::max_work_group_size`, and its name wouldn't change).

- `kernel_device_specific::max_work_group_size` - The maximum total number of work-items in a work-group in all dimensions for a specific kernel running on a specific device, _and the maximum number that can be returned by `work_group::size()` for that combination of kernel and device_. (This is the existing `kernel_device_specific::work_group_size`).

The ones for the "number of work-groups" are less clear, and seem like they might be contentious.  I'll just write them out with `id` -- we could use `range` instead and the meaning would be the same.  I have a strong preference for `id`, because I've never liked the way that SYCL uses `range`.

- `device::max_work_group_id` - The maximum number of work-groups in each dimension of an nd-range kernel, _and the maximum number that can be returned by `work_group::id()`_. (This is the proposed `device::max_work_group_sizes`).
- `device::max_work_group_linear_id` - The maximum total number of work-groups in an nd-range kernel across all dimensions, _and the maximum number that can be returned by `work_group::linear_id()`_. (This is the proposed `device::max_work_group_size`).

These are unambiguous, and we could extend them to other group types (e.g., sub-groups, root-groups) in an obvious way.

Pennycook:
> Ok, if I understand correctly I will update this PR with what @Pennycook proposed:
> 
> * `max_work_group_id<N>`  instead of `max_work_group_sizes<N>` (I'm also updating the macro name to `SYCL_KHR_MAX_WORK_GROUP_QUERY`)
> * `max_work_group_linear_id` instead of `max_total_work_groups`
> * filename and apidef as `max-work-group-query`
> 
> And maybe open another PR to update the existing names?

Sorry... You understood correctly, but I just realized there's an off-by-one error in my proposal. ☹️ 

An id is in [0, range) so, strictly speaking, `max_work_group_id` doesn't represent the maximum number of work-groups.  I think we need to either:

1. Keep `max_work_group_id`, say that it returns the maximum `id`, but drop the part about that being "the total number of work-groups".  The maximum number of work-groups somebody could launch would be the value returned by this query + 1.

2. Change it to `max_work_group_range`, say that it returns the maximum value of `work_group::range`, and keep the part about it being "the total number of work-groups".  The maximum number of work-groups somebody could launch would be the value returned by this query.

@gmlueck, do you have an opinion on what's clearer here?  I still don't really like using "range" like this, but I'm worried that requiring developers to remember the +1 might be too confusing.

gmlueck:
> @gmlueck, do you have an opinion on what's clearer here? I still don't really like using "range" like this, but I'm worried that requiring developers to remember the +1 might be too confusing.

The part about "+1" is definitely too confusing.  The query should return the "total number of work-groups".

adoc/extensions/sycl_khr_max_work_group_query.adoc
```diff
@@ -0,0 +1,122 @@
+[[sec:khr-max-work-group-query]]
+= sycl_khr_max_work_group_id
```
gmlueck:
```suggestion
= sycl_khr_max_work_group_query
```

I assume this was just a typo?

adoc/extensions/sycl_khr_max_work_group_query.adoc
```diff
@@ -0,0 +1,122 @@
...
+    using return_type = range<Dimensions>;
+};
+} // namespace sycl::khr::info::device
+----
+
+_Remarks:_ Template parameter to [api]#device::get_info#.
+
+_Constraints:_ Available only when [code]#Dimensions# is 1, 2 or 3.
+
+_Returns:_ The maximum number of work-groups that can be submitted in each dimension of an [code]#nd_range<Dimensions># kernel. The minimum value is 1 in each dimension if the device is not [code]#info::device_type::custom#.
```
gmlueck:
```suggestion
_Returns:_ The maximum number of work-groups that can be submitted in each dimension of an [code]#nd_range<Dimensions># kernel. The minimum value is 1 in each dimension if the device is not [api]#info::device_type::custom#.
```

Kind of an editorial nit, but this will make it a link to the definition.

Aympab:
I'm not sure what's wrong about using `range`?

adoc/extensions/sycl_khr_max_work_group_query.adoc
```diff
@@ -0,0 +1,122 @@
...
+namespace sycl::khr::info::device {
+struct max_work_group_linear_id {
+    using return_type = size_t;
+};
+} // namespace sycl::khr::info::device
+----
+
+_Remarks:_ Template parameter to [api]#device::get_info#.
+
+_Returns:_ The total maximum number of work-groups that can be submitted to an ND-range kernel. The minimum value is 1 if the device is not [code]#info::device_type::custom#.
```
gmlueck:
```suggestion
_Returns:_ The total maximum number of work-groups that can be submitted to an ND-range kernel. The minimum value is 1 if the device is not [api]#info::device_type::custom#.
```

Here too

gmlueck:
> I'm not sure what's wrong about using `range`?

After reading sycl_khr_group_interface again, I think the word "range" is consistent with that KHR.  Therefore, the two queries in this PR should be named:

* `device::max_work_group_range` - The maximum number of work-groups in each dimension of an nd-range kernel.  This is the maximum value that can be returned in each dimension from `khr::work_group::range()`.
* `device::max_work_group_linear_ranged` - The maximum total number of work-groups in an nd-range kernel across all dimensions.  This is one more than the largest value that can be returned from `khr::work_group::linear_id()`.

Pennycook:
One small correction, I think `max_work_group_linear_range` should probably be:

> * `device::max_work_group_range_size` - The maximum total number of work-groups in an nd-range kernel across all dimensions.  This the maximum value that can be returned from `khr::work_group::range().size()`.

The KHR didn't define a "linear range" because the `range` class already has the concept of a "size" exposed via `size()`.  That's still consistent with the idea of a "size" being a 1D quantity that represents the product of all the values in a multi-dimensional quantity.

(**NB**: An `id` doesn't have a size despite being multi-dimensional because it represents a point in an iteration space.  You can't compute a "linear ID" without doing so with respect to a specific iteration space.)

Aympab:
Ok, looks like we are converging! :smile: 
I'll wait for @gmlueck 's reply before updating.

Also, the CI is not passing, if anyone can take a look and tell me what's wrong that'll be great because I'm not sure what's not OK in the formatting.

Pennycook:
> Also, the CI is not passing, if anyone can take a look and tell me what's wrong that'll be great because I'm not sure what's not OK in the formatting.

The issue is with reflow conformance.  "Reflow" is a script that we run on the AsciiDoc files to make sure that lines are the correct length, that each sentence begins on a fresh line, and things like that.

If you run this command locally, you'll be able to see which files are failing:
```
$ ./adoc/scripts/verify_reflow_conformance.sh
```

Then you can reformat each failing file like this:
```
$ ./adoc/scripts/reflow.py -overwrite FILENAME
```

adoc/extensions/sycl_khr_max_work_group_query.adoc
```diff
@@ -0,0 +1,137 @@
+[[sec:khr-max-work-group-query]]
+= sycl_khr_max_work_group_query
+
+This extension allows developers to query iteration bounds in each dimension and
+in total for an ND-range kernel.
+The application must ensure that the number of work-groups in each dimension
+[code]#N# of an ND-range kernel is less than or equal to
+[code]#max_work_group_range<N>#.
```
gmlueck:
```suggestion
The application must ensure that the number of work-groups of an ND-range kernel is within the range of values returned by these queries.
```

I think we should make this sentence a bit more generic.  Maybe something like this?  (Note you will need to insert a line break at 80 columns.)

Aympab:
Ok I'll do that now!
Thanks a lot for the attentive reviews

adoc/chapters/acknowledgements.adoc
```diff
@@ -96,6 +96,7 @@
   * Peter Thoman, University of Innsbruck
   * Biagio Cosenza, University of Salerno
   * Paul Preney, University of Windsor
+  * Aymeric Millan, CEA, Maison de la Simulation
```
gmlueck: The organization names are alphabetized.  Can you move this up in this list after "Caster Communications" and before "Codeplay"?

gmlueck:
Good point.  We have "sycl_khr_work_item_queries" proposed in #682.  We should be consistent with singular vs. plural.  I'd suggest renaming this PR to "sycl_khr_max_work_group_queries".  Does anyone have any objections?

BTW, the title of this PR should be changed also to match the KHR name.  The title still says "sycl_khr_max_work_group_sizes".

illuhad:
In the implementation PR the question has come up whether the product of all `max_work_group_range[i]` across all dimensions must always be >= `max_work_group_range_size` (ignoring overflows for the purpose of this discussion), or whether an implementation can return `numeric_limits<size_t>::max()` whenever `max_work_group_range_size` cannot be directly mapped to a native backend API.

My reading of the wording suggests that the former should be the case, but it might be worth clarifying this since this might impact directly how users need to combine the queries to reliably configure a kernel launch of max size.

gmlueck:
> My reading of the wording suggests that the former should be the case, but it might be worth clarifying this since this might impact directly how users need to combine the queries to reliably configure a kernel launch of max size.

I think either is OK.  The spec just says that applications must respect both queries when submitting an nd-range kernel.  Therefore, it should be OK for an implementation to simply return `numeric_limits<size_t>::max()` if there's no hardware imposed limit.

PR #716, opened 2025-02-12
Fix definition of marray::size()

### LINKED ISSUE #525: Change the definition of marray::size() to return NumElements
The spec [defines](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.members.marray) `marray::size()` to return the number of _bytes_ in the marray instead of the number of elements. This is at odds with every other definition of a member named `size()` in SYCL, where `size()` would return the number of elements stored / accessed, and where applicable, a corresponding `byte_size()` function is provided to obtain the number of bytes.

I propose to change the definition to return the number of elements instead to avoid confusion and give the user a method of querying the element count without resorting to metaprogramming:

- Change `marray::size()` to read "Returns `NumElements`."
- (optionally) Add `marray::byte_size()` that reads "Returns the size of this SYCL `marray` in bytes.".

(Accidental) prior art:

- DPC++ [already implements this](https://github.com/intel/llvm/blob/5a76f0ff9c8c259a54e6c74fae46ff18e6d6dd62/sycl/include/sycl/marray.hpp#L140C34-L140C34) in the manner proposed
- The CTS [assumes any SYCL type](https://github.com/KhronosGroup/SYCL-CTS/blob/9691bc093d4d5b8965be68bfb0716e20fda6205b/tests/common/value_operations.h#L234) with `size()` and `operator[]` members to behave in the proposed way, and will fail a SYCL implementation that adheres to the spec for this member.

Pennycook:
This should return the number of elements in the marray, not its size in bytes.

Closes #525.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19525,7 +19525,7 @@ a@
 ----
 static constexpr std::size_t size() noexcept
 ----
-   a@ Returns the size of this SYCL [code]#marray# in bytes.
+   a@ Returns the number of elements in this SYCL [code]#marray# (i.e., [code]#NumElements#).
 
 a@
 [source]


TApplencourt:
\o/

TApplencourt (approved):

gmlueck (approved):
Yikes!  Good catch.

I verified that DPC++ already implements this as "number of elements" and not "size in bytes".  Also, it looks like the CTS is testing it that way also.  Therefore, I think this is what we meant all along.

tomdeakin:
WG approved to merge

PR #718, opened 2025-02-12
Fix incorrect reference to vec::as

### LINKED ISSUE #393: Using rounding_mode in "as" or "convert"?
In section 4.14.2.5. Rounding modes, the spec said "The various rounding modes that can be used in the as member function template are described in Table 146". But the `as` member function of `class vec` cannot accept `rounding_mode` as template arg.
Should this sentence be changed to "The various rounding modes that can be used in the convert member function template are described in Table 146" ?
```
67 template <typename ConvertT,
68 rounding_mode RoundingMode = rounding_mode::automatic>
69 vec<ConvertT, NumElements> convert() const;
70
71 template <typename AsT> AsT as() const;
```

Pennycook:
This paragraph is talking about rounding modes, which apply to `vec::convert`.

Closes #393.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -19320,8 +19320,8 @@ _Returns:_ A [code]#vec# object that represents the result of the operation.
 
 ==== Rounding modes
 
-The various rounding modes that can be used in the [code]#as# member function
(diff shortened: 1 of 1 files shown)


gmlueck (approved):

gmlueck:
Merging as editorial

PR #719, opened 2025-02-12
Clarify that RandomFiller uses one random number

### LINKED ISSUE #198: Example from 4.12.1. Defining kernels as named function objects is misleading
The example
```c++
 class RandomFiller {
 public:
  RandomFiller(accessor<int> ptr)
      : ptr_ { ptr } {
    std::random_device hwRand;
    std::uniform_int_distribution<> r { 1, 100 };
    randomNum_ = r(hwRand);
  }
  void operator()(item<1> item) const { ptr_[item.get_id()] = get_random(); }
  int get_random() { return randomNum_; }

 private:
  accessor<int> ptr_;
  int randomNum_;
};

void workFunction(buffer<int, 1>& b, queue& q, const range<1> r) {
    myQueue.submit([&](handler& cgh) {
      accessor ptr { buf, cgh };
      RandomFiller filler { ptr };

      cgh.parallel_for(r, filler);
    });
}
```
is kind of misleading because of the useless call to `get_random()` instead of returning directly `randomNum_`, which might suggest with the call operator that a different random number is used for each work-item.

Pennycook:
This commit adjusts the `RandomFiller` example and its introduction in text to clarify that all elements of the buffer are initialized to the same random number, as opposed to generating a random number on each work-item.

Closes #198.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17177,7 +17177,7 @@ If the [code]#operator()# function writes to any of the member variables, the
 behavior is undefined.
 
 The following example defines a <<sycl-kernel-function>>, _RandomFiller_, which
-initializes a buffer with a random number.
+initializes all elements of a buffer with the same random number.
 The random number is generated during the construction of the function object
 while processing the command group.
 The [code]#operator()# member function of the function object receives an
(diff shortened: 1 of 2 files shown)


tomdeakin (approved):
Good catch - editorial fix.

gmlueck (approved):

TApplencourt:
Or should we call `r(hwRand)` inside the `()` ?

gmlueck:
Merging as editorial

Issue #340, opened 2023-01-12
Reductions and scans: types allowed


Melirius:
In the spec [4.17.4.5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_reduce) and [4.17.4.6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_exclusive_scan_and_inclusive_scan) all the functions are defined for "fundamental types" that are defined in [5.5, Table 183](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:scalartypes) and only mentioned in the spec [once outside of these sections](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reduction). They do not include `sycl::half`, however, that is an oversight. Two solutions are

1.  Add it to the table of SYCL fundamental types like `double` (that is also optional on device side). Maybe it is not optimal, taking into account what traits should be guaranteed for C++ fundamental type. Then should SYCL fundamental type imply all the properties of C++ fundamental type?
2. Add it to every mention of "fundamental types" in reduce/scan functions (or introduce something like "reducible types" = "fundamental types" + `sycl::half`, and use this term).

nliber:
I'm not against (2) per se.  However, I don't like the direction that every feature has a slightly different subset of types it can be used on.

For instance:  we can't have a `sycl::atomic_ref<half>`, so unless that is just an oversight, that is yet another slightly different subset.

Pennycook:
> For instance: we can't have a sycl::atomic_ref<half>, so unless that is just an oversight, that is yet another slightly different subset.

The reason we can't have `sycl::atomic_ref<half>` is that atomic capabilities are currently tied to hardware support for 32- and 64-bit atomics.  All SYCL devices support 32-bit atomics, but only some support 64-bit atomics (see `aspect::atomic64`).

If there is a desire to add `sycl::half` to a list of fundamental types and use that everywhere, I think we'd just need a new way to describe the limits of `sycl::atomic_ref<>`.  For example, we could say that `sycl::atomic_ref<>` can always be constructed for a `sycl::half`, but that it may not work on some devices (similar to how we handle support for `double` and `long`).

fraggamuffin:
Test already has sycl half
does not affect test progress
is it optional or fundamental? definitely optional, but not fundamental according to Table 183
also scalar types should use sycl::byte
[...]

PR #723, opened 2025-02-13
Clarify group algorithms don't support iterators

### LINKED ISSUE #409: Group algorithms cannot take iterators
The first paragraph of [4.17.4. Group algorithms library](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:algorithms) says that iterators can be passed to group algorithms:

> The first argument to each function is a [group](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#group), and data ranges can be described using **pointers, iterators or instances of the multi_ptr class**

However, all the `joint_*` functions have a restriction that the input is a pointer. Example: [4.17.4.1. any_of, all_of and none_of](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_any_of_all_of_and_none_of)

```c++
template <typename Group, typename Ptr, typename Predicate>
bool joint_any_of(Group g, Ptr first, Ptr last, Predicate pred); // (1)
```

> Constraints: Available only if sycl::is_group_v<std::decay_t<Group>> is true and Ptr is a pointer.

So are the group algorithms meant to also take in iterators? (I would presume this should be restricted to random access iterators, to be more precise.) Since these algorithms are like their `std` counterparts, I think they probably should.  For example, for an accessor `acc`, it would make sense to do something like `joint_any_of(group, acc.begin(), acc.end())` , but as the spec is, this is not necessarily allowed.

Pennycook:
The working group decided that SYCL 2020 should be limited to pointers. Extending these interfaces to accept iterators is under consideration for SYCL-Next.

Closes #409.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -21703,7 +21703,7 @@ reported by [code]#Group::fence_scope#).
 SYCL provides an algorithms library based on the functions described in Section
 28 of the {cpp17} specification.
 The first argument to each function is a <<group>>, and data ranges can be
-described using pointers, iterators or instances of the [code]#multi_ptr# class.
+described using pointers or instances of the [code]#multi_ptr# class.
 The functions defined in this section are free functions available in the
 [code]#sycl# namespace.
 
@@ -21724,8 +21724,9 @@ Unless the description of a function says otherwise, how the elements of a range
 are processed by the work-items in a group is undefined.
 
 SYCL provides separate functions for algorithms which use the work-items in a
-group to execute an operation over a range of iterators and algorithms which are
-applied to data held directly by the work-items in a group.
+group to execute an operation over a range (specified by a start pointer and an
+end pointer) and algorithms which are applied to data held directly by the
+work-items in a group.
 An example of the usage of these functions is given below:
 
 [source,,linenums]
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -21703,7 +21703,7 @@ reported by [code]#Group::fence_scope#).
 SYCL provides an algorithms library based on the functions described in Section
 28 of the {cpp17} specification.
 The first argument to each function is a <<group>>, and data ranges can be
-described using pointers, iterators or instances of the [code]#multi_ptr# class.
+described using pointers or instances of the [code]#multi_ptr# class.
```
gmlueck, line 21706: Maybe this is a separate PR, but the constraints on the functions below don't make it clear that `multi_ptr` is allowed.  They just say "`Ptr` is a pointer".

Pennycook: I'd prefer to do it in a separate PR, if you don't mind.  It probably makes sense to fix the `Ptr` constraints while addressing #340, because we're already going to have to clarify what they can point _to_.

Pennycook: Addressed as part of #730.

gmlueck (approved):

tomdeakin (approved):

tomdeakin:
WG approved to merge for SYCL 2020.

PR #724, opened 2025-02-13
Clarify `vec::as` behavior

### LINKED ISSUE #455: Should be sycl::vec::as<bool> clarified/prohibited?
After recent SYCL-CTS [change](https://github.com/KhronosGroup/SYCL-CTS/commit/affbf36dd917dd6e7be4dc0bd7b7e5228bf130b6), we've noticed that [here](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/common/common_vec.h#L460) we're calling `sycl::vec::as<bool>` for different `sycl::vec<T>`, where T is char, int, float and so on, like:
```
auto inputVec = sycl::vec<char, 8>(0, 1, 2, 3, 4, 5, 6, 7);
auto asVec = inputVec.as<sycl::vec<bool, 8>>();
```
Looks like it's not legal. SYCL2020 defines `template <typename asT> asT as() const` method as: 
> _**Bitwise reinterprets**_ this SYCL vec as a SYCL vec of a different element type and number of elements specified by asT. The new SYCL vec type must have the same storage size in bytes as this SYCL vec.

According to https://en.cppreference.com/w/cpp/numeric/bit_cast:
`If there is no value of type To corresponding to the value representation produced, the behavior is undefined.`

So, it looks like `as<bool>` should be prohibited as it's UB.

Pennycook:
The original description did not cover various corner-cases that could lead to undefined behavior (e.g., reinterpreting using `vec::as<bool>`).

Since C++20 defines `std::bit_cast`, and SYCL 2020 pre-adopts `sycl::bit_cast`, defining `vec::as` equivalent to `bit_cast<AsT>` has two advantages:

- It simplifies the specification of `vec::as` significantly; and
- It enables `vec::as` to inherit undefined behavior, etc from `bit_cast`.

Closes #455.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -18019,14 +18019,9 @@ vec<ConvertT, NumElements> convert() const
 a@
 [source]
 ----
-template <typename asT> asT as() const
+template <typename AsT> AsT as() const
 ----
-   a@ Bitwise reinterprets this SYCL [code]#vec# as a SYCL [code]#vec# of a
-      different element type and number of elements specified by [code]#asT#.
-      The new SYCL [code]#vec# type must have the same storage size in bytes as
-      this SYCL [code]#vec#, and the size of the elements in the new SYCL
-      [code]#vec# ([code]#NumElements * sizeof(DataT)#) must be the same as the
-      size of the elements in this SYCL [code]#vec#.
+   a@ Equivalent to [code]#bit_cast<AsT>(*this)#.
 
 a@
 [source]


TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -18019,14 +18019,9 @@ vec<ConvertT, NumElements> convert() const
...
-template <typename asT> asT as() const
+template <typename AsT> AsT as() const
 ----
-   a@ Bitwise reinterprets this SYCL [code]#vec# as a SYCL [code]#vec# of a
-      different element type and number of elements specified by [code]#asT#.
-      The new SYCL [code]#vec# type must have the same storage size in bytes as
-      this SYCL [code]#vec#, and the size of the elements in the new SYCL
-      [code]#vec# ([code]#NumElements * sizeof(DataT)#) must be the same as the
-      size of the elements in this SYCL [code]#vec#.
+   a@ Equivalent to [code]#bit_cast<AsT>(*this)#.
```
gmlueck, line 18024: Do we need some additional words to say that bit-cast from a 3-element `vec` to a 4-element `vec` produces undefined results?  See #447.

Pennycook: I think this might be covered by the definition of `bit_cast`, but we could still add some clarification.

The [C++20 definition](https://timsong-cpp.github.io/cppwp/n4861/bit.cast) says that "Each bit of the value representation of the result is equal to the corresponding bit in the object representation of from".  The [definition in the latest draft](https://eel.is/c++draft/bit.cast) is much more verbose, and includes statements like "A bit in the value representation of the result is indeterminate if it does not correspond to a bit in the value representation of from or corresponds to a bit for which the smallest enclosing object is not within its lifetime or has an indeterminate value."

By either definition, I think the intended behavior is clear for `vec::as`.  Since a 3-element `vec` is specified to have the same alignment and size as a 4-element `vec`, the bits in the fourth element of the result must come from padding bits that are not part of the value representation of the 3-element `vec`.  By the C++20 definition, the bits in the fourth element of the result will be padding bits of unknown value. By the C++23 definition, the bits in the fourth element of the result are explicitly "indeterminate" because they didn't come from the value representation.

Maybe we could just say something like:
> _Note_: Since the object representation of a `vec<T, 3>` contains padding bits (see [Section 4.14.2.6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#memory-layout-and-alignment)), using `as` or `bit_cast` to create a `vec` with a different number of elements can lead to undefined behavior.

...or we could even put the note in Section 4.14.2.6, since it applies to use of `bit_cast` as well.

Note that I said _can_ lead to undefined behavior, because for certain operations some of the bits are guaranteed to be well-defined.  For example, when going from `vec<T, 3>` to `vec<T, 4>`, the first three elements should be guaranteed to have well-defined values (and this is the behavior of DPC++, at least).

I realize now that this is actually weaker than the wording that was there before, but I think it's a more accurate reflection of how implementations behave.  And it would be odd for `as` to have a restriction that a developer could work around trivially by using `bit_cast` instead.

keryell, line 18024: Which `bit_cast` are you talking about?
`std::bit_cast`? If so it should be precised.

gmlueck: I agree.  I think the non-normative note you propose is sufficient.

Pennycook: I've added this note in f6e719d5805ac893f0bbaa5ddd4f1b40f4b3c7e2.

Pennycook: Good point.  Fixed in 44e1fea7a308e38dc15b274c9a940617fa750c9a.

adoc/chapters/programming_interface.adoc
```diff
@@ -18019,14 +18019,14 @@ vec<ConvertT, NumElements> convert() const
...
 ----
-   a@ Bitwise reinterprets this SYCL [code]#vec# as a SYCL [code]#vec# of a
-      different element type and number of elements specified by [code]#asT#.
-      The new SYCL [code]#vec# type must have the same storage size in bytes as
-      this SYCL [code]#vec#, and the size of the elements in the new SYCL
-      [code]#vec# ([code]#NumElements * sizeof(DataT)#) must be the same as the
-      size of the elements in this SYCL [code]#vec#.
+   a@ Equivalent to [code]#sycl::bit_cast<AsT>(*this)#.
+
+NOTE: Since the object representation of a [code]#vec<T, 3># contains padding
```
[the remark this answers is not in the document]
Pennycook: There's one "implementation note" that uses the `NOTE:` syntax, but it looks like it's the only one and it should probably be fixed.

I've switched to the `[NOTE]` syntax in e91503cdfbb65fd34feb45254a3f53b16d3d6a53.

gmlueck (approved):

PR #587, opened 2024-07-20
Update context description to new format

gmlueck:
Update the section describing the `context` class to the new format.

PR #725, opened 2025-02-14
Improve description of information descriptors

### LINKED ISSUE #410: Explain the rationale for this meta-programming usage.
Yes, we could explain the rationale for this meta-programming usage.

_Originally posted by @keryell in https://github.com/KhronosGroup/SYCL-Docs/pull/407#pullrequestreview-1413323304_

Pennycook:
Previously, the description of information descriptors was spread out across multiple sections and appendices. This commit introduces a new section and a glossary entry to more clearly explain the concept of an information descriptor prior to detailing the behavior of each SYCL class.

Closes #410.  Addresses feedback originally raised by @nmnobre in #407. 

---

I deliberately added this to the end of the "Common interface" section so that it wouldn't affect renumbering, should we choose to cherry-pick it to SYCL 2020.

### DIFF
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -287,9 +287,9 @@ For the full description please refer to <<subsec:buffers>>.
 [[information-descriptor]]information descriptor::
     A class encapsulating an information query and its return type.
     For example, an information descriptor called
-    [code]#sycl::info::class::name# would describe a query for the name of the
-    entity represented by that class, and would define a return type of
-    [code]#std::string#.
+    [code]#sycl::info::class::name# with a return type of [code]#std::string#
+    would describe a query for the name of the entity represented by that class,
+    where the name is an arbitrary string.
     Any information descriptors supported by a SYCL class are listed alongside
     the definition of that class, or in a <<backend>> specification.
 
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -290,8 +290,8 @@ For the full description please refer to <<subsec:buffers>>.
     [code]#sycl::info::class::name# would describe a query for the name of the
     entity represented by that class, and would define a return type of
     [code]#std::string#.
-    A full list of information descriptors for various SYCL classes can be found
-    in <<sec:information-descriptors>>.
+    Any information descriptors supported by a SYCL class are listed alongside
+    the definition of that class.
 
 [[input]]input::
     A state which a <<kernel-bundle>> can be in, representing
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -289,7 +289,7 @@ For the full description please refer to <<subsec:buffers>>.
     For example, an information descriptor called
     [code]#sycl::info::class::name# with a return type of [code]#std::string#
     would describe a query for the name of the entity represented by that class,
-    where the name is an arbitrary string.
+    where the name can be an arbitrary string.
     Any information descriptors supported by a SYCL class are listed alongside
     the definition of that class, or in a <<backend>> specification.
 
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -284,6 +284,15 @@ For the full description please refer to <<subsec:buffers>>.
   * [code]#sycl::group# : The <<group>> class that contains the
     <<work-group-id>> and the member functions on a <<work-group>>.
 
+[[information-descriptor]]information descriptor::
+    A class encapsulating an information query and its return type.
+    For example, an information descriptor called
+    [code]#sycl::info::class::name# would describe a query for the name of the
+    entity represented by that class, and would define a return type of
+    [code]#std::string#.
+    A full list of information descriptors for various SYCL classes can be found
+    in <<sec:information-descriptors>>.
+
 [[input]]input::
     A state which a <<kernel-bundle>> can be in, representing
     <<sycl-kernel-function,SYCL kernel functions>> as a source or intermediate
--- a/adoc/chapters/glossary.adoc
+++ b/adoc/chapters/glossary.adoc
@@ -291,7 +291,7 @@ For the full description please refer to <<subsec:buffers>>.
     entity represented by that class, and would define a return type of
     [code]#std::string#.
     Any information descriptors supported by a SYCL class are listed alongside
-    the definition of that class.
+    the definition of that class, or in a <<backend>> specification.
 
 [[input]]input::
     A state which a <<kernel-bundle>> can be in, representing
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -716,6 +716,40 @@ Construct a SYCL [code]#property_list# with zero or more properties.
 
 |====
 
+=== Information queries
+
+Each of the following classes provides a generic mechanism for querying the
+class for information: [code]#context#, [code]#device#, [code]#event#,
+[code]#kernel#, [code]#platform#, and [code]#queue#.
+
+Each available query is described by an <<information-descriptor>>, which is is
+a class or class template implementing the interface below:
+
+[source]
+----
+class __InformationDescriptor__
+{
+ public:
+  using return_type = /* return type for this information query */;
+};
+----
+
+==== Information query interface
+
+The information query interface consists of a [code]#get_info()# function
+template, which is templated on an <<information-descriptor>>.
+
+[source,role=synopsis]
+----
+template <typename Param>
+typename Param::return_type get_info() const;
+----
+
+_Constraints_: [code]#Param# must be an <<information-descriptor>> compatible
+with the class, as listed in <<sec:information-descriptors>>.
+
+_Returns_: The information corresponding to [code]#Param# in the table of
+information descriptors associated with the class.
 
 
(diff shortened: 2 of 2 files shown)


TApplencourt:
>Previously, the description of information descriptors was spread out across multiple sections and appendices. 

This is a stupid remark, but should we remove the other text that has now been consolidated in the new section?

Pennycook:
> This is a stupid remark, but should we remove the other text that has now been consolidated in the new section?

I did consider this, but I'd rather leave this to a later PR.  I wasn't really sure how far to go with it.  I like the way that @gmlueck reformatted the description of the various aspects, and I think it probably makes sense to do something similar with information descriptors.  But removing the information descriptor appendix and cleaning up those tables seemed like a huge change.

TApplencourt:
Work for me! 
We are Lanister and will pay our technical debt!  (Man, this is an old reference...)

adoc/chapters/programming_interface.adoc
```diff
@@ -716,6 +716,40 @@ Construct a SYCL [code]#property_list# with zero or more properties.
...
+[code]#kernel#, [code]#platform#, and [code]#queue#.
+
+Each available query is described by an <<information-descriptor>>, which is is
+a class or class template implementing the interface below:
+
+[source]
+----
+class __InformationDescriptor__
+{
+ public:
```
TApplencourt, line 732: I think the other part of the spec indent has `2` space, but that ok. Need to make a PR to format all the [source] with clang format

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -716,6 +716,40 @@ Construct a SYCL [code]#property_list# with zero or more properties.
 
 |====
 
+=== Information queries
+
+Each of the following classes provides a generic mechanism for querying the
+class for information: [code]#context#, [code]#device#, [code]#event#,
+[code]#kernel#, [code]#platform#, and [code]#queue#.
+
+Each available query is described by an <<information-descriptor>>, which is is
```
nmnobre, line 725:
```suggestion
Each available query is described by an <<information-descriptor>>, which is
```

nmnobre (changes requested):
Thanks a lot for doing this @Pennycook.
I think it's a very good readability improvement. :)

adoc/chapters/programming_interface.adoc
Pennycook: Good catch, thanks!

gmlueck:
I'm not wild about this change for a few reasons:

* The new section "Information queries" is not really normative.  Instead, it describes a general pattern that some of the other classes follow.  However, those other classes have a normative description of their information descriptors.  Therefore, the new section seems redundant to me.

* The new section "Information queries" has a list of classes that follow this pattern.  We will need to remember to update this list if we ever add information descriptors to other classes.

* I especially don't like the new references to "Appendix A" because my goal is to remove that appendix.

I agree with @nmnobre that the old format we had in the SYCL specification made the structure of information descriptors non-obvious, just as he says in #407.  However, I think the new format fixes this because it shows the exact definition of each information descriptor.  We use this format for most of the classes that use information descriptors (`context`, `device`, `platform`, `queue`).  This new format did not exist when #407 was filed.

The only classes with information descriptors that don't follow the new format are `event` and `kernel`.  I think that the right fix for #407 should be to update these sections to the new format, and then we can remove appendix A.

nmnobre:
> However, I think the new format fixes this because it shows the exact definition of each information descriptor. [...]
> The only classes with information descriptors that don't follow the new format are `event` and `kernel`. I think that the right fix for #407 should be to update these sections to the new format, and then we can remove appendix A.

Ah! This is awesome, I wasn't aware of this reformat! For clarity, you're talking about #567, #587, #601, #606 (and their cherry-picks to `sycl-2020`), correct? I agree the new format is sufficient and, since it's already in `sycl-2020`, am I correct in assuming the content of this PR will never be published before or without that new format? If that's the case, it might be redundant, yeah. And I now understand, so is #729?

Pennycook:
> * The new section "Information queries" is not really normative.  Instead, it describes a general pattern that some of the other classes follow.  However, those other classes have a normative description of their information descriptors.  Therefore, the new section seems redundant to me.
> * The new section "Information queries" has a list of classes that follow this pattern.  We will need to remember to update this list if we ever add information descriptors to other classes.

I was following the precedent of set by [Section 4.5.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics), [Section 4.5.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:byval-semantics) and [Section 4.5.4](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_properties), all of which describe a general pattern that other classes follow and provide a list of classes. Are you suggesting that we get rid of these sections too?

I think there is value in providing a high-level overview of why commonality may exist in an interface like this.  I agree with you that showing the actual interface and listing the classes is probably unnecessary, but I'd like us to be consistent.

To clarify: I'm not opposed to taking the interface and class list out of this PR, leaving only the introduction to the concept of an information descriptor and the glossary entry.  But if we do that, I think it would make sense to define all of the operators available in SYCL runtime classes explicitly, instead of just saying something like "This class has common reference semantics."

> * I especially don't like the new references to "Appendix A" because my goal is to remove that appendix.

I agree we should remove the appendix.  I can remove the references to the appendix, to make that easier.

> The only classes with information descriptors that don't follow the new format are `event` and `kernel`. I think that the right fix for #407 should be to update these sections to the new format, and then we can remove appendix A.

I agree we should do this.

Pennycook:
> I'm not sure what the right thing to do here is yet. I think it might be more clear to explicitly list all the common reference semantics member functions, and then have a description that says something like:
> 
> > See section XX for a description of these functions.
> 
> where section XX defines the common reference semantics.

I agree that this would be clearer.  The first time I read the SYCL specification, I thought the comment that says "Common member functions" was a header for the functions beneath it, rather than a substitute for some function definitions.

> I don't think we need to do this for `get_info`. It's easy to describe what the function does along with all the other member functions. I don't think it would make the spec any clearer to have a common section describing `get_info`. In fact, I think it makes the spec less clear because we want the _Constraints_, etc. to be tailored for each class.
> 
> If you remove the interface and the class list, what would remain in the section "Information queries"? I'd like to avoid listing all the classes that define `get_info` for the reason I mention above. It seems like this leaves very little in the section.

I've pushed a change in c31cfcae71b248b52f79980ba13cd4db39470918 to show what this could look like.

As you say, it does end up being quite a small section.  But I still think there's some value in highlighting somewhere that there's a single `get_info()` function design that is shared by a bunch of classes.  If a first time reader starts with the early sections before diving into the class definitions, it will teach them to expect that several classes expose this function.  Without a section like this, a reader won't realize `get_info()` is a common interface until they've read at least two class definitions.

adoc/chapters/programming_interface.adoc
```diff
@@ -716,6 +716,22 @@ Construct a SYCL [code]#property_list# with zero or more properties.
...
+information.
+
+Each available query is described by an <<information-descriptor>>, which is a
+class or class template that encapsulates an information query and its return
+type.
+
+==== Information query interface
+
+The information query interface consists of a [code]#get_info()# function
+template, which is templated on an <<information-descriptor>>.
```
gmlueck, line 731: This isn't entirely true because there is also `get_backend_info`, which also takes an information descriptor.

adoc/chapters/glossary.adoc
```diff
@@ -284,6 +284,15 @@ For the full description please refer to <<subsec:buffers>>.
   * [code]#sycl::group# : The <<group>> class that contains the
     <<work-group-id>> and the member functions on a <<work-group>>.
 
+[[information-descriptor]]information descriptor::
+    A class encapsulating an information query and its return type.
+    For example, an information descriptor called
+    [code]#sycl::info::class::name# would describe a query for the name of the
+    entity represented by that class, and would define a return type of
+    [code]#std::string#.
```
gmlueck, line 292: Why is the return type `std::string`?  Just because a "name" would normally be a string?

gmlueck (commented):
A couple comments on this one.  Feel free to merge this when you think it's ready.

This is probably my last review for a while as I have finally finished my travels and have arrived in New Zealand!

adoc/chapters/programming_interface.adoc
Pennycook: Good catch!  I always forget about this backend-specific stuff.  I've added references to backends in 2aa55a09faf23e1319573ec4a3b14cdc112d54bd.

adoc/chapters/glossary.adoc
Pennycook: Right, exactly.

I've tried to reword this glossary entry in b70832fed89922457c751e78a1e5c2c62b1dd619 to avoid this issue: instead of saying that it "would" return a `std::string` just because it's called "name", I've flipped it to say that if a descriptor was called "name" and had a return type defined to be `std::string`, the information query should be expected to return the entity's name as a string.

tomdeakin:
More reviews needed.

keryell (approved):
Thanks!

gmlueck (approved):

gmlueck:
WG approved

PR #726, opened 2025-02-14
Add integral overloads for `id` and `range` operators

### LINKED ISSUE #539: id<1> operators and integer literals
The current version of the SYCL 2020 specification allows conversion of `size_t` to and from `id<1>` and all operators are defined with a variant taking a `size_t` RHS operand. However, the way this is defined, the following code would be ambiguous:

```c++
sycl::id<1> x{1};
x = x + 1; // Ambiguous: Should it use `int + int` or `id<1> + id<1>`?
```

Possible solution: Change the operators with `size_t` arguments to take `T` instead, where `std::is_integral_v<T>` is `true. For example, see https://godbolt.org/z/TcM8hrWzE.

Inspired by https://github.com/KhronosGroup/SYCL-CTS/pull/870.

### LINKED ISSUE #711: Intended behavior of operator* in `id<1>`
What is the intended behavior of the following code?

```c++
#include <sycl/sycl.hpp>

int foo(sycl::id<1> i, int j)
{
    return i * j;
}
```

It compiles with DPC++, but fails with SimSYCL (https://github.com/celerity/SimSYCL/issues/40) due to an ambiguous overload.

I wrote a toy implementation of `id` based on what I think the specification says (https://godbolt.org/z/P8MTWW1s7) and the operator overload works as I'd expect.

Pennycook:
The specification previously only listed an overload for `size_t`. However, all existing SYCL implementations provide overloads for
all integral types to avoid ambiguity in operator overloads.

This change aligns the specification with the behavior of existing SYCL implementations.

Closes #539, closes #711.

---

Note that I also re-ordered the synopses for `id` and `range` based on an observation made during discussion in #711.  It might be easier to review the changes from the two commits separately: the first commit adjusts the overloads, and the second commit reorders things.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12076,6 +12076,7 @@ a@
 [source]
 ----
 // Available only when std::is_integral_v<T> is true
+template <typename T>
 range operatorOP(const range& lhs, const T& rhs)
 ----
    a@ Where [code]#OP# is: [code]#pass:[+]#, [code]#-#, [code]#*#,
@@ -12097,6 +12098,7 @@ a@
 [source]
 ----
 // Available only when std::is_integral_v<T> is true
+template <typename T>
 range operatorOP(const T& lhs, const range& rhs)
 ----
    a@ Where [code]#OP# is: [code]#pass:[+]#, [code]#-#, [code]#*#,
@@ -12131,6 +12133,7 @@ a@
 [source]
 ----
 // Available only when std::is_integral_v<T> is true
+template <typename T>
 range& operatorOP(range& lhs, const T& rhs)
 ----
    a@ Where [code]#OP# is: [code]#pass:[+=]#, [code]#-=#,[code]#*=#, [code]#/=#, [code]#%=#, [code]#+<<=+#, [code]#>>=#, [code]#&=#, [code]#|=#, [code]#^=#.
@@ -12420,6 +12423,7 @@ a@
 [source]
 ----
 // Available only when std::is_integral_v<T> is true
+template <typename T>
 id operatorOP(const id& lhs, const T& rhs)
 ----
    a@ Where [code]#OP# is: [code]#pass:[+]#, [code]#-#, [code]#*#, [code]#/#, [code]#%#, [code]#<<#, [code]#>>#,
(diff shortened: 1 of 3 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -12075,7 +12075,8 @@ is the cast to [code]#size_t#.
 a@
 [source]
 ----
-range operatorOP(const range& lhs, const size_t& rhs)
+// Available only when std::is_integral_v<T> is true
```
TApplencourt, line 12078: Should this be a constrain?

TApplencourt:
Need to wait for CTS update I assume

adoc/chapters/programming_interface.adoc
Pennycook: "Available only when" is how we used to specify constraints.  We _should_ add a "Constraints" paragraph to the description of these operators, but I don't think it makes sense to do that while they're in the old format (i.e., in a table).

TApplencourt (approved):

keryell (approved):
Thanks!

adoc/chapters/programming_interface.adoc
```diff
@@ -12075,7 +12075,8 @@ is the cast to [code]#size_t#.
...
+range operatorOP(const range& lhs, const T& rhs)
```
gmlueck, line 12079: Here and elsewhere, I think you are missing the `template<typename T>`.

Pennycook: Good catch!  Fixed in 73c3bc509c662f09931725c852a846530e7bd429.

adoc/headers/id.h
```diff
@@ -33,17 +33,24 @@ template <int Dimensions = 1> class id {
   // OP is: +, -, *, /, %, <<, >>, &, |, ^, &&, ||, <, >, <=, >=
   friend id operatorOP(const id& lhs, const id& rhs) { /* ... */
   }
-  friend id operatorOP(const id& lhs, const size_t& rhs) { /* ... */
+
+  // OP is: +, -, *, /, %, <<, >>, &, |, ^, &&, ||, <, >, <=, >=
+  // Available only when std::is_integral_v<T> is true
+  friend id operatorOP(const id& lhs, const T& rhs) { /* ... */
```
gmlueck, line 39: Missing the `template<typename T>` in the headers also.

Pennycook: Oops.  Fixed in e986d1a2e80cb080f0cd896c5f1cd90eebaade6b.

gmlueck (approved):

gmlueck:
Merge conflicts resolved.  Merging as the WG gave approval at 9/11/2025 meeting.

PR #727, opened 2025-02-14
Clarify the meaning of different device types

### LINKED ISSUE #7: Device definitions
https://github.com/KhronosGroup/SYCL-Docs/blob/4a2dff3a779ae6fa4a477b0f02002c6ab82a3088/latex/programming_interface.tex#L437

Currently the device selectors define returning a device of a specific type but there is no definition of what those types mean within this document. To find the definition of what a GPU device type one must look in the OpenCL specification and even there the definition of what a device is quite sparse. It would be nice to have some definition of what a particular device selector will give you.

### LINKED ISSUE #545: A definition of a GPU
Very minor nitpick:

Currently, [the spec says](https://github.com/KhronosGroup/SYCL-Docs/blob/acc98bed254817b217dddbd186d1157bc276bec3/adoc/chapters/programming_interface.adoc?plain=1#L2626-L2633) that a GPU is:

> A device that can also be used to accelerate a 3D graphics API. 

That implies some kind of interoperability with graphics APIs and also that, e.g., [NVIDIA A100 in MIG mode is not a GPU](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/#app-considerations).

For the reference, OpenCL defines [`CL_DEVICE_TYPE_GPU`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_TYPE_GPU) as:

> An OpenCL device similar to a GPU (Graphics Processing Unit). Many systems include a dedicated processor for graphics or rendering that may be considered a GPU OpenCL device.

Which is a bit tautological, but SYCL already defines `aspect::cpu` as "A device that runs on a CPU." (which is an awkward phrasing, IMO, but it gets the point across).

Perhaps it would be better to redefine `aspect::gpu` as "A GPU or a device with GPU-like architecture"?

Pennycook:
The description of what constitutes a CPU or GPU has always been unclear.

This commit adjusts these descriptions to clarify that it is ultimately up to the implementation to decide whether a device is identified as a specific device type. It also attempts to separate normative behaviors (like the connection between device type aspects and `info::device_type`) from non-normative descriptions of what a device type may mean.

Closes #7, closes #545.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3283,7 +3283,7 @@ describes them.
 .[apidef]#aspect::cpu#
 [role=synopsis,id=api:aspect-cpu]
 --
-Indicates that this device identfies itself as a CPU, and has device type
+Indicates that this device identifies itself as a CPU, and has device type
 [api]#info::device_type::cpu#.
 
 {note}A device with this aspect will typically share some or all of the
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3283,8 +3283,8 @@ describes them.
 .[apidef]#aspect::cpu#
 [role=synopsis,id=api:aspect-cpu]
 --
-Indicates that this device identifies itself as a CPU, and has device type
-[api]#info::device_type::cpu#.
+Indicates that the implementation identifies this device as a CPU that has
+device type [api]#info::device_type::cpu#.
 
 {note}A device with this aspect will typically share some or all of the
 execution resources available to the host {cpp} application.
@@ -3296,8 +3296,8 @@ execution resources available to the host {cpp} application.
 .[apidef]#aspect::gpu#
 [role=synopsis,id=api:aspect-gpu]
 --
-Indicates that this device identifies itself as a GPU, and has device type
-[api]#info::device_type::gpu#.
+Indicates that the implementation identifies this device as a GPU that has
+device type [api]#info::device_type::gpu#.
 
 {note}A device with this aspect may have additional capabilities for
 accelerating graphics operations, via SYCL image functionality and/or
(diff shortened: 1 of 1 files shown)


TApplencourt:
Amazing, thanks a lot! I wanted to make those kinds of changes but didn't know how to tackle them. They look nice! 

>Indicates that this device identfies itself as a CPU, and has device type

Typos, Should be `identifies`. 

PS: I will not make any joke about using self-affirming terminology on the SYCL Spec...

Pennycook:
> Typos, Should be `identifies`.

Good catch, thanks!

> PS: I will not make any joke about using self-affirming terminology on the SYCL Spec...

Since you brought it up... I did wonder if this wording could be problematic, because somebody might think we're trying to make a joke.  Maybe:

> Indicates that the implementation identifies this device as a CPU, and has device type [api]#info::device_type::cpu#.

...would be better?  It's arguably more accurate, too, since different implementations may choose to represent the same device in different ways.  I'll make that change.

TApplencourt:
>    Indicates that the implementation identifies this device as a CPU, and has device type [api]#info::device_type::cpu#.

Look good to me!

TApplencourt (approved):

tomdeakin (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -3283,38 +3283,51 @@ describes them.
...
 --
-A dedicated accelerator device, usually using a peripheral interconnect for
-communication.
-Devices with this aspect have device type [api]#info::device_type::accelerator#.
+Indicates that the implementation identifies this device as an accelerator that
+has device type [api]#info::device_type::accelerator#.
+
+{note}A device with this aspect will typically be a dedicated accelerator
+device, with a peripheral interconnect for communication.
+{endnote}
```
gmlueck, line 3318: I think FPGA devices are normally exposed as this type.  Would a reader think that a "dedicated accelerator" might be an FPGA?  This is an honest question.  I don't know what most people would think of when reading the term "dedicated accelerator".

Pennycook: Good question.  I have no idea, and I only used this phrasing because it was already there.

My reading of "accelerator" is something like "a piece of hardware that is not the host CPU", and I think "dedicated" implies that it's not going to be used to run anything except kernels you enqueue to it (i.e., it's not running anything related to the OS).  An FPGA would fit that description.  I guess an implementation might also choose to expose a "GPU" without graphics API support or one that isn't running a display as an "accelerator" instead of a "GPU".

gmlueck (approved):

AerialMantis (approved):

tomdeakin:
WG approved to merge. Fix to 2020

PR #728, opened 2025-02-14
Fix regex in api_xrefs

etomzak:
~~Based on line 72, I *think* the intent was always that `-` would be a valid character for api-xrefs IDs (even though it's not a valid character for C++ identifiers). Prior to this fix, API IDs with `-` characters (e.g., `[api]#my-api#` in asciidoc) were silently not getting the `href` treatment, because they were missed by the regex on line 117. I.e., `[api]#my_api#` would get a link in the HTML, `[api]#my-api#` would not, and no error was generated.~~

~~This took a while to debug.~~

~~Caveat: I know nothing about Ruby.~~

* Make `-` an invalid character in API names
* Improve error messages when invalid characters used in `api-xrefs` variable
* Replace `*` with `+`, since I don't think matching 0-length strings was ever intended

### DIFF
--- a/adoc/config/api_xrefs/extension.rb
+++ b/adoc/config/api_xrefs/extension.rb
@@ -70,16 +70,16 @@ class AddApiXrefs < Extensions::Postprocessor
       api_id_array = []
       if document.attr? 'api-xrefs'
         (document.attr 'api-xrefs').scan(/([^=\s]+)=([^=\s]+)/) do |api, id|
-          if not all_ids.key?(id)
-            logger.error "Id '#{id}' from api-xrefs is not defined"
-            next
-          end
           if not /^[\w:]+$/.match?(api)
             logger.error "API '#{api}' from api-xrefs contains invalid characters"
             next
           end
           if not /^[\w:-]+$/.match?(id)
             logger.error "Id '#{id}' from api-xrefs contains invalid characters"
+            next
+          end
+          if not all_ids.key?(id)
+            logger.error "Id '#{id}' from api-xrefs is not defined"
             next
           end
           api_id_array.push([api, id])
--- a/adoc/config/api_xrefs/extension.rb
+++ b/adoc/config/api_xrefs/extension.rb
@@ -53,10 +53,10 @@ include ::Asciidoctor
 class AddApiXrefs < Extensions::Postprocessor
   include Asciidoctor::Logging
 
-  Id = /<\w+ id="([\w:-]*)"/
-  ApiSpan = /<span class="api">([\w:]*)<\/span>/
-  ApiDefSpan = /<span class="apidef">([\w:]*)<\/span>/
-  ApiIdDiv = /<div id="([\w:-]*)"/
+  Id = /<\w+ id="([\w:-]+)"/
+  ApiSpan = /<span class="api">([\w:]+)<\/span>/
+  ApiDefSpan = /<span class="apidef">([\w:]+)<\/span>/
+  ApiIdDiv = /<div id="([\w:-]+)"/
 
   def process document, output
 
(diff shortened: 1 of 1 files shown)


TApplencourt:
Make sense `\w == [a-zA-Z0-9_]`; honestly, I think `\S*?` (aka everything that is not space) will work best.

TApplencourt (approved):

gmlueck:
My weak preference would be to change line 72 to this:

```
(document.attr 'api-xrefs').scan(/([\w:]*)=([\w:-]*)/) do |api, id|
```

I.e. removing the `-` from the regex, making it consistent with `ApiSpan` and `ApiDefSpan`.  This makes sense to me because the "api name" is always a C++ identifier in our current specification.  I tried doing this, and the SYCL spec still builds, so I think it would work fine for the SYCL repo.

I assume you're using the `[apidef]` stuff in your fork of our repo in some other way that does not use C++ identifiers as the "api name"?  Can you explain your use case?

I'll be out for a few weeks, so others in the groups should decide what to do.  I'm not super opposed to this PR.  However, if we are going to expand the set of legal characters for an "api name", it seems like we should allow anything, not just add `-` to the list.

TApplencourt:
One posiblity will be add a new reference ID

etomzak:
> Can you explain your use case?

Sure. The short version is that in SYCL SC, we're looking at creating some "tags" -- essentially glorified comments that can be placed anywhere in the spec. So the spec can say, "`foo()` conforms to `snafu-1`," and when you click on `snafu-1`, you go to the section explaining what `snafu-1` means. So it's not exactly the same thing as an API definition, but it's similar enough that it makes sense to handle it with the same documentation facility (instead of, say, copy-pasing the api_xrefs script and creating a subtly different version).

The longer version for people not familiar with this corner of spec tooling (based on my best understanding today):

The new spec format uses asciidoc `[api]` and `[apidef]` attributes to create clickable links to specific APIs. For example, `[api]#context::get_backend#` produces a fixed-width formatted clickable link to `[apidef]#context::get_backend#` (where `get_backend` is defined). The problem is that there's no way in asciidoc to create the same style of link to a section. For example, it's not possible to create a clickable link `context` (i.e., `[api]#context#` in asciidoc) that takes you to the context class section (`sec:context-class`). This is a painful limitation, because there are many places where we'd like to talk about a SYCL class and get the class name to link to the relevant section.

api_xrefs is a post-processing script that gets applied to the HTML output (so it's not available in the PDF). api_xrefs creates additional clickable links based on a key-value map. `[api]#context#` can be mapped to the `sec:context-class` anchor, and api_xrefs will create the corresponding `href` in the HTML.

So while api_xrefs is primarily intended for C++ identifiers, I don't see any reason for limiting it to C++ identifiers. As I've discovered, `-` could be useful. Just allowing all characters also seems reasonable, because this would, for example, allow for link text with a wildcard (e.g., `foo_*_bar`). IMO the goal here should really be to limit the surprise factor for people editing the spec: "Why can I use `_` in my link, but not `+`!?"

adoc/config/api_xrefs/extension.rb
```diff
@@ -69,12 +69,20 @@ def process document, output
...
+            next
           end
+          if not /^[\w:]+$/.match?(api)
+            logger.error "API '#{api}' from api-xrefs contains invalid characters"
+            next
+          end
+          if not /^[\w:-]+$/.match?(id)
+            logger.error "Id '#{id}' from api-xrefs contains invalid characters"
+            next
+          end
```
TApplencourt, line 84: 2 Point:
- I think we should check the `id` in keys _after_  verify the validity of the key.
- Maybe better to avoid the `[^=\s]+)=([^=\s]+)]`, use `split(`=`, 2) a little more explicit Imo.


So I propose something like (didn't tried it, so maybe typo / wrong indent -- I use https://github.com/rubocop/rubocop at as linter when I need one ) 
```
Valid_api_regex = /[\w:]+/
Valid_id_regex = /[\w:-]+/
 
api_id_array =  (document.attr('api-xrefs') || [])  # avoid both `api_id_array = [] and first `if` I think)
				.map{ |x| x.split("=",2) } # First split who replace the scan
				.filter { |api,id| # Filter to avoid the `do` + `next` + `push`

	    if not valid_api_regex.match?(api)
	        logger.error "API '#{api}' from api-xrefs contains invalid characters"
	        next
	    end
	          
	    if not valid_id_regex.match?(id)
	        logger.error "Id '#{id}' from api-xrefs contains invalid characters"
	        next
	    end
	
	 	if not all_ids.key?(id) {
	 		logger.error "Id '#{id}' from api-xrefs is not defined"
	 		next
	 	}
}
```

etomzak: > I think we should check the id in keys after verify the validity of the key.

I've reordered the checks.

> So I propose something like

I tried your suggestion, but I quickly ran into an `undefined method 'map' for an instance of String` error. If you can get your suggestion to work, then I'm happy to use it, but I don't have the bandwidth to get deep enough into Ruby to debug this.

gmlueck (approved):

PR #730, opened 2025-02-18
Clarify group algorithm constraints

### LINKED ISSUE #340: Reductions and scans: types allowed

### LINKED ISSUE #461: Should the reduce and scan group algorithms support marray/vec?
From [4.17.4. Group algorithms library](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:algorithms):

> All group functions are supported for the fundamental scalar types supported by SYCL (see [Table 184](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.types.fundamental)) and instances of the SYCL vec and marray classes.

However, not all the specifications of the group functions adhere to this rule. Specifically, the reduce and scan algorithms do not allow for vec and marray, e.g. [4.17.4.5. reduce](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_reduce):

> ```
> template <typename Group, typename Ptr, typename BinaryOperation>
> std::iterator_traits<Ptr>::value_type
> joint_reduce(Group g, Ptr first, Ptr last, BinaryOperation binary_op); // (1)
> ```
> [...]
> Constraints: Available only if sycl::is_group_v<std::decay_t<Group>> is true, **Ptr is a pointer to a fundamental type**, and BinaryOperation is a SYCL function object type.

However, this could be argued to be fine (although in my opinion, confusing) because of the following statement in 4.17.4:

> Any restrictions from the standard algorithms library apply. Some of the functions in the SYCL algorithms library introduce additional restrictions in order to maximize portability across different devices and to minimize the chances of encountering unexpected behavior.

Should the reduce and scan specifications be amended to allow for vec and marray?

Pennycook:
This PR makes three clarifications:

- When an algorithm accepts a pointer, it can be a `multi_ptr`;
- When an algorithm accepts a value, it can be a `vec` or `marray`;
- When an algorithm accepts a value, it can be `sycl::half`.

A list of types compatible with group algorithms is removed in favor of listing the types explicitly for each algorithm, since some algorithms have different constraints (e.g., some algorithms support all trivially copyable types).

Closes #340 and closes #461.

---

Two other things to note here...

I deliberately didn't add `__swizzle__`, because I don't think that was part of our original intent.  We could explore this as an extension, but adding the overloads for `__swizzle__` is non-trivial; the return type would have to be different (i.e., `vec`), and I don't think existing implementations actually support this case.

I realize that the group algorithms section is now _really_ long and repetitive, but my goal with this PR is simply to fix the wording.  The section would definitely benefit from another formatting and consolidation pass.  Defining a term that means "C++ fundamental type, `half`, `marray` or `vec` would also help.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -21712,10 +21712,6 @@ Some of the functions in the SYCL algorithms library introduce additional
 restrictions in order to maximize portability across different devices and to
 minimize the chances of encountering unexpected behavior.
 
-All algorithms are supported for the fundamental scalar types supported by SYCL
-(see <<table.types.fundamental>>) and instances of the SYCL [code]#vec# and
-[code]#marray# classes.
-
 The <<group>> argument to a SYCL algorithm denotes that it should be performed
 collaboratively by the work-items in the specified group.
 All algorithms act as group functions (as defined in <<sec:group-functions>>),
@@ -21752,9 +21748,10 @@ Boolean conditions applied to data held directly by the work-items in a group.
 include::{header_dir}/algorithms/any_of.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if
-    [code]#sycl::is_group_v<std::decay_t<Group>># is true and [code]#Ptr# is a
-    pointer.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_group_v<std::decay_t<Group>># is true; and
+    - [code]#Ptr# is a pointer or [code]#multi_ptr#.
+
 +
 --
 _Preconditions:_ [code]#first# and [code]#last# must be the same for all
@@ -21812,9 +21809,10 @@ _Returns:_ true if [code]#pred# is true for any work-item in group [code]#g#.
 include::{header_dir}/algorithms/all_of.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if
-    [code]#sycl::is_group_v<std::decay_t<Group>># is true and [code]#Ptr# is a
-    pointer.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_group_v<std::decay_t<Group>># is true; and
+    - [code]#Ptr# is a pointer or [code]#multi_ptr#.
+
 +
 --
 _Preconditions:_ [code]#first# and [code]#last# must be the same for all
@@ -21872,9 +21870,10 @@ _Returns:_ true if [code]#pred# is true for all work-items in group [code]#g#.
 include::{header_dir}/algorithms/none_of.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if
-    [code]#sycl::is_group_v<std::decay_t<Group>># is true and [code]#Ptr# is a
-    pointer.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_group_v<std::decay_t<Group>># is true; and
+    - [code]#Ptr# is a pointer or [code]#multi_ptr#.
+
 +
 --
 _Preconditions:_ [code]#first# and [code]#last# must be the same for all
@@ -21943,8 +21942,10 @@ shifting values a fixed number of work-items to the left or right.
 include::{header_dir}/algorithms/shift.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if [code]#std::is_same_v<std::decay_t<Group>,
-    sub_group># is true and [code]#T# is a trivially copyable type.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_same_v<std::decay_t<Group>, sub_group># is true; and
+    - [code]#T# is a trivially copyable type.
+
 +
 --
 _Preconditions:_ [code]#delta# must be the same for all work-items in the group.
@@ -21963,8 +21964,10 @@ _Returns:_ the value of [code]#x# from the work-item whose group local id
 size, but the value returned in this case is unspecified.
 --
 
-  . _Constraints:_ Available only if [code]#std::is_same_v<std::decay_t<Group>,
-    sub_group># is true and [code]#T# is a trivially copyable type.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_same_v<std::decay_t<Group>, sub_group># is true; and
+    - [code]#T# is a trivially copyable type.
+
 +
 --
 _Preconditions:_ [code]#delta# must be the same for all work-items in the group.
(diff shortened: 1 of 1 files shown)


TApplencourt:
Look good, but should CTS but updated?

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -21943,8 +21938,10 @@ shifting values a fixed number of work-items to the left or right.
 include::{header_dir}/algorithms/shift.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if [code]#std::is_same_v<std::decay_t<Group>,
-    sub_group># is true and [code]#T# is a trivially copyable type.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_same_v<std::decay_t<Group>, sub_group># is true; and
+    - [code]#T# is a trivially copyable type.
```
gmlueck, line 21943: Note that the spec does not guarantee that `vec` or `marray` are trivially copyable.  I believe there is an internal issue which suggests that we should change this.  At one point, DPC++ did **not** implement these as trivially copyable, though I think we may have fixed that by now.

adoc/chapters/programming_interface.adoc
```diff
@@ -22074,10 +22077,12 @@ used for forward compatibility with future SYCL versions.
...
 ----
 
-  . _Constraints:_ Available only if
-    [code]#sycl::is_group_v<std::decay_t<Group>># is true, [code]#Ptr# is a
-    pointer to a fundamental type, and [code]#BinaryOperation# is a SYCL
-    function object type.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_group_v<std::decay_t<Group>># is true;
+    - [code]#Ptr# is a pointer or [code]#multi_ptr# to: a fundamental type,
+      [code]#sycl::half#, [code]#marray# or [code]#vec#; and
```
gmlueck, line 22083: Note that `marray` is allowed to have elements that are not fundamental types.  In fact, the elements are even allowed to be types that are not trivially copyable (or even device copyable).  I suspect you do not intend to support `marray` in these cases?

adoc/chapters/programming_interface.adoc
```diff
@@ -22074,10 +22077,12 @@ used for forward compatibility with future SYCL versions.
...
+    - [code]#BinaryOperation# is a SYCL function object type.
+
 +
 --
 _Mandates:_ [code]#binary_op(*first, *first)# must return a value of type
```
gmlueck, line 22088: (This comment is for the line below, but GitHub won't let me make a comment there.)

I'm sure this is opening a can of worms, but we do not say anywhere that `std::iterator_traits` is specialized for `multi_ptr`.  This becomes relevant with this PR because it is clarifying that `Ptr` can be a `multi_ptr`.

adoc/chapters/programming_interface.adoc
```diff
@@ -21752,9 +21744,10 @@ Boolean conditions applied to data held directly by the work-items in a group.
 include::{header_dir}/algorithms/any_of.h[lines=4..-1]
 ----
 
-  . _Constraints:_ Available only if
-    [code]#sycl::is_group_v<std::decay_t<Group>># is true and [code]#Ptr# is a
-    pointer.
+  . _Constraints:_ Available only if all of the following conditions are met:
+    - [code]#sycl::is_group_v<std::decay_t<Group>># is true; and
+    - [code]#Ptr# is a pointer or [code]#multi_ptr#.
+
```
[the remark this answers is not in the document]
Pennycook: Thanks for checking.  This is actually one of the reasons why I removed the general statements!

The type requirements that apply to the other algorithms don't apply to `any_of`, `all_of` and `none_of`.  The reason is that these algorithms all accept a user-defined `Predicate` function that converts the input into a `bool`, and it's the resulting `bool` values that need to be shared across all of the work-items in the group.

adoc/chapters/programming_interface.adoc
Pennycook: You're right, internal issue 564 talks about this.  I'll open a pull request to make these types trivially copyable, because I think that's the right fix.  I don't want to change the constraint here, because I think it's correct.

adoc/chapters/programming_interface.adoc
Pennycook: I agree this is a can of worms.  I think we need to discuss it in the working group.

There is an implication that `std::iterator_traits` is specialized for `multi_ptr`, but it's not particularly strong. `accessor` and `local_accessor` expose an iterator interface and their `difference_type` is declared via `std::iterator_traits`. However, the specification says that the iterator returned for these types is unspecified.  I guess that means that if an implementation returns `multi_ptr` from `accessor::begin()`, `std::iterator_traits` must be specialized for `multi_ptr`, but otherwise ther eis no guarantee?

adoc/chapters/programming_interface.adoc
Pennycook: Good catch.  Fixed in cd2801cd02b3cd8d7c8261099e46eeecd3223eb7.

I used the wording "`marray` of a fundamental type", because there are quite a few places in the specification where we already use "`marray` of `T`" to mean "an instantiation of `marray` where `DataElement` is `T`".  For example, the descriptions of the various math overloads often say "pointer to `marray` of `int`".

adoc/chapters/programming_interface.adoc
Pennycook: Okay, I looked into this some more.  It doesn't look to be as complicated as we feared, and I don't think we actually have to do anything.

C++17 defines iterator traits [here](https://timsong-cpp.github.io/cppwp/n4659/iterator.traits#2), as:

> If `Iterator` has valid ([[temp.deduct]](https://timsong-cpp.github.io/cppwp/n4659/temp.deduct)) member types `difference_­type`, `value_­type`, `pointer`, `reference`, and `iterator_­category`, `iterator_­traits<Iterator>` shall have the following as publicly accessible members:

> ```
>  using difference_type   = typename Iterator::difference_type;
>  using value_type        = typename Iterator::value_type;
>  using pointer           = typename Iterator::pointer;
>  using reference         = typename Iterator::reference;
>  using iterator_category = typename Iterator::iterator_category;
> ```
> Otherwise, `iterator_­traits<Iterator>` shall have no members by any of the above names.

My understanding of this is that `std::iterator_traits` works automatically for any type that provides all of these members, and that seems to line up with the results of this simple test: https://godbolt.org/z/36TsK7xsq.  (If you remove any member from that test, the compiler complains that `difference_type` does not name a type.)  `multi_ptr` provides all of these members (see [Section 4.7.7.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:multiptr)) so `std::iterator_traits` should work.

@nliber - Could you please double-check the above and let me know whether you agree?

tomdeakin:
Blocked by #765

keryell (approved):
Thanks!

Issue #371, opened 2023-03-07
Clarify support for long double


tahonermann:
C++ has supported the `long double` type since C++ was first standardized. The SYCL specification does not currently acknowledge the existence of this type. It is therefore unclear what the intention is with regard to use of `long double` in SYCL programs.

Omitting support for `long double` has consequences beyond preventing programmers from directly using the type. C++11 introduced support for user-defined literals (UDLs) and the support for non-raw floating-point UDLs requires `long double` in the UDL function signature. See [\[lex.ext\]p4](http://eel.is/c++draft/lex.ext#4). For example:
```
struct my_fp_type { ... };
my_fp_type operator ""_my_fp_udl(long double);
``` 
The lack of `long double` support therefore prevents use of non-raw floating-point UDLs in SYCL device code (raw floating-point UDLs can still be defined and used).

If the intent is to omit support for `long double` in portable SYCL device code, then it would be helpful to amend the SYCL specification as follows:
- Amend [section 5.5 (Built-in scalar data types)](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:scalartypes) to explicitly note that `long double` is not specified for device code or that its size and alignment are not required to match the host. (It would be helpful to do likewise for other fundamental C++ types like `wchar_t`, `char8_t`, `char16_t`, and `char32_t`).
- Amend [section 2 (Introduction)](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#introduction) and [section 5.4 (Language restrictions for device functions)](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:language.restrictions.kernels) to list both `long double` and non-raw floating-point UDLs with the list of other C++ features that are restricted in device code.

Omission of `long double` is not the only option. There are a range of options along several dimensions. The following are a list of considerations.
- Whether the type is optionally supported (with a corresponding aspect like `fp64` for `double`).
- Whether it would be required for the size and alignment to match the host.
- Whether it would be required for the object representation to match the host.
- Whether object representation and values of the type must conform to an IEEE 754 format.
- Whether objects of the type can transit the host and device boundary and/or be accessed via USM.
[...]

PR #733, opened 2025-02-18
Clarify requirements for types in device code

### LINKED ISSUE #372: Clarify requirements for the C++ fixed width type aliases
[Section 5.5 (Built-in scalar data types)](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:scalartypes) states:

>The standard C++ fixed width types, e.g. `int8_t`, `int16_t`, `int32_t`, `int64_t`, should have the same size as defined by the C++ standard for host and device.

The use of "should" is curious here since the C++ standard already specifies (in [\[cstdint.syn\]p3](http://eel.is/c++draft/support.arith.types#cstdint.syn-3)) a particular fixed size for each of the (optional) types listed. Perhaps the intent is to state that type aliases such as `int_fast8_t`, `int_least8_t`, `intmax_t`, `intptr_t` and their unsigned variants should be defined to types that have the same size (and alignment?) on the host and device? Is it intentional, perhaps for ABI reasons, that the aliases are not required to be defined to matching types on the host and device?

This issue has been filed to update the SYCL specification to more explicitly state the intended requirements for these type aliases.

### LINKED ISSUE #373: Clarify requirements for the `std::size_t` and `std::ptrdiff_t` type aliases
The SYCL specification relies on the `std::size_t` and `std::ptrdiff_t` type aliases, but does not currently impose requirements that they be defined on the host and device to either the same type, or to types with matching size and alignment. Can objects of this type transit the host and device boundary or be accessed via USM? Since these types are used in the signatures of standard library functions, their exact definition has ABI implications.

This issue has been filed to update the SYCL specification to more explicitly state the intended requirements for these type aliases.

Pennycook:
The goal of this PR is to clarify:
- Which types are guaranteed to be available in device code; and
- What guarantees exist for consistency of types across host and device.

As part of this change I removed the table of fundamental types, which duplicated a lot of information defined in C++.  Just saying that we support the types with the same meaning as C++ seems simpler.  I kept the list of types, but long-term I'd prefer to be able to say that SYCL supports all the C++ fundamental types and leave it at that.

Closes #372 and closes #373.

---

I can't assign this to you, @tahonermann, but since you originally opened these issues I'd appreciate it if you could take a look at this proposed fix.

I've opened it as a draft because it currently doesn't build, due to a broken reference to the table I removed.  That reference will ve removed if we merge #730, though, so I wanted to open a PR and get some feedback.

### DIFF
--- a/adoc/chapters/device_compiler.adoc
+++ b/adoc/chapters/device_compiler.adoc
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
 <<device-function>>.
 ====
 
-[[subsec:scalartypes]]
-== Built-in scalar data types
+== Data types
 
-In a SYCL device compiler, the device definition of all standard {cpp}
-fundamental types from <<table.types.fundamental>> must match the host
-definition of those types, in both size and alignment.
-A device compiler may have this preconfigured so that it can match them based on
-the definitions of those types on the platform, or there may be a necessity for
-a device compiler command-line option to ensure the types are the same.
+The following types must be available in device code:
 
-The standard {cpp} fixed width types, e.g. [code]#int8_t#, [code]#int16_t#,
-[code]#int32_t#,[code]#int64_t#, should have the same size as defined by the
-{cpp} standard for host and device.
+- The fundamental data types defined by the {cpp} core language (i.e.,
+  [code]#bool#, [code]#char#, [code]#signed char#, [code]#unsigned char#,
+  [code]#short int#, [code]#unsigned short int#, [code]#int#, [code]#unsigned
+  int#, [code]#long int#, [code]#unsigned long int#, [code]#long long int#,
+  [code]#unsigned long long int#, [code]#float# and [code]#double#);
 
+- The fixed width types defined by the {cpp} core language (e.g.,
+  [code]#int8_t#, [code]#int16_t#, [code]#int32_t#, [code]#int64_t#); and
 
-[[table.types.fundamental]]
-.Fundamental data types supported by SYCL
-[width="100%",options="header",separator="@",cols="65%,35%"]
-|====
-@ Fundamental data type @ Description
-a@
-[source]
-----
-bool
-----
-   a@ A conditional data type which can be either true or false. The value
-      true expands to the integer constant 1 and the value false expands to the
-      integer constant 0.
+- Any type aliases defined by the {cpp} core language (e.g.,
+  [code]#std::size_t#, [code]#std::ptrdiff_t#).
 
-a@
-[source]
-----
-char
-----
-   a@ A signed or unsigned 8-bit integer, as defined by the {cpp} core language
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
+A device compiler may rely on additional command-line options to provide this
+guarantee for specific combinations of host and device.
 
-a@
-[source]
-----
-signed char
-----
-   a@ A signed 8-bit integer, as defined by the {cpp} core language
-
-a@
-[source]
-----
-unsigned char
-----
-   a@ An unsigned 8-bit integer, as defined by the {cpp} core language
-
-a@
-[source]
-----
-short int
-----
-   a@ A signed integer of at least 16-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-unsigned short int
-----
-   a@ An unsigned integer of at least 16-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-int
-----
-   a@ A signed integer of at least 16-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-unsigned int
-----
-   a@ An unsigned integer of at least 16-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-long int
-----
-   a@ A signed integer of at least 32-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-unsigned long int
-----
-   a@ An unsigned integer of at least 32-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-long long int
-----
-   a@ An integer of at least 64-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-unsigned long long int
-----
-   a@ An unsigned integer of at least 64-bits, as defined by the {cpp} core language
-
-a@
-[source]
-----
-float
-----
-   a@ A 32-bit floating-point. The float data type must conform to the IEEE 754
-      single precision storage format.
-
-a@
-[source]
-----
-double
-----
-   a@ A 64-bit floating-point. The double data type must conform to the IEEE 754
-      double precision storage format.  This type is only supported on devices
-      that have [code]#aspect::fp64#.
-
-|====
-
-
+The availability of these types in device code does not guarantee that they will
+be supported by all devices.
+Some types are only supported on devices with specific device aspects, as
+descrived in <<sec:device-aspects>>.
 
 == Preprocessor directives and macros
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17809,25 +17809,13 @@ std::error_code make_error_code(errc e) noexcept;
 
 == Data types
 
-SYCL as a {cpp} programming model supports the {cpp} core language data types,
-and it also provides the ability for all SYCL applications to be executed on
-SYCL compatible devices.
-The scalar and vector data types that are supported by the SYCL system are
-defined below.
-More details about the SYCL device compiler support for fundamental and backend
-interoperability types are found in <<subsec:scalartypes>>.
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
 
+=== Scalar types
 
-=== Scalar data types
-
-The fundamental {cpp} data types which are supported in SYCL are described in
-<<table.types.fundamental>>.
-Note these types are fundamental and therefore do not exist within the
-[code]#sycl# namespace.
-
-Additional scalar data types which are supported by SYCL within the [code]#sycl#
-namespace are described in <<table.types.additional>>.
-
+In addition to the scalar types defined by {cpp}, SYCL defines a number of
+scalar types as described in <<table.types.additional>>.
 
 [[table.types.additional]]
 .Additional scalar data types supported by SYCL


nliber:
Still exhausted from the WG21 meeting, so hopefully this is coherent:

If a type is an extended integer type, is that guaranteed to be supported on both the host and device?  Same question goes for extended floating point types.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
-----
-double
-----
-   a@ A 64-bit floating-point. The double data type must conform to the IEEE 754
-      double precision storage format.  This type is only supported on devices
-      that have [code]#aspect::fp64#.
-
-|====
+- Any type aliases defined by the {cpp} core language (e.g.,
+  [code]#std::size_t#, [code]#std::ptrdiff_t#).
```
tahonermann, line 233: The C++ core language doesn't define any type aliases. `std::size_t`, `std::ptrdiff_t`, `std::max_align_t`, and `std::nullptr_t` are library provided type aliases of fundamental types, some of which don't necessarily have a name otherwise. I don't think this list is intended to include all standard library type aliases, so I think this list should be explicit.

Other type aliases that are probably worth listing include:
- `int_leastN_t` (for 8, 16, 32)
- `uint_leastN_t` (for 8-, 16, 32)
- `int_fastN_t` (for 8, 16, 32)
- `uint_fastN_t` (for 8, 16, 32)
- `intptr_t`
- `uintptr_t`
- `intmax_t`
- `uintmax_t`

It should probably be specified that numeric limits have to match for all (provided) fundamental types as well.

I didn't do an exhaustive audit; there are likely other ones that should be listed.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
 
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
+A device compiler may rely on additional command-line options to provide this
+guarantee for specific combinations of host and device.
```
tahonermann, line 238: I don't think the latter sentence is helpful. The intent is to state what is required of a conforming implementation. An implementation that doesn't meet the stated requirements is non-conforming regardless of how it was configured.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
 
+The availability of these types in device code does not guarantee that they will
+be supported by all devices.
+Some types are only supported on devices with specific device aspects, as
+descrived in <<sec:device-aspects>>.
```
tahonermann, line 243:
```suggestion
The availability of these types in device code does not guarantee that they will
be supported by all devices.
Some types are only supported on devices with specific device aspects, as
described in <<sec:device-aspects>>.

Typo fix; this looks good otherwise.
```

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
-      single precision storage format.
-
-a@
-[source]
-----
-double
-----
-   a@ A 64-bit floating-point. The double data type must conform to the IEEE 754
-      double precision storage format.  This type is only supported on devices
-      that have [code]#aspect::fp64#.
```
tahonermann, line 340: Removing this loses the requirement that `float` and `double` correspond to the IEEE 754 specification. C++ doesn't restrict these types to IEEE 754 (there are C++ implementations where these types are not IEEE 754 by default). If it is intentional that SYCL strengthens the requirements relative to the C++ standard, then this still needs to be stated somewhere. I wouldn't be surprised if these requirements were not intended to deviate from the C++ standard; most C++ programmers tend to think of these types as IEEE 754 types. I'm not sufficiently familiar with their use in SYCL to have an opinion on whether this requirement is warranted.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
-
-|====
+- Any type aliases defined by the {cpp} core language (e.g.,
+  [code]#std::size_t#, [code]#std::ptrdiff_t#).
 
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
```
tahonermann, line 236: I think enumeration types should be addressed in this section as well. There is an implementation requirement as well as a user requirement. For enumerations with a fixed underlying type (e.g., `enum X : char`), there should be a requirement that the same fixed type be specified for host and device code. Similarly, for enumerations without a fixed underlying type, there should be a requirement that the implementation select the same underlying type for host and device code.

tahonermann (changes requested):
Thanks, @Pennycook. Please see inline comments.

adoc/chapters/device_compiler.adoc
Pennycook: I agree; I only left it there originally because there was a similar sentence there before.  I've removed it in 22334ab7a7def1244223901363c0d70d3d4ae271.

adoc/chapters/device_compiler.adoc
Pennycook: Good catch, thanks!  Fixed in d3238a4e649d6aad6ddc15a38f0820efcae5a762.

adoc/chapters/device_compiler.adoc
Pennycook: There was some discussion of this in internal issue 322.

This tie to IEEE 754 seems to have originated in OpenCL.  Almost every other behavior that was inherited from OpenCL has been relaxed, deprecated, or removed entirely to enable SYCL to be implemented on top of other backends.  It strikes me as odd that we've kept this restriction, especially given that C++ doesn't require it.

@gmlueck, you were involved in internal issue 322.  What do you think we should do here?  Can we relax this requirement in SYCL-Next?  Or is it something we should have removed when transitioning between SYCL 1.2.1 and SYCL 2020?

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +216,31 @@ The restriction waiver in <<discarded-statement>> or
...
+The following types must be available in device code:
 
-The standard {cpp} fixed width types, e.g. [code]#int8_t#, [code]#int16_t#,
-[code]#int32_t#,[code]#int64_t#, should have the same size as defined by the
-{cpp} standard for host and device.
+- The fundamental data types defined by the {cpp} core language (i.e.,
+  [code]#bool#, [code]#char#, [code]#signed char#, [code]#unsigned char#,
+  [code]#short int#, [code]#unsigned short int#, [code]#int#, [code]#unsigned
+  int#, [code]#long int#, [code]#unsigned long int#, [code]#long long int#,
+  [code]#unsigned long long int#, [code]#float# and [code]#double#);
```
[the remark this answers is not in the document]
Pennycook: I think we should separate clarifying what we think the existing behavior is from adding support for additional types (e..g, `long double`). 

Do we need to list all the types by name, or could we say something like:
- `void`
- `bool`
- Narrow character types
- Standard integer types
- `float`
- `double`
- The underlying type of `std:::nullptr_t`.

adoc/chapters/device_compiler.adoc
Pennycook: That makes sense to me.  I've added wording very similar to what you wrote here in f4373e2.

adoc/chapters/device_compiler.adoc
Pennycook: Listing everything might get very confusing as we move towards supporting multiple C++ versions from the same specification.  How would you feel about a list like:
- `std::size_t`
- `std::ptrdiff_t`
- `std::max_align_t`
- `std::nullptr_t`
- The type aliases for integer types defined in `<cstdint>`

Do you think we need to talk about numeric limits specifically?  Or should we have a more general note that any type traits defined in the standard library must have the same definitions for host and device?

tahonermann:
> Could we say something to the effect of: "Whether extended integer and/or extended floating-point types are available in device code is implementation-defined. However, if these types are available, the size, alignment, etc must match the host."

That makes sense to me.

adoc/chapters/device_compiler.adoc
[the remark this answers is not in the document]
tahonermann: As long as the list defers to the defined terms in the C++ standard, grouping them seems fine to me. It does make it slightly more likely that new additions within those categories will be overlooked, but I'm not too concerned about that.

The categories the standard uses overlap in some cases. For example, `signed char` and `unsigned char` (but not `char` or `char8_t`) are in both _narrow character types_ and _standard integer types_. Such overlap is ok, but can be confusing. I suggest:
- `void`
- `bool`
- _character types_ (includes `char`, `wchar_t`, `char8_t`, `char16_t`, and `char32_t`; alternatively just `char`).
- _standard integer types_
- `float`
- `double` (optionally; contingent on `aspect::fp64`)
- `long double` (optionally; no corresponding aspect)
- The underlying type of `std:::nullptr_t`

adoc/chapters/device_compiler.adoc
tahonermann: Deferring to the headers seems fine to me. It would be nice if the C++ standard defined terms for the type alias collections, but absent that, your suggestion seems fine to me.

I was just thinking of a general note about the numeric limits as well. Explicitly listing all of them would be tedious and likely of little (or possibly negative) value.

adoc/chapters/device_compiler.adoc
tahonermann: The updates for enumeration types looks good to me.

adoc/chapters/device_compiler.adoc
gmlueck: Yes, it seems weird to me also.  I think it would make sense to remove the requirement that `float` and `double` have IEEE 754 storage format.

adoc/chapters/device_compiler.adoc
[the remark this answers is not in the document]
gmlueck: You say this in the PR description:

> I kept the list of types, but long-term I'd prefer to be able to say that SYCL supports all the C++ fundamental types and leave it at that.

Could we say that device code allows all C++ fundamental types except for those listed in the section "Language restrictions for device functions"?  We could add `long double` to that list, since this is currently disallowed.  Are there other fundamental types that are not currently allowed?

It seems like it would be clearer to have an exclusion list rather than trying to list all the valid types.

Regarding type aliases in the `std` namespace, it might be better to have a new section that lists the `std` things that are guaranteed to be safe to use in device code.  This could include type aliases, some classes, and some functions.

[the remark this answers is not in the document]
tahonermann: An exclusion list is fine so long as we aren't worried about unintentionally implicitly including potential new types in newer C++ standards (C++29+) like for decimal floating-point or bit-precise integers without explicit consideration.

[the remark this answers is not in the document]
Pennycook: I like this idea.  Having an exclusion list will force us to understand the impact of adding a new C++ version in a way that having an inclusive list would not.  With an inclusive list, I fear it would be too easy to say that we had added a new C++ version, but without actually adding any new features from that version.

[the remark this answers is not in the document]
tahonermann: Is there a way to annotate the document with items that require further consideration when adding support for a new C++ version? Something like a FIXME or AUDIT comment that states something to the effect of "When adding support for a new C++ standard, consider the implications for any new fundamental types added since the previous C++ standard". Alternatively, perhaps a checklist and a GH issue for each new C++ standard to track needed updates.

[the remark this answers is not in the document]
Pennycook: Exclusion list added in e690144874fc6c10c1dd6366ab9d179299634980.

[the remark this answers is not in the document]
Pennycook: > Regarding type aliases in the `std` namespace, it might be better to have a new section that lists the `std` things that are guaranteed to be safe to use in device code. This could include type aliases, some classes, and some functions.

I've added this new section in 7b489a9689d29d040841e60a800745437aca178c.

For now, I've only added the things that we're discussing in this PR.  It talks about type aliases because they were mentioned by the old wording, and talks about `numeric_limits` because @tahonermann suggested this was an important part of device support for certain types.

I'm not opposed to adding more sections here, but we might want to consider deferring things to a later PR just to constrain the scope of this one.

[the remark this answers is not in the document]
tahonermann: I like the handling of the type aliases; looks good to me.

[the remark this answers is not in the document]
tahonermann: Should use of `long double` really result in undefined behavior? I would lean towards use of `long double` being "unspeciformed" (a WG21 CWG term meaning that it is unspecified whether the code is well-formed or has well-defined behavior). Ideally, I think it would be best to specify that it is implementation-defined whether `long double` may be used in device code, but that if it is allowed, that it must meet the "have the same size, alignment ..." requirements. But maybe that is more than we can promise.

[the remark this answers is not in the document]
tahonermann: I was hoping to avoid undefined behavior just because it is rather user unfriendly if there is a reasonable alternative. However, the issue is only intended to clarify the specification and the status quo is that `long double` is undefined behavior by omission. So, in retrospect, I agree undefined behavior is probably the right choice for now and any change to that status quo can be pursued via #371.

adoc/chapters/device_compiler.adoc
```diff
@@ -204,6 +204,7 @@ Amongst other things, this restriction makes it illegal for a
     However, a function may be defined in another translation unit if the
     implementation defines the [code]#SYCL_EXTERNAL# macro as described in
     <<subsec:syclexternal>>.
+  * Use of [code]#long double# types results in undefined behavior.
```
tahonermann, line 207: There is only one `long double` type.
```suggestion
  * Use of the [code]#long double# type results in undefined behavior.
```

Pennycook: Good catch, thanks!

Pennycook:
> Just to add a useless grain of salt, a few of our users are already using `long-double` on GPU. They just use [-mlong-double-64](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mlong-double-64) and hope for the best...

This is entirely valid under "undefined behavior".  If an implementation wants to say that it supports `long double`, it can.  If an implementation wants to say nothing but it happens to work, that's fine.

The reason I'm suggesting we don't go with "implementation-defined" is that all existing implementations would then have to update their documentation to say something about `long double`.  We'd then merge a PR (hopefully soon) which adds something like `aspect::fp128` which officially adds a way to query for `long double` support, and implementations would have to update their documentation again.

TApplencourt:
Oh yes, sorry, It was not clear.  It was 100% comment to say that I liked your clarification of new wording!

nliber:
It would be nice in the future if we could narrow down the possible behaviors.  I can think of three possibilities:

- Works
- Doesn't compile
- Random results

Is random results really a possibility?

tahonermann (approved):
This looks good to me, so I'm giving it my stamp of approval for whatever that is worth. Thanks @Pennycook!

nliber (approved):

Pennycook:
> It would be nice in the future if we could narrow down the possible behaviors. I can think of three possibilities:
> 
> * Works
> * Doesn't compile
> * Random results
> 
> Is random results really a possibility?

I might be wrong about this, but I convinced myself that random results could really be a possibility with existing SYCL 2020 implementations.

Prior to this change, we didn't say anything about `long double` at all, which means that there was no requirement for implementations to diagnose use of `long double` or to ensure that the host and device compilers agreed on what `long double` means.  Some compilers provide options to control what `long double` means (e.g., clang provides `-mlong-double-128`, `-mlong-double-64` and `-mlong-double-80`), so I'm worried it would be very easy to get in a situation where the host and device disagree about `sizeof(long double)` and things fail spectacularly.

I agree that we don't want this to be the behavior forever, but I think the right way to fix things will be via a KHR or SYCL-Next feature. There are a few things to work out (e.g., whether we'd need separate aspects for `fp80` and `fp128`, how to query which is used if a device supports both...)

adoc/chapters/device_compiler.adoc
gmlueck: I'm unresolving this conversation.  It seems confusing to me to say that applications are required to choose the same underlying type for enumerations.  Why state this specifically when there are other things an application could do that cause types to be different on host and device?  For example, if an application defines members of a struct to be different on host and device, you would have a similar problem.

If we feel we need to say something about this, perhaps we could say it more generally like:

> Applications that define types to be have different size, alignment requirement, or representation on host vs. device via the `__SYCL_DEVICE_ONLY__` macro have undefined behavior.

Or maybe it should be "... implementation defined behavior"?

I think it is useful to specifically list `__SYCL_DEVICE_ONLY__` here because that is the only way that an application possibly could define types differently.  At least I can't think of any other way.  You might point out that there are other macros like this in DPC++, but those are vendor extensions and are therefore out of scope of this spec.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +217,25 @@ The restriction waiver in <<discarded-statement>> or
...
+Some types are only supported on devices with specific device aspects, as
+described in <<sec:device-aspects>>.
 
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
 
+For enumerations with a fixed underlying type, the same fixed type must be
+specified for host and device.
+For enumerations without a fixed underlying type, the implementation must select
+the same underlying type on host and device.
```
gmlueck, line 238: This second sentence is directed at implementors of the spec.  I think this is redundant.  We already say that "All types which are available in device code" must have the same size and representation.  This statement covers enumerations also.

adoc/chapters/device_compiler.adoc
```diff
@@ -204,6 +204,7 @@ Amongst other things, this restriction makes it illegal for a
     However, a function may be defined in another translation unit if the
     implementation defines the [code]#SYCL_EXTERNAL# macro as described in
     <<subsec:syclexternal>>.
+  * Use of the [code]#long double# type results in undefined behavior.
```
gmlueck, line 207: FWIW, I'd say:

> Device code must not use the `long double` type.

Statements in this list are inconsistent.  Some say that using a prohibited feature causes undefined behavior, while others say that application should not use the feature.  It seems to me that we should consistently say that device code must not use these features.

You may argue that an extension might lift some of these restrictions.  That's fine.  An extension is allowed to lift any prohibition in the spec and assign some defined meaning.  There is no need to use the "undefined behavior" wording to allow an extension to do this.

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +217,25 @@ The restriction waiver in <<discarded-statement>> or
 <<device-function>>.
 ====
 
-[[subsec:scalartypes]]
```
gmlueck, line 219: Note that the latest version of the "main" branch has text that refers to this section identifier.  (This was part of the recent `vec` clarification.)  You will need to reword that new text somehow to avoid the reference.

adoc/chapters/device_compiler.adoc
tahonermann: That statement partially covers enumerations, but a different underlying type would still be observable via [`std::underlying_type`](https://en.cppreference.com/w/cpp/types/underlying_type). What is gained by not requiring implementations to use consistent underlying types? Motivation for UB allowances should be strong.

adoc/chapters/device_compiler.adoc
tahonermann: I think it is reasonable to relax the requirement that (user specified) fixed underlying types must match so long as device copyability rules retain a requirement for the same (or at least a compatible) fixed underlying type.

Pennycook: I think a general note like the one @gmlueck suggests would be useful for application developers, but I didn't think that was what @tahonermann was talking about here.

The wording I added in f4373e2ab29ac15636fd4c36c1cac8db109ce6b8 is in the "Device compiler" section, and I didn't intend it to be advice for application developers. The intent was to clarify that the implementation must guarantee that enumerations have matching types.  The first sentence just says that enumerations with different fixed underlying types are invalid -- that means a device compiler (and more broadly, the implementation) shouldn't introduce any of those.  The second sentence is probably the more important one, because it means that a device compiler may require specific logic to ensure that it selects the same underlying type as the host compiler if an enumeration has no fixed underlying type.

adoc/chapters/device_compiler.adoc
gmlueck: OK.  I was not aware of [std::underlying_type](https://en.cppreference.com/w/cpp/types/underlying_type), so I agree this statement is useful.

adoc/chapters/device_compiler.adoc
gmlueck: > I had previously suggested that we address `long double` via #371. I think undefined behavior is the status quo.

I still like the direction proposed in #371 -- that is that we make `long double` an optional device feature that is tied to some new aspect.  However, that is probably something that needs to wait for SYCL-Next.

I think the status quo in DPC++ is that use of `long double` in device code is ill-formed.  I think the compiler diagnoses a compile-time error for this.  (Sorry, haven't checked lately, but this is my memory.)

adoc/chapters/device_compiler.adoc
```diff
@@ -216,132 +217,25 @@ The restriction waiver in <<discarded-statement>> or
...
+The availability of a type in device code does not guarantee that it will be
+supported by all devices.
+Some types are only supported on devices with specific device aspects, as
+described in <<sec:device-aspects>>.
 
+All types which are available in device code must have the same size, alignment
+requirements, and representation on both host and device.
 
+For enumerations with a fixed underlying type, the same fixed type must be
+specified for host and device.
```
gmlueck, line 236: My comment [here](https://github.com/KhronosGroup/SYCL-Docs/pull/733/files#r1974343938) might have been confusing.  I was referring to this sentence "For enumerations with a fixed underlying type, the same fixed type must be specified for host and device."

This sentence seems to be directed to application developers.  If an enumeration has a fixed underlying type, it means the source code explicitly specifies the type.  Therefore, there's no possibility that the device compiler might choose some other type -- it has to use the type specified in the application source code.

Therefore, I think this sentence must be telling application developers not to use `__SYCL_DEVICE_ONLY__` to specify a different underlying type in host vs. device code.  This seems weirdly specific because there are other ways an application can use that macro to cause types to have different representations on host vs. device.

adoc/chapters/device_compiler.adoc
Pennycook: > I think the status quo in DPC++ is that use of `long double` in device code is ill-formed. I think the compiler diagnoses a compile-time error for this. (Sorry, haven't checked lately, but this is my memory.)

I was curious about this so I just went to check.

By default, DPC++ throws an error because `long double` defaults to 128-bit, which isn't supported in SPIR-V:
```
test.cpp:10:51: error: 'max' requires 128 bit size 'long double' type support, but target 'spir64-unknown-unknown' does not support it
   10 |         ld[0] = std::numeric_limits<long double>::max();
```

You can circumvent that check by adding the `-mlong-double-64` flag.  In other words, DPC++ doesn't check for `long double`, it just fails if you try to use 128- or 80-bit data types.

SimSYCL also doesn't do any checks for use of `long double`, and because it's a header-only implementation I can't immediately think of a way they could add these checks.  It's also worth noting that device code using `long double` always works in SimSYCL, because the code is running on the host CPU.

So I think the status quo is undefined behavior.  Sometimes it's ill-formed, sometimes you can make it work (by accident), and sometimes it works.

adoc/chapters/device_compiler.adoc
gmlueck: IMO we should remove the chapter "SYCL device compiler" and merge it into "SYCL programming interface".  The spec should not mandate that there is a distinct "device compiler".  Some implementations (using multi pass compiler) might have a "device compiler", but other implementations might not.

Therefore, I think the spec should not try to draw a distinction between "device compiler" and "implementation".  If an implementation has a device compiler, it's just part of the implementation.

Pennycook: I agree 100%!  I can never figure out where to introduce new sections in the specification, because I don't understand where the boundaries are drawn.  Merging all the implementation stuff into one section would be great.

But without a boundary between the device compiler and implementation, I think what I said holds: the implementation includes some headers, and those headers might not (probably won't) be completely identical for the host and device. Do you agree that the specification needs to point out which parts of the headers _are_ required to be identical?  If so, why wouldn't we extend this to the underlying types of enums?

adoc/chapters/device_compiler.adoc
TApplencourt: >You can circumvent that check by adding the -mlong-double-64 flag. In other words, DPC++ doesn't check for long double, it just fails if you try to use 128- or 80-bit data types.

Yes, and some people are Argonne is using this flag in production

gmlueck: > You can circumvent that check by adding the `-mlong-double-64` flag. In other words, DPC++ doesn't check for `long double`, it just fails if you try to use 128- or 80-bit data types.

I'm confused.  What SPIR-V does DPC++ generate if you compile with `-mlong-double-64`?

Or, are you saying that the flag allows you to check the trait, but the compiler still raises a diagnostic if the device code declares a variable of type `long double`?

Pennycook: > I'm confused. What SPIR-V does DPC++ generate if you compile with `-mlong-double-64`?

It just treats every instance of `long double` as if it were `double`, so everything is generated using `OpTypeFloat 64`.

I'm not convinced it works properly, because all of the tests I've tried so far have returned `nan`.  But the code compiles without producing any errors or warnings, and runs without producing any errors or warnings.

gmlueck: Oh, I see.

> Yes, and some people are Argonne is using this flag in production

Why?  Is it just because you have existing code that uses `long double` and it's easier to pass the flag than to rewrite the code to use `double`?

TApplencourt: Exactly. It begs the question why it even works, and why they `long double` in the first place, but 🤷🏽

adoc/chapters/device_compiler.adoc
tahonermann: With regard to explicit fixed types, there is both an implementation and a requirement on application developers.

I expect device code restrictions to be relaxed over time to allow more use of the C and C++ standard libraries in device code. Device code can already link with device-only TUs via `SYCL_EXTERNAL`. Maintaining a single C and C++ standard library that works for both host and device code isn't particularly realistic, especially when the host implementations are provided by, e.g., GNU C, libstdc++, Microsoft, etc... This increases the likelihood that enumeration types defined in these libraries will be differently specified for the host and the device. This imposes a requirement on the SYCL implementation to ensure such enumeration types are consistently defined if objects of those types might transition the host/device divide. We can either require all such types to be consistently defined, or we can require specific ones to be consistently defined so that those specific types satisfy device copyability rules, or we can say its all implementation-defined and code that uses such types is non-portable.

Likewise, user-defined enumeration types must be consistently defined to be copied between the host and device. For those cases, we can require all such types to be consistently defined, or we can place the requirement in the device copyable restrictions (with or without a compatible size/alignment allowance and corresponding `std::underlying_type` observability).

adoc/chapters/device_compiler.adoc
tahonermann: Perhaps Argonne needs it because they are using user-defined literals (UDLs). A request for UDL support is what lead me to file #371 in the first place.

adoc/chapters/device_compiler.adoc
tahonermann: Perhaps my recent response [here](https://github.com/KhronosGroup/SYCL-Docs/pull/733#discussion_r1976084856) would have been more appropriate as a reply to this comment.

The C++ standard doesn't specify many enumeration types for the standard library. Where the C++ standard does specify an enumeration and doesn't explicitly specify an underlying type, I think implementors are free to specify a fixed underlying type if they desire. That is one way that type definitions might diverge.

SYCL 2020 has exactly one match for the term "standard library" and that is a reference to the C++ standard library in a note about differences from SYCL 1.2.1. I guess it is currently undefined which portions of the C++ standard library may be used in device code. Some portions can be inferred as required to be available in device code based on usage in SYCL interfaces (e.g., `std::vector`). It seems like a lot of work would be required to specify what is actually required to be usable in device code.

A strong +1 for excising the term "device compiler" from the SYCL specification :)

adoc/chapters/device_compiler.adoc
```diff
@@ -387,6 +281,25 @@ literal with greater value.
 In addition, for each <<backend>> supported, the preprocessor macros described
 in <<sec:backends>> must be defined by all conformant implementations.
 
+[[sec:standard-library-support]]
+== {cpp} standard library support
+
```
gmlueck, line 286: This section is the first time we specifically guarantee that anything in the `std` namespace can be used in device code.  I think we need to prefix this section with an introductory statement something like:

> In general, any use of functions or types from the `std` namespace in device code produces undefined behavior.  However, the functions and types listed in this section are guaranteed to have defined semantics when used in device code.

adoc/chapters/device_compiler.adoc
gmlueck: > The C++ standard doesn't specify many enumeration types for the standard library. Where the C++ standard does specify an enumeration and doesn't explicitly specify an underlying type, I think implementors are free to specify a fixed underlying type if they desire. That is one way that type definitions might diverge.

I agree that we want to disallow an implementation like this.  However, I think this case is already covered by this statement in the spec:

> All types which are available in device code must have the same size, alignment requirements, and representation on both host and device.

That statement applies to *all* types, which includes enumeration types from the C++ standard library.

I don't think we need to introduce another statement into the spec that redundantly provides this guarantee for enumeration types.

adoc/chapters/device_compiler.adoc
Pennycook: Added your proposed paragraph in da1c0db2e23e1ad79f35c6970975ec49fd370201.

adoc/chapters/device_compiler.adoc
Pennycook: I've gone ahead and removed this in a339d3b7c442d90ed2dda6b709fd5b3dace2e871.

I kept the note about the types without specified underlying types -- I don't think that behavior is as obvious and there's a good chance that an implementation will have to do something special to enforce it.  I think it's worth calling out cases like that, even if it's arguably redundant.  (If people think it's too redundant, I would be comfortable with the note about these enums being non-normative.)

adoc/chapters/device_compiler.adoc
```diff
@@ -387,6 +279,30 @@ literal with greater value.
...
+defined semantics when used in device code.
+
+=== Type aliases
+
+The following type aliases must alias the same types on host and device:
+
+  * [code]#std::size_t#;
+  * [code]#std::ptrdiff_t#;
+  * [code]#std::max_align_t#;
+  * [code]#std::nullptr_t#; and
```
keryell, line 297: Thinking about it, I was wondering about how this is implemented in C++,  looking at `/usr/include/x86_64-linux-gnu/c++/14/bits/c++config.h` on my laptop, I can see:
```c
  typedef decltype(nullptr)	nullptr_t;
```
which follows https://en.cppreference.com/w/cpp/types/nullptr_t

But how an implementation using for example a different host and device compiler can ensure that these are the same types on host and device side? :thinking: 
Probably at the end, we do not care and the only important aspect is *`sizeof(std::nullptr_t)` is equal to `sizeof(void *)`*.

keryell (approved):
Thanks for the simplification and clarification.

adoc/chapters/device_compiler.adoc
gmlueck: > Probably at the end, we do not care and the only important aspect is _`sizeof(std::nullptr_t)` is equal to `sizeof(void *)`_.

I think we do care that they are the same type on host vs. device.  Otherwise, you could construct constexpr values that are different on host vs. device via `decltype`.

tahonermann: It is also important that the types of the type alias targets match so that mangled names can be correlated across host and device compilation. Consider a SYCL kernel name type that is a class template specialization with a `std::size_t` template argument.

keryell: The problem with `nullptr_t` is that there is a chicken-egg problem since it is defined as the type of `nullptr` and to pass such a value to a device you need to have the device type at the first place, but this is defined from the value...
The mangled name aspect is interesting. Probably the C++ ABI has some mangling for `nullptr_t` too.

tomdeakin:
@Pennycook @gmlueck - is this still in drafting?

gmlueck:
> @Pennycook @gmlueck - is this still in drafting?

I think it needs a new owner.  Some of the comments above (some of which are from me) are not resolved.

PR #717, opened 2025-02-12
Replace "inverse sqrt" with "reciprocal" sqrt

Pennycook:
The "inverse" of sqrt is x*x. Referring to "reciprocal" sqrt is more accurate.

Closes #655.

PR #736, opened 2025-02-24
Replace "inverse sqrt" with "reciprocal" sqrt

gmlueck:
Cherry pick #717 from main
(cherry picked from commit 768e91325ba840cf12fca22d930a344593c046b5)


gmlueck:
Merging because the WG already approved #717

PR #738, opened 2025-02-24
Remove minimum max_mem_alloc_size

### LINKED ISSUE #378: possibly incorrect minimum for info::device::max_mem_alloc_size
The current required minimum for info::device::max_mem_alloc_size is:

> Returns the maximum size of memory object allocation in bytes. The minimum value is max (1/4th of info::device::global_mem_size,128*1024*1024) if this SYCL device is not of device type info::device_type::custom.

I suspect this definition came from OpenCL 1.2, which had the same definition for CL_DEVICE_MAX_MEM_ALLOC_SIZE.

The [definition in OpenCL 2.0 and newer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_MEM_ALLOC_SIZE) though is:

> Max size of memory object allocation in bytes. The minimum value is max(min(1024 × 1024 × 1024, 1/4th of CL_DEVICE_GLOBAL_MEM_SIZE), 32 × 1024 × 1024) for devices that are not of type CL_DEVICE_TYPE_CUSTOM.

Note that this caps the minimum at 1GB.

Should the SYCL minimum have the same 1GB cap?  Or is the difference intentional, and the SYCL minimum should be 1/4 of the total global memory size, even when the total global memory size exceeds 4GB?

Pennycook:
This minimum value was originally inherited from OpenCL 1.2. Most OpenCL 1.2-specific text was removed between SYCL 1.2.1 and SYCL 2020, but this was missed.

This change doesn't prevent OpenCL-based implementations from returning these values, but removes the restriction that implementations using other backends must enforce the same limits as OpenCL.

Closes #378.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2112,9 +2112,6 @@ struct max_mem_alloc_size {
 _Remarks:_ Template parameter to [api]#device::get_info#.
 
 _Returns:_ The maximum size of memory object allocation in bytes.
-The minimum value is max (1/4th of
-[code]#info::device::global_mem_size#,128*1024*1024) if this device is not of
-device type [api]#info::device_type::custom#.
 
 '''
 


gmlueck (approved):

tomdeakin:
Wg approved to merge, clarification to 2020

PR #740, opened 2025-02-24
Clarify `is_available`

### LINKED ISSUE #347: Device availability in not specified in details
In the class [Device](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-class) there is a method `Param::return_type get_info() ` which returns [Device information descriptors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors). 
The one of them is **`info::device::is_available`** with the description: 

> Returns true if the SYCL device is available and returns false if the device is not available.

What does this **availability** mean? - This information is missing in the specifications. 
Does it mean that device just available in the system in general? Or does it mean that the device is ready for execution?

Pennycook:
The `is_available` query is based on `CL_DEVICE_AVAILABLE` from OpenCL. The most recent OpenCL standard includes an extra sentence that wasn't copied over to SYCL 2020, and this PR fixes that.

I also added an extra sentence to clarify that what causes a device to move between available and not available is implementation-defined.  This wasn't part of OpenCL, but I can't see a way to avoid specifying it this way in SYCL: the least surprising way to implement this on an OpenCL backend would be to say "It returns the result of `CL_DEVICE_AVAILABLE`", but that is only possible if it's implementation-defined.

Closes #347.

---

My personal opinion is that this query isn't very useful, and that we should deprecate it and remove it in SYCL-Next. But that's a separate issue to what the behavior should be in SYCL 2020.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2759,6 +2759,8 @@ _Remarks:_ Template parameter to [api]#device::get_info#.
 
 _Returns:_ The value [code]#true# if the device is available and [code]#false#
 if the device is not available.
+A device is considered to be available if the device can be expected to
+successfully execute commands enqueued to the device.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -2761,6 +2761,8 @@ _Returns:_ The value [code]#true# if the device is available and [code]#false#
 if the device is not available.
 A device is considered to be available if the device can be expected to
 successfully execute commands enqueued to the device.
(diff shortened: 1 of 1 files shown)


gmlueck (approved):

gmlueck:
> My personal opinion is that this query isn't very useful, and that we should deprecate it and remove it in SYCL-Next. But that's a separate issue to what the behavior should be in SYCL 2020.

+1

keryell (approved):
Thanks for the clarification.

tomdeakin (approved):

AerialMantis (approved):

PR #743, opened 2025-02-24
First draft of process documentation

gmlueck:
Add the first draft of a document describing our repository processes.

### DIFF
--- a/README_PROCESS.md
+++ b/README_PROCESS.md
@@ -106,7 +106,7 @@ After cherry picking PRs from "main" to "sycl-xxxx", it's useful to compare the
 two branches and check that there are no unexpected differences:
 
 ```
-$ git diff sycl-2020 main
+$ git diff sycl-xxxx main
 ```
 
 
@@ -116,7 +116,7 @@ $ git diff sycl-2020 main
 We should check this process again when publishing revision 10 to make sure they
 are accurate.
 
-The working group normally accumulates a number "clarification" PRs in this
+The working group normally accumulates a number of "clarification" PRs in this
 GitHub repository before publishing them as an update revision of the current
 specification.
 When the working group decides it is time to publish an update revision, we
--- a/README_PROCESS.md
+++ b/README_PROCESS.md
@@ -1,0 +1,214 @@
+## Repository branch structure
+
+The SYCL specification intends to have periodic major versions, where each is
+named according to the year in which it is approved by the working group.
+Although there were previous major versions, this repository starts with the
+SYCL 2020 specification.
+From this point forward, the SYCL specification is "unified", meaning that a
+single document serves as the official specification for all major versions
+from SYCL 2020 onwards.
+
+There are often two active branches in the repository: a branch named "main" and
+a branch whose name has the form "sycl-xxxx" (e.g. "sycl-2020").
+The "main" branch always exists, and it represents the most current version of
+the specification.
+Depending on the release cycle, "main" may represent the major version that is
+currently published, or it may represent a future major version that is not yet
+published.
+When "main" represents an unpublished major version, there is always a branch
+named "sycl-xxxx", which represents the major version that is currently
+published.
+This allows us to work on the next unpublished major version while still doing
+maintenance releases on the published version.
+
+Bug fixes and clarifications to the specification are generally committed to
+both the "main" branch and to the "sycl-xxxx" branch (when it exists).
+New content that is destined only for the next as-of-yet unpublished major
+version is committed only to the "main" branch.
+
+Once a new major version is published from "main", it is also the specification
+for all older major versions.
+Therefore, when we publish a new major version from "main", we delete the
+"sycl-xxxx" branch because it is no longer needed.
+
+We create a new "sycl-xxxx" branch immediately before making the first commit to
+"main" that we don't want to appear in the currently published major version.
+
+
+## Cherry pick strategy
+
+Because most changes should be made to the most current specification, we
+normally draft PRs against the "main" branch.
+Once a PR is merged, we may also cherry pick it to the "sycl-xxxx" branch if
+the change is a clarification (and if the "sycl-xxxx" branch exists).
+
+We currently use a manual process for these cherry picks, though we would like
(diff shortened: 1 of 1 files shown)


gmlueck:
@Pennycook since you volunteered to cover spec editor duties while I am away, I think it makes sense for you to review this PR.

README_PROCESS.md
```diff
@@ -0,0 +1,221 @@
...
+The SYCL specification intends to have periodic major versions, where each is
+named according to the year in which it is approved by the working group.
+Although there were previous major versions, this repository starts with the
+SYCL 2020 specification.
+From this point forward, the SYCL specification is "unified", meaning that a
+single document serves as the official specification for all major versions
+from SYCL 2020 onwards.
+
+There are often two active branches in the repository: a branch named "main" and
+a branch whose name has the form "sycl-xxxx" (e.g. "sycl-2020").
```
Pennycook, line 12: Probably not worth it at this point, but it might be worth exploring if a name like "develop" would help people to understand that the branch does not correspond to a released specification.

README_PROCESS.md
```diff
@@ -0,0 +1,221 @@
...
+The "main" branch always exists, and it represents the most current version of
+the specification.
+Depending on the release cycle, "main" may represent the major version that is
+currently published, or it may represent a future major version that is not yet
+published.
```
Pennycook, line 17: I'm a little confused by this.

It sounds like you're saying that "main" might contain:
- A specification that has been ratified and published, and nobody's made any changes to it yet; or
- A specification that hasn't been ratified or published.

...but I'm not sure I understand why this is relevant to GitHub users, so I think I might have misunderstood something. It seems like we wouldn't be in the first situation for very long.

README_PROCESS.md
```diff
@@ -0,0 +1,221 @@
...
+  it is pushed to the PR, but this doesn't help.
+  I think GitHub determines the author from the owner of the PR (not from any
+  field in the commit), and GitHub does not provide a way to change the owner of
+  a PR.
+
+After cherry picking PRs from "main" to "sycl-xxxx", it's useful to compare the
+two branches and check that there are no unexpected differences:
+
+```
+$ git diff sycl-2020 main
```
Pennycook, line 109:
```suggestion
$ git diff sycl-xxxx main
```

README_PROCESS.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+
+## Publishing an update revision
+
+**TODO:** These are my notes from publishing SYCL 2020 revision 9.
+We should check this process again when publishing revision 10 to make sure they
+are accurate.
+
+The working group normally accumulates a number "clarification" PRs in this
```
Pennycook, line 119:
```suggestion
The working group normally accumulates a number of "clarification" PRs in this
```

README_PROCESS.md
```diff
@@ -0,0 +1,221 @@
...
+GitHub repository before publishing them as an update revision of the current
+specification.
+When the working group decides it is time to publish an update revision, we
+follow this process.
+
+As noted above, the currently published major version may either correspond to
+the "main" branch or to the "sycl-xxxx" branch, depending on release cycle.
+The instructions below refer to the branch that contains the currently published
+major version as `<current>`.
```
Pennycook, line 128: It's still not clear to me how to determine which branch to use here.

If we're publishing an update, why wouldn't it always be the "sycl-xxxx" branch?

README_PROCESS.md
gmlueck: Based on your comments, I think I did not explain the branch structure well.  I tried rewriting the first section in db4aeec94efa742077b2a2c0e0ea3fe787d8473c.  Does the new introduction make it clearer?

README_PROCESS.md
```diff
@@ -8,31 +8,41 @@ From this point forward, the SYCL specification is "unified", meaning that a
...
+the "main" branch at this point, and we will publish update revisions from this
+branch.
+
+At some later point, we will start development of another major release
+SYCL-Next2.
+Immediately before making the first commit that is specific to SYCL-Next2, we
+will create a branch named "sycl-xxxx" (where "xxxx" represents the year in
+which SYCL-Next was approved by the working group).
+Bug fixes and clarifications to SYCL-Next will happen on this "sycl-xxxx" branch
+while new development happens on "main".
```
Pennycook, line 45: Thanks for rewriting this section, this new text is much clearer.  I only have one question about this highlighted part...

Why would we go back to having separate branches once we've released our first "unified" specification?  Isn't the benefit of things being "unified" that we no longer have to worry about maintaining multiple versions?

If we _don't_ create a "sycl-xxxx" branch for SYCL-Next2, any clarifications and bugfixes made in "main" will automatically apply retroactively.  Any new features that are specific to SYCL-Next2 would be guarded by text like "Only in SYCL-Next2".

Is the issue here that we want to allow SYCL-Next updates even before SYCL-Next2 features are ratified?

gmlueck: > Is the issue here that we want to allow SYCL-Next updates even before SYCL-Next2 features are ratified?

Yes, exactly.  This is the reason we have two branches now ("main" and "sycl-2020").  We want to be able to work on SYCL-Next in the "main" branch while also publishing update revisions to SYCL-2020.  We can't publish an update revision from "main" right now because it contains unratified text for SYCL-Next.

The paragraph here just moves forward in time to a point where SYCL-Next is published and we are working on the SYCL-Next-Next release (which I name "SYCL-Next2" in the text).  At that point, we can't publish updates from "main" because it contains unratified text from SYCL-Next-Next.  Therefore, we need a new branch "sycl-xxxx" that allows us to publish updates to the (ratified) SYCL-Next.

Pennycook: I see, okay.  I suppose the alternative (working directly on the unified version, and rebasing SYCL-Next-Next on top of it) is just as manual as doing the cherry-picking.  Thanks for the explanation.

Pennycook (approved):

gmlueck:
Merging as editorial

PR #749, opened 2025-02-24
Fix definition of RET for vec operators

gmlueck:
Cherry pick #666 from main
(cherry picked from commit e8a32745cb79d80a00410300ce1750de0cf60fd6)


gmlueck:
Merging because the WG already approved #666

PR #754, opened 2025-02-25
Remove references to `compile`/`build_with_source`

### LINKED ISSUE #389: References to build_with_source and compile_with_source
This section of the SYCL 2020 spec refers to `build_with_source` and `compile_with_source` functions, but they aren't described anywhere else in the spec.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_offline_linking_with_opencl_c_libraries

I suspect this is an artifact leftover from a previous SYCL version?  If so, these references should be removed.  If not, how should they be used?

It may be possible to cleanup other parts of this section as well, or to remove it entirely if it's no longer applicable.

Pennycook:
These functions were removed alongside the `program` object in SYCL 2020.

The section correctly pointed to `kernel_bundle` as the replacement, but referred to member functions that didn't exist.

Closes #389.

---

I'm honestly not sure whether I've fixed this correctly, and I was surprised to discover that we define both a `make_kernel_bundle` and a `create_bundle` function.  I _think_ this is correct, though -- a developer can create a `cl_program` from source using OpenCL functionality, and then create a `kernel_bundle` from that.

### DIFF
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -1056,9 +1056,9 @@ C libraries during offline compilation or during online compilation by the
 Linking with OpenCL C kernel functions offline is an optional feature and is
 unspecified.
 Linking with OpenCL C kernel functions online is performed via the
-[code]#sycl::make_kernel_bundle# or [code]#sycl::opencl::create_bundle# functions, which both
-provide a mechanism to create an instance of a SYCL [code]#kernel_bundle# from a
-[code]#cl_program#.
+[code]#sycl::make_kernel_bundle# or [code]#sycl::opencl::create_bundle#
+functions, which both provide a mechanism to create an instance of a SYCL
+[code]#kernel_bundle# from a [code]#cl_program#.
 
 OpenCL C functions that are linked with, using either offline or online
 compilation, must be declared as [code]#extern "C"# function declarations.
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -1056,7 +1056,7 @@ C libraries during offline compilation or during online compilation by the
 Linking with OpenCL C kernel functions offline is an optional feature and is
 unspecified.
 Linking with OpenCL C kernel functions online is performed via the
-[code]#make_kernel_bundle# or [code]#create_kernel_bundle# functions, which both
+[code]#sycl::make_kernel_bundle# or [code]#sycl::opencl::create_bundle# functions, which both
 provide a mechanism to create an instance of a SYCL [code]#kernel_bundle# from a
 [code]#cl_program#.
 
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -1055,9 +1055,10 @@ C libraries during offline compilation or during online compilation by the
 
 Linking with OpenCL C kernel functions offline is an optional feature and is
 unspecified.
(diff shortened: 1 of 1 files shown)


adoc/chapters/opencl_backend.adoc
```diff
@@ -1055,9 +1055,10 @@ C libraries during offline compilation or during online compilation by the
 
 Linking with OpenCL C kernel functions offline is an optional feature and is
 unspecified.
-Linking with OpenCL C kernel functions online is performed by using the SYCL
-[code]#kernel_bundle# class to compile and link an OpenCL C source; using the
-[code]#compile_with_source# or [code]#build_with_source# member functions.
+Linking with OpenCL C kernel functions online is performed via the
+[code]#make_kernel_bundle# or [code]#create_kernel_bundle# functions, which both
```
gmlueck, line 1059:
```suggestion
[code]#sycl::make_kernel_bundle# or [code]#sycl::opencl::create_bundle# functions, which both
```

Note that the two functions are in different namespaces, so it's probably good to show the fully qualified names here.

gmlueck (commented):
> I was surprised to discover that we define both a `make_kernel_bundle` and a `create_bundle` function

This occurs because the signature of `make_kernel_bundle` has to have a specific set of parameters that are dictated by the `backend_traits`.  See [section 4.5.1.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-interoperability-make).  Personally, I'd like to remove all these `make_` functions from the core spec and let each backend define its own set of functions that construct SYCL objects from backend objects.

Pennycook:
> This occurs because the signature of `make_kernel_bundle` has to have a specific set of parameters that are dictated by the `backend_traits`. See [section 4.5.1.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-interoperability-make). Personally, I'd like to remove all these `make_` functions from the core spec and let each backend define its own set of functions that construct SYCL objects from backend objects.

I agree.  Trying to force all of the backend interoperability functions to use the same interface is just going to make lots of work for the WG and implementers: either the interface has to keep being updated to chase the latest implementation requirements, or implementers have to constrain features and capabilities in order to satisfy interoperability.

@tomdeakin, could we make time in the SYCL WG agenda to discuss potentially deprecating all of these `make_` functions?

AerialMantis (approved):

illuhad (approved):

tomdeakin (approved):

bcalidas:
We are ok with this PR.  Would also like to add that the make_kernel path is  being used for OpenCL interop. In general,  the  make_ functions are extremely important for us and should remain in the core spec.

Thanks,
Balaji Calidas
Qualcomm

Pennycook:
> We are ok with this PR. Would also like to add that the make_kernel path is being used for OpenCL interop. In general, the make_ functions are extremely important for us and should remain in the core spec.

Would you be opposed to moving them into a dedicated OpenCL namespace?  The deprecation path I see is something like:

- No change to SYCL 2020.
- In SYCL-Next, `sycl::make_kernel` is deprecated in favor of something like `sycl::opencl::make_kernel`.
- In SYCL-Next-Next, `sycl::make_kernel` is removed.

So, it wouldn't be removed any time soon, there'd be a simple migration path (just a new namespace) and we'd be free to let each backend define its own interoperability interfaces.

keryell (approved):

tomdeakin:
WG approved to merge

PR #758, opened 2025-02-27
Fix description of `sycl::free`

Pennycook:
This PR makes three changes, which I've split into three commits to simplify review.

1. The description of `sycl::free` is migrated to the new format.

2. An unnecessary statement about "waiting" is removed.  This statement was already covered by (and in fact, inconsistent with!) `sycl::free` having undefined behavior in the case where `sycl::free` is called while a command is in-progress.  Whether an implementation waits for in-progress commands or not doesn't matter, because that usage is invalid.

3. I added a non-normative note to highlight that `sycl::free` is not guaranteed to be blocking or non-blocking.  The old wording implied that `sycl::free` was always guaranteed to be non-blocking, but this isn't the case for existing implementations.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11777,35 +11777,31 @@ necessary [code]#device#.
 
 ==== Memory deallocation functions
 
-[[table.usm.free]]
-.USM Deallocation Functions
-[width="100%",options="header",separator="@",cols="65%,35%"]
-|====
-@ Function @ Description
-a@
-[source]
+.[apidef]#free#
+[source,role=synopsis,id=api:free]
 ----
-void sycl::free(void* ptr, const context& syclContext)
+void free(void* ptr, const context& ctxt); (1)
+void free(void* ptr, const queue& q);      (2)
 ----
-a@ Frees an allocation.
-The memory pointed to by [code]#ptr# must have been allocated using one of the
-USM allocation routines or it must be a null pointer.
-If [code]#ptr# is not null, the [code]#syclContext# must be the same
-[code]#context# that was used to allocate the memory.
-If [code]#ptr# is null, this function has no effect.
-Otherwise, the memory is freed without waiting for <<command, commands>>
-operating on it to be completed.
-If <<command, commands>> that use this memory are in-progress or are enqueued,
-the behavior is undefined.
 
-a@
-[source]
-----
-void sycl::free(void* ptr, const queue& syclQueue)
-----
-a@ Alternate form where [code]#syclQueue# provides the [code]#context#.
+*Overload (1):*
 
-|====
+_Preconditions_:
+
+* [code]#ptr# points to memory allocated against [code]#ctxt# using one of the
+  USM allocation routines, or is a null pointer; and
+
+* There are no in-progress or enqueued <<command, commands>> using the memory
+  pointed to by [code]#ptr#.
+
+_Effects_: The memory pointed to by [code]#ptr# is freed, without waiting for
+<<command, commands>> operating on it to be completed.
+
+_Remarks_: If [code]#ptr# is null, this function has no effect.
+
+*Overload (2):*
+
+_Effects_: Equivalent to [code]#return free(ptr, q.get_context());#.
 
 === Unified shared memory pointer queries
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11789,7 +11789,9 @@ void free(void* ptr, const queue& q);      (2)
 _Preconditions_:
 
 * [code]#ptr# points to memory allocated against [code]#ctxt# using one of the
-  USM allocation routines, or is a null pointer; and
+  USM allocation routines, or is a null pointer;
+
+* [code]#ptr# has not previously been deallocated; and
 
 * There are no in-progress or enqueued <<command, commands>> using the memory
   pointed to by [code]#ptr#.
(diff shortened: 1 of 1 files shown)


TApplencourt:
Thanks from moving to the new format!

>The old wording implied that sycl::free was always guaranteed to be non-blocking, but this isn't the case for existing implementations.

What do you mean? 

- `cudaFree`is blocking for example, so I suspect all the cuda back to be blocking.
- And the fact that DPCPP may use `zeMemFree` and not the new `zeMemFree` ext with the `ZE_DRIVER_MEMORY_FREE_POLICY_EXT_FLAG_BLOCKING_FREE` is an implementation detail.

So I'm not really sure buy removing the `non-blocking`.  It seems to be a breaking change; who may break codes (at least 
 conceptually)

Pennycook:
> What do you mean?

I mean that the specification didn't actually say "`sycl::free` is non-blocking" -- it just said that calling `sycl::free` while a command was executing is undefined behavior.

> So I'm not really sure buy removing the `non-blocking`. It seems to be a breaking change; who may break codes (at least conceptually)

I'm not sure what you mean.

You said that existing CUDA backends are probably blocking (and I agree).  That's what I meant about the behavior of existing implementations: there are existing implementations of SYCL that implement `sycl::free` as blocking.

gmlueck (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -11789,12 +11789,14 @@ void free(void* ptr, const queue& q);      (2)
 _Preconditions_:
 
 * [code]#ptr# points to memory allocated against [code]#ctxt# using one of the
-  USM allocation routines, or is a null pointer; and
+  USM allocation routines, or is a null pointer;
+
```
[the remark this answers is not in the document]
Pennycook: I don't think we strictly need it.  I try and leave an empty line between bullets for readability (especially if the bullet text wraps onto a second line) but we don't seem to be very consistent about it.

adoc/chapters/programming_interface.adoc
```diff
@@ -11777,35 +11777,36 @@ necessary [code]#device#.
...
+* [code]#ptr# has not previously been deallocated; and
+
+* There are no in-progress or enqueued <<command, commands>> using the memory
+  pointed to by [code]#ptr#.
+
+_Effects_: Causes the memory pointed to by [code]#ptr# to be deallocated.
+
+{note}Whether [code]#free# is blocking or non-blocking is unspecified.
+Applications should not rely on [code]#free# for synchronization, nor assume
+that [code]#free# cannot cause deadlocks.{endnote}
```
keryell, line 11803: Is this an advertisement for memory leaks?

Pennycook: The intent is to highlight that uses of `sycl::free` like the ones below may have surprising behavior.

```c++
int* m = sycl::malloc_shared<int>(...);
initialize(m);
q.parallel_for(..., [=]) {
  kernel(m);
};
sycl::free(m); // UB because developer didn't call wait() first
```

keryell: This code makes sense to be UB, since malloc/free are dangerous. :smiley_cat: 
But I thought blocking in the sense of the implementation itself, like `sycl::free()` implementation might not be wait-free or lock-free if a kernel is running on device.
But perhaps this is what you mean too.
Since SYCL is an asynchronous framework, I imagine that, if this is a problem, a SYCL implementation should use a scheduler to postpone the action up to when it can really happen.
With the problem that since `sycl::free` is not associated to a `queue`, a user cannot wait on it to really be sure the memory has been released. Waiting for the default context?

Pennycook: > This code makes sense to be UB, since malloc/free are dangerous. 😺 But I thought blocking in the sense of the implementation itself, like `sycl::free()` implementation might not be wait-free or lock-free if a kernel is running on device. But perhaps this is what you mean too.

Yes, exactly.  These things are all closely related: it's dangerous to call `sycl::free` on a pointer that a kernel is using, because the implementation might not wait for kernels to complete before actually freeing the memory. (This is the "don't rely on `sycl::free` for synchronization" case.)

> Since SYCL is an asynchronous framework, I imagine that, if this is a problem, a SYCL implementation should use a scheduler to postpone the action up to when it can really happen.

Right, and this behavior is still valid with the new wording.  Some backends might be incapable of freeing memory while the device is busy (e.g., executing a kernel): SYCL implementations layered on such backends might postpone the action until later, but might also just decide to wait (i.e., block) until the device is no longer busy.  (This is the "`sycl::free` might cause deadlocks if you're not careful" case.)

> With the problem that since `sycl::free` is not associated to a `queue`, a user cannot wait on it to really be sure the memory has been released. Waiting for the default context?

The way I think about this case is that returning from `sycl::free` doesn't guarantee that the memory has been released, it just guarantees that the implementation is aware that the user is done with the allocation. The user isn't allowed to use the pointer after calling `sycl::free`, and it's up to the implementation when it releases the memory back to the OS/device/whatever.

Although there's no way for the user to wait until the memory has been released, there are plenty of ways (e.g., `queue::wait()`, `event::wait()`) for them to wait until they're done with the memory before calling `sycl::free`.

tomdeakin (approved):

tomdeakin:
WG approved to merge for SYCL 2020

TApplencourt:
Just for posterity,  I finally understand the `blocking` problem.

It's not that `sycl::free` is blocking , aka waiting for the memory to be free before returning. The problem is that in CUDA, sycl::free(A,Q) will do a Q.wait() before freeing the memory (or even worse, I suppose, a context.wait()).

So it's not the "normal" blocking / async.  It's more a `dependency on previous usage`

PR #767, opened 2025-03-17
Fix reference to rule of five in common-byval.h

Pennycook:
Cherry pick #737 from main
(cherry picked from commit 65bb36484fac73e3a23b7bc899e78c7f275cc8e3)


Pennycook:
Merging because the WG already approved #737.

PR #784, opened 2025-04-04
Update to latest docker image

gmlueck:
Also update the process document to describe the process around Docker image versions.


gmlueck:
I compared the generated HTML and PDF files, and there were no differences relative to the previous version of the Docker image.

Merging this as editorial.

PR #796, opened 2025-04-07
Update process doc for publishing a revision

gmlueck:
Update the "Publishing an update revision" section of the process document.  This is verified against the process I used to publish revision 10.


gmlueck:
Merging as editorial

PR #798, opened 2025-04-09
Fix issues using buffer in examples

reble:
Some examples using buffers fail to compile: 
Buffer type is either incorrectly specified or incompatible range parameter is passed.
With SYCL 2020 we can improve examples, writing more compact code.
Aligning all other examples using buffers will require a follow up PR.

### DIFF
--- a/adoc/code/queueShortcuts.cpp
+++ b/adoc/code/queueShortcuts.cpp
@@ -7,7 +7,7 @@ queue myQueue;
 auto usmPtr = malloc_device<int>(1024, myQueue);  // USM pointer
 
 int* data = /* pointer to some data */;
-buffer<int, 1> buf{data, range<1>{1024}};
+buffer buf{data, {1024}};
 accessor acc{buf};  // Placeholder accessor
 
 // Queue shortcut for a kernel invocation
--- a/adoc/code/queueShortcuts.cpp
+++ b/adoc/code/queueShortcuts.cpp
@@ -7,7 +7,7 @@ queue myQueue;
 auto usmPtr = malloc_device<int>(1024, myQueue);  // USM pointer
 
 int* data = /* pointer to some data */;
-buffer buf{data, 1024};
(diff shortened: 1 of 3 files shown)


adoc/code/queueShortcuts.cpp
```diff
@@ -7,7 +7,7 @@ queue myQueue;
 auto usmPtr = malloc_device<int>(1024, myQueue);  // USM pointer
 
 int* data = /* pointer to some data */;
-buffer buf{data, 1024};
+buffer<int, 1> buf{data, range<1>{1024}};
```
gmlueck, line 10:
```suggestion
buffer buf{data, {1024}};
```

I think this is all you need.  (I checked on DPC++, and this does compile.)

The type `int` is deduced via CTAD from that `data` parameter and the dimension `1` is deduced from the range.  We just needed to add curlys around the `1024`, so that it list-initializes a `sycl::range`.

We added all the CTAD stuff to SYCL 2020 to make code less verbose, and I think our examples should make use of it.

reble: > We added all the CTAD stuff to SYCL 2020 to make code less verbose, and I think our examples should make use of it.

Thanks @gmlueck ! Should we align other examples to make it consistent?

gmlueck: I haven't looked at all the examples to see if they could be improved, but I agree that consistency is good.  If you know of examples that could be made better, feel free to add to this PR or creates others.

gmlueck (approved):

adoc/code/queueShortcuts.cpp
reble: I've fixed the obvious issues now with this PR. I'd suggest doing a follow up PR for alignment.
Alignment of examples is a lot and might require some discussion.

keryell (approved):
Thanks! Quite nicer.

gmlueck:
This just cleans up some of the examples, so merging as editorial.

PR #801, opened 2025-04-24
Specify return type for "queue_profiling"

gmlueck:
The synopsis for `sycl::info::device::queue_profiling` was missing the return type.  Fix this.

I'm certain that `bool` is correct because that is what we said in revision 9 of the SYCL-2020 spec.  It also makes sense because the definition of this info descriptor says it returns the same value as `device::has(aspect::queue_profiling)`, which is a `bool`.


TApplencourt (approved):

keryell (approved):
Oops. Thanks!

gmlueck:
Merging as editorial

PR #808, opened 2025-05-05
clarify_event_complete_synchronization

TApplencourt:
We didn't specify that from a synchronization perspective, an event returning `complete` is the same as a `wait`. 

It provides the same warranty (it will be more useful when we will have a flush, and we can have a `completed` event without ever calling wait).

I guess we need to port this thing to the `non-table` new way, but did the lazy hack...

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4917,6 +4917,8 @@ info::event_command_status::complete
 ----
    a@ Indicates that the command has finished running on the device.
       Attempting to wait on such an event will not block.
+      
+      Synchronization: Equivalent to event::wait().
 
 |====
 


adoc/chapters/programming_interface.adoc
```diff
@@ -4917,6 +4917,8 @@ info::event_command_status::complete
 ----
    a@ Indicates that the command has finished running on the device.
       Attempting to wait on such an event will not block.
+      
+      Synchronization: Equivalent to event::wait().
```
Pennycook, line 4921:
```suggestion
      Synchronization: Equivalent to [code]#event::wait()#.
```

The wording looks fine to me, but I'm on the fence about whether it should be specified here, or whether we need to make it clear somehow that it is the call to `get_info` which has the synchronization behavior, and only if the return value is `complete`?

Perhaps the easiest thing to do is to make the change as you suggested, and then rewrite this section later (removing the table, etc).

TApplencourt: Agree.  The "no-table" allows for more flexibility.
At least with the PR it's a little better than before (we try to explain), and in a new PR we can make it crystal clear :')

We can see what other people think in the following WG call 🤞🏽

adoc/chapters/programming_interface.adoc
```diff
@@ -4917,6 +4917,8 @@ info::event_command_status::complete
 ----
    a@ Indicates that the command has finished running on the device.
       Attempting to wait on such an event will not block.
+      
+      Synchronization: Equivalent to [code]#event::wait()#.
```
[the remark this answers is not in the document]
TApplencourt: Good idea!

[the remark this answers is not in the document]
TApplencourt: Ok, the rebase didn't work, and I have to remove the line `info::device::parent_device` . If that confuse you too much, I can do some `cherry-pick / rebase` magic.

adoc/chapters/programming_interface.adoc
```diff
@@ -4917,6 +4916,9 @@ info::event_command_status::complete
 ----
    a@ Indicates that the command has finished running on the device.
       Attempting to wait on such an event will not block.
+      
+      Synchronization: When event::get_info returns this value,
+      the synchronization is equivalent to event::wait.
```
gmlueck, line 4921: Please add `[code]#...#` decoration.

adoc/chapters/programming_interface.adoc
```diff
@@ -3155,7 +3155,6 @@ backend of this device is not OpenCL.
 [source,role=synopsis,id=api:info-device-parent-device]
 ----
 namespace sycl::info::device {
-info::device::parent_device
```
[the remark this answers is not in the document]
TApplencourt: Yes, a rebase one.

adoc/chapters/programming_interface.adoc
TApplencourt: Do you want `code` or `api`?

gmlueck: `api` only works for the classes that have been reformatted into the new style (because there needs to be a matching `apidef` whenever you use `api`).  The `event` class hasn't been reformatted yet, so you need to use `code` here.

TApplencourt:
Used `[code]` and fixed bad rebase. Should be good now.

gmlueck (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -4917,6 +4917,9 @@ info::event_command_status::complete
 ----
    a@ Indicates that the command has finished running on the device.
       Attempting to wait on such an event will not block.
+      
+      Synchronization: When [code]#event::get_info# returns this value,
+      the synchronization is equivalent to [code]#event::wait#.
```
gmlueck, line 4922:
```suggestion
Synchronization: When [code]#event::get_info# returns this value,
the synchronization is equivalent to [code]#event::wait#.
```

Sorry, one more comment ... The indentation needs to be fixed like this, otherwise it gets rendered as a code block.

TApplencourt: Oh, I noticed that `reflow.py` didn't seem to work with the table. I should have double-checked the PDF. Thanks a lot!

tomdeakin (approved):

tomdeakin:
WG approved to merge.

PR #809, opened 2025-05-05
[KHT] Queue flush

TApplencourt:
To get the ball rolling : ) 
I'm sure we will need to tweak the description and notes.

Maybe also a "why this is useful / what problem it solve" section?

### DIFF
--- a/adoc/extensions/index.adoc
+++ b/adoc/extensions/index.adoc
@@ -11,3 +11,4 @@ specification, but their design is subject to change.
 // include::sycl_khr_extension_name.adoc[leveloffset=2]
 
 include::sycl_khr_default_context.adoc[leveloffset=2]
+include::sycl_khr_queue_flush.adoc[leveloffset=2]
--- a/adoc/extensions/sycl_khr_queue_flush.adoc
+++ b/adoc/extensions/sycl_khr_queue_flush.adoc
@@ -1,0 +1,64 @@
+[[sec:khr-queue-flush]]
+= sycl_khr_queue_flush
+
+This extension allows developers to query the queue's emptiness, meaning if all
+commands submitted to a queue have been completed.
+
+[[sec:khr-queue-flush-dependencies]]
+== Dependencies
+
+This extension has no dependencies on other extensions.
+
+[[sec:khr-queue-flush-feature-test]]
+== Feature test macro
+
+An implementation supporting this extension must predefine the macro
+[code]#SYCL_KHR_QUEUE_FLUSH# to one of the values defined in the table below.
+
+[%header,cols="1,5"]
+|===
+|Value
+|Description
+
+|1
+|Initial version of this extension.
+|===
+
+[[sec:khr-queue-flush-funct]]
+== New Queue Command to Force Submission
+
+This extension adds the following function to the [code]#sycl::queue# class,
+which allows you to ensure forward progress without the need to call
+[code]]#sycl::queue::wait()# '''
+
+.[apidef]#queue::khr_flush#
+[source,role=synopsis,id=api:queue-khr-flush]
+----
+void khr_flush() const
+----
+
+_Effects:_ Behave as if another host thread submitted a
+[code]#sycl::queue::wait#.
+
+{note} This ensures that the event of at least one command submitted to the
+queue will reach the [code]#info::event_command_status::running#.
+{endnote} '''
+
+[[sec:khr-queue-flush-example]]
+== Example
+
+The example below demonstrates the usage of this extension.
+
+[source,,linenums]
+----
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue Q;
+  auto e = Q.single_task([] {});
+  Q.khr_flush(); // UB without it
+  while (e.get_info<sycl::info::event::command_execution_status>() !=
+         sycl::info::event_command_status::complete) {
+  };
+}
+----


Pennycook (commented):
This looks really good.  Just a few nit-picky suggestions...

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,64 @@
...
+.[apidef]#queue::khr_flush#
+[source,role=synopsis,id=api:queue-khr-flush]
+----
+void khr_flush() const
+----
+
+_Effects:_ Behave as if another host thread submitted a
+[code]#sycl::queue::wait#.
+
+{note} This ensures that the event of at least one command submitted to the
```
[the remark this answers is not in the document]
TApplencourt: "This is the SYCL Oracle of Delphi: Be careful, this function will depend on your functions' action." :D

TApplencourt:
Oh, by the way, implementation-wise, we should be good.
```
namespace sycl {
  class queue :  {
    public:
      void khr_flush() {
        (void) std::async(std::launch::async, [&] {this->wait();});
      }
  };
}
```

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+|1
+|Initial version of this extension.
+|===
+
+[[sec:khr-queue-flush-funct]]
+== New Queue Command to Force Submission
+
+This extension adds the following function to the [code]#sycl::queue# class,
+which allows you to ensure forward progress without the need to call
```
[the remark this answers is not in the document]
TApplencourt: Used that!

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,64 @@
+[[sec:khr-queue-flush]]
+= sycl_khr_queue_flush
+
+This extension allows developers to query the queue's emptiness, meaning if all
+commands submitted to a queue have been completed.
```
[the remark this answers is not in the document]
TApplencourt: Because we have only one function, I removed the function description and just put it on top. It avoids some "code" duplication.

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,67 @@
...
+----
+void khr_flush() const
+----
+
+_Effects:_ Ensure that device code is able to make forward progress, as if an
+unspecified host thread providing concurrent forward progress guarantees called
+[code]#sycl::queue::wait#.
+
+{note} This ensures that the event of at least one command previously submitted
+to the queue will enter the [code]#info::event_command_status::running# state.
```
gmlueck, line 45: The part about the "event" seems superfluous.  The `khr_flush` call will ensure that a command runs even if it has no associated event.  Could we say this instead:

> Assuming the queue has pending commands, calling this function ensures that at least one command from the queue starts running within some finite period of time.

I wonder if this is true, though.  If there is another queue on the same device, and there is a deadlocked command on that other queue, I think there is no guarantee that commands on this queue will start running.

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,67 @@
...
+|1
+|Initial version of this extension.
+|===
+
+[[sec:khr-queue-flush-funct]]
+== New Queue Command to Force Submission
+
+This extension adds the following function to the [code]#sycl::queue# class.
+
+'''
```
gmlueck, line 32: I'd prefer that we keep a consistent style between the KHRs.  Our existing sycl_khr_default_context KHR also adds only one member function, yet we still showed the synopsis for the class.

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+.[apidef]#queue::khr_flush#
+[source,role=synopsis,id=api:queue-khr-flush]
+----
+void khr_flush() const
+----
+
+_Effects:_ Ensure that device code is able to make forward progress, as if an
+unspecified host thread providing concurrent forward progress guarantees called
+[code]#sycl::queue::wait#.
```
gmlueck, line 42:
```suggestion
[api]#queue::wait#.
```

If you do this, it will give you a link to the definition.

adoc/extensions/sycl_khr_queue_flush.adoc
TApplencourt: My train of thought was to map it to something observable. The "start running" was a little too vague for my taste, but we use the same in the running description. So yes, cleaner! Thanks.

>I wonder if this is true, though. If there is another queue on the same device, and there is a deadlocked command on that other queue, I think there is no guarantee that commands on this queue will start running.

Yes, I considered using "in non-pathological case" / "non-deadlocking case", but that seemed too imprecise. "Will allow comment to execute if possible" is simple, but super impressive (what does "possible" mean?). This is the nice thing with the " as if wait ...".  The `wait` does all the heavy lifting. 

Ready of forward progress quarenty, we said:

> When a host thread blocks on the completion of a command previously submitted to a SYCL queue (for example, via the sycl::queue::wait function), it blocks with forward progress guarantee delegation.

We have ` forward progress guarantee delegation` for the waiting thread, so I think deadlocking is UB. Is this true @Pennycook ? In this case, it solve the problem of the node :)

adoc/extensions/sycl_khr_queue_flush.adoc
TApplencourt: Ok, then will copy/paste the sentence. Please let not change it anymore then :D

TApplencourt: Good point. Fixed with 32e3d1c

adoc/extensions/sycl_khr_queue_flush.adoc
Pennycook: > We have  forward progress guarantee delegation for the waiting thread, so I think deadlocking is UB. Is this true @Pennycook ? In this case, it solves the problem of the node :)

My understanding is that deadlocking is _not_ UB.

It comes down to how "progress" is defined in [[intro.progress]](https://eel.is/c++draft/intro.progress).  I won't copy-paste the whole thing here, but I think the key points are:
- "Making progress" just guarantees that the thread is executing _something_;
- "Something" can be an atomic operation (e.g., reading a condition in an infinite loop); and
- The latest draft adds new wording to explicitly allow "trivial infinite loops".

My interpretation of this is that, if we launch a kernel that just sits somewhere spinning, the thread of execution running that kernel _is_ making forward progress.

---

To @gmlueck's point, I agree that we need to tighten some wording here.  `flush` cannot provide a stronger guarantee than `wait`, and somebody might interpret the non-normative note that way.

How about:

> Calling this function is an implementation-independent way to ensure that pending commands can start executing on the device without blocking the calling thread. Exactly when pending commands start executing is unspecified, as it depends on the scheduling behavior(s) of implementations and individual devices. However, if a call to `wait` would have been unblocked in a finite amount of time, a call to `flush` guarantees that all pending commands will execute in a finite amount of time.

The specification probably isn't clear enough about when calls to `wait` are guaranteed to be unblocked in a finite amount of time, but lots of applications are depending on that behavior today -- so this wording is an attempt to effectively say that if a SYCL program is "valid", inserting `flush` calls will do what the developer expects.

---

One aside that occurred to me while I was writing this up... I don't think there's anything about this new `flush` behavior that would guarantee that two commands (e.g., from different queues) ever execute "at the same time".  Even if two different host threads block on a call to `wait`, I think it is valid for a SYCL implementation to have something like a global lock in place and only ever execute one SYCL kernel at a time.

This is still compatible with "blocking with forward progress guarantee delegation", but it might not be what people expect.  Basically, the thread which calls `wait` inside `flush` could enter a loop that spins and tries to acquire permission to execute on the device, and it will eventually succeed even if there are other threads out there doing the same thing (which is the forward progress guarantee).

I don't think we're trying to solve this problem with `flush`, but for posterity: I don't think `flush` is sufficient to solve this problem, and if we do want to solve this problem, we'll need to provide stronger scheduling guarantees.

TApplencourt: >   o this wording is an attempt to effectively say that if a SYCL program is "valid", inserting flush calls will do what the developer expects.
>The specification probably isn't clear enough about when calls to wait are guaranteed to be unblocked in a finite amount of time

Do we really think we need to said anything? Wait, just wait for all the commands to finish. If they never finish, too bad for you. 

> so this wording is an attempt to effectively say that if a SYCL program is "valid", inserting flush calls will do what the developer expects.

💯  What about "If a call to wait would have finished, then flush will ensure that the commands will start running."? 
It's really hard to define flush without using "wait".  So maybe the note can just clarify/rephrase the "thread calling wait".  Nobody, "in reality," will care about the deadlock scenario. Of course, if they have broken code that deadlocks / "not-finish-in-a-finite-time-or-event-infinite-one" flushing will not magically make their code work.

The note was @gmlueck  idea, so at the end, I'm ok letting greg having the last word on this. We can always clarify the letter if users are confused.

----

>One aside that occurred to me while I was writing this up... I don't think there's anything about this new flush behavior that would guarantee that two commands (e.g., from different queues) ever execute "at the same time". Even if two different host threads block on a call to wait, I think it is valid for a SYCL implementation to have something like a global lock in place and only ever execute one SYCL kernel at a time.

IMO We don't need to do anything :). The same problem with 2 kernels without dependency in a Out of Order queue. Executing them "at the same time" is purely an optimization detail. Relying on that, is UB.

gmlueck: I like @Pennycook's proposal for the non-normative note.  I'd suggest just a small tweak:

> Calling this function is an implementation-independent way to ensure that pending commands can start executing on the device without blocking the calling thread. Exactly when pending commands start executing is unspecified, as it depends on the scheduling behavior(s) of implementations and individual devices. However, if a call to `queue::wait` would have been unblocked in a finite amount of time, a call to `queue::khr_flush` guarantees that pending commands on that queue will execute in a finite amount of time.

gmlueck: > I don't think we're trying to solve this problem with `flush`, but for posterity: I don't think `flush` is sufficient to solve this problem, and if we do want to solve this problem, we'll need to provide stronger scheduling guarantees.

Are you saying that users may expect that commands will run in parallel with each other?  I think we agree that the spec doesn't guarantee this.  It would valid for an implementation to execute all commands consecutively.  Why do you think the `khr_flush` command makes it more likely for customers to think that commands will run in parallel?  It seems like people might have this mistaken impression even when calling `wait`.

Pennycook: > Are you saying that users may expect that commands will run in parallel with each other? I think we agree that the spec doesn't guarantee this. It would valid for an implementation to execute all commands consecutively. Why do you think the `khr_flush` command makes it more likely for customers to think that commands will run in parallel? It seems like people might have this mistaken impression even when calling `wait`.

Consider a case like this:
```c++
sycl::queue q1;
sycl::queue q2;

q1.submit(..., KernelA());
q2.submit(..., KernelB());

q1.flush(); // lazy interpretation: "start running Kernel A"
q2.flush(); // lazy interpretation: "start running Kernel B"
```

I think our wording is clear that the only guarantee here is that `KernelA` and `KernelB` will finish in finite time.  But we have to be careful in training materials, etc, that we don't encourage mental model shortcuts that would lead to an interpretation like the above.

TApplencourt: Merged new notes with c29a1a8c765219b27c0ea05199f9a345e2a0c869

TApplencourt:
Should we add a note saying  "Sorry, this extension will break conformant single-threaded implementation. Fortunately, none exist."

PeterTh:
My example needs two *user* threads, and the SYCL implementation cannot choose to run the host task on the same thread it was submitted on.

It can, however, decide to execute the host task on the thread the user calls wait on, which means it would still not have to *spawn* a thread, which does mean that `flush` would impose that as a new requirement on the SYCL implementation. (Which I don't think is a problem, but yes, it is a change)

Are you specifically looking for an example with a single-threaded application **and** a single-threaded implementation?

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+{note} Calling this function is an implementation-independent way to ensure that
+pending commands can start executing on the device without blocking the calling
+thread.
+Exactly when pending commands start executing is unspecified, as it depends on
+the scheduling behavior(s) of implementations and individual devices.
+However, if a call to [api]#queue::wait# would have been unblocked in a finite
+amount of time, a call to [api]#queue::khr_flush# guarantees that pending
+commands on that queue will execute in a finite amount of time.
+{endote}
```
gmlueck, line 64:
```suggestion
{endnote}
```

TApplencourt:
>My example needs two user threads, and the SYCL implementation cannot choose to run the host task on the same thread it was submitted on.

Oh I see, good point! 

>Are you specifically looking for an example with a single-threaded application and a single-threaded implementation?

I think so, because if the user code can use a thread, I don't see why the sycl runtime cannot... But it's a good restriction.

The concern is that `flush` adds a new restriction to implementers. I was trying to find a counter-example to say that the restriction already existed. But maybe it's a vain approach :)

Pennycook:
So... I thought about this a little bit more, and I realized that there is some precedent in C++ that we could try to align with here.  C++ distinguishes between "freestanding" and "hosted" implementations, where a freestanding implementation is one where:

- The entire C++ language is supported;
- A subset of C++ library functions are guaranteed to be supported;
- Support for additional C++ library functions is implementation-defined; and
- It is implementation-defined whether there can be more than one thread of execution (see [[intro.multithread.general]](https://eel.is/c++draft/intro.multithread.general#1)).

A developer can use the feature-test macros defined in `<version>` (see [[version.syn]](https://eel.is/c++draft/version.syn#:macro,feature-test)) to check whether specific library functionality is available.

---

So, here's a corresponding proposal for this KHR / SYCL-Next that I think might satisfy everybody:

- Define a new "freestanding" SYCL feature set;
- Say that `queue::flush()` is not part of "freestanding", but its availability can be tested with `__sycl_lib_queue_flush`;
- Describe `queue::flush()` as in this KHR (i.e., if it is available, it provides the required guarantees).

This would ensure that:
- Any implementation conformant to the full SYCL specification would support `flush()`.
- Programs targeting the full SYCL specification could reliably use `flush()` without checking aspects or handling exceptions.
- Programs that care about targeting freestanding implementations could check if `flush()` is supported _at compile time_.

We could pick a different name if anybody thinks re-use of "freestanding" would be confusing, but I don't think that would change anything else about the proposal.

etomzak:
> So, here's a corresponding proposal for this KHR / SYCL-Next that I think might satisfy everybody:
> 
>     * Define a new "freestanding" SYCL feature set;
> 
>     * Say that `queue::flush()` is not part of "freestanding", but its availability can be tested with `__sycl_lib_queue_flush`;
> 
>     * Describe `queue::flush()` as in this KHR (i.e., if it is available, it provides the required guarantees).

I wonder if what you're describing as "freestanding" is actually SYCL SC. Or maybe it's one of the things that SYCL SC should aspire to be.

That said, I don't think the idea of a "freestanding" subset of (Base) SYCL is going to work in practice. My big concern about this KHR is somewhat philosophical: It's a tiny addition to the spec, but it has surprisingly (and disproportionately) large impacts on the interpretation of the SYCL spec throughout. One of the value propositions of SYCL is that the user doesn't know or need to know the details of the scheduling policy (which is good for portability, simplicity, teachability, etc.) --- implementations are allowed to be eager, lazy, or do smart scheduling things (because implementations have visibility into the configuration and state of the underlying hardware that the user doesn't have or need to have). This KHR blurs the separation of concerns between the user and the SYCL implementation, making it less clear who's responsible for triggering kernel execution.

Introducing "freestanding" SYCL in support of this KHR would be cranking the surprising and disproportionate knock-on effects of this KHR to 11.

Also, I have the impression that freestanding C++ gets barely any attention and is basically unusable. I imagine the situation would be even worse for a "freestanding" SYCL. IMO, "freestanding SYCL" would essentially be an equivalent to `/dev/null` where difficult discussions could be piped.

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+.[apidef]#queue::khr_flush#
+[source,role=synopsis,id=api:queue-khr-flush]
+----
+void khr_flush() const
+----
+
+_Effects:_ Ensure that device code is able to make forward progress, as if an
+unspecified host thread providing concurrent forward progress guarantees called
+[api]#queue::wait#.
```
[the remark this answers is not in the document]
Pennycook: My position on this is no, because we don't say that the function blocks until something else happens.

But I concede that, like @gmlueck said yesterday, it is difficult for a developer to tell whether something is blocking or just taking a very long time.  So it might not be possible for us to actually forbid what you described.

An implementation that ran for a little bit and then returned to the host would have to ensure that it never blocked forever, though, which I maintain would be quite hard to implement in the general case.  Certainly much harder than just using two threads.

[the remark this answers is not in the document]
etomzak: Either way, I think the intent needs to be clarified here.

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+[source,,linenums]
+----
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue Q;
+  auto e = Q.single_task([] {});
+  Q.khr_flush(); // UB without it
+  while (e.get_info<sycl::info::event::command_execution_status>() !=
```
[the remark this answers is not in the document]
TApplencourt: I don't understand your point.


>I'm not at all convinced by this use case. This code is just the user doing a bad job at re-implementing event::wait().
What it's a bad job reimplmentation?

>If users are writing this code, then the solution is education and/or some tooling to flag up the problematic use case. If the code is coming from SYCLomatic, then that sounds like a SYCLomatic bug.

This is what people do in CUDA. I'm here to blame them. 
`while (cudaStreamQuery(d) != cudaSuccess) { usleep(50000) };` This is a super common pattern.


Are you arguing against flush, against the example, or against cuda developer?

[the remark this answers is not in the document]
TApplencourt: No. FLush is not a performance hint. It change the behavior and the semantic of the code

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
TApplencourt: > Are we OK with an implementation of khr_flush() that blocks until some/all work on the queue has been executed?

Yes, if it's spawn by another thread. Not if it's by the main thread

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,85 @@
...
+The example below demonstrates the usage of this extension.
+
+[source,,linenums]
+----
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue Q;
+  auto e = Q.single_task([] {});
+  Q.khr_flush(); // UB without it
```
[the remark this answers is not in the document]
TApplencourt: No. It's UB. It May or may not be infint loop, it depend on implementation without it.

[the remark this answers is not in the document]
TApplencourt: Maybe my talk https://www.overleaf.com/read/rntxcrnyphct#12ccfb (the pdf uploaded previously on cosway) will help clarify a few thing

TApplencourt:
Can we at least merge a version V1 of this? 

If people don't want to implement it, they can don't define the macro for now. Not perfect, but doesn't seem to hurt.
In a few months/years, when we want to merge into the main spec, we may revisit (or create a v2 adding new query / set . whatever) 

The `queue::khr_empty`, is mostly only useful if we get this one done too.

It just wants to balance the "theoretical spec issue" versus the "practical this-is-what-the-user-needs" concern.

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
TApplencourt: I don't understand which clarification you want? Can you propose a fix?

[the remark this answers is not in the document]
etomzak: > Yes, if it's spawn by another thread. Not if it's by the main thread

> I don't understand which clarification you want? Can you propose a fix?

From my reading, the current normative text is compatible with `khr_flush()` blocking on kernel executions. @TApplencourt sounds OK with this, but @Pennycook is opposed. Could the proponents of this KHR decide what they want and then put that in the normative text?

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
etomzak: What's the undefined behavior in this specific example if not the infinite loop? I agree that the more complex examples in your slide deck potentially have other instances of UB.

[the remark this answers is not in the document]
TApplencourt: You seem to imply that you should _must_ hit the Infinite while loop without the flush. I just said that it's not required. The behavior is undefined. You may or you may not. It's UB. 

We can say "The code is UB without the flush. The while-loop exit may never exits"

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
TApplencourt: I think we have a misunderstanding. 

>  _Effects:_ Ensure that device code is able to make forward progress, as if an
unspecified host thread providing concurrent forward progress guarantees called
[api]#queue::wait#.

It is correct, both John and I agree.  Do you think it's incorrect? 

>      From my reading, the current normative text is compatible with khr_flush() blocking on kernel executions. 

No. Because if it was the case, this code will be UB:
```
std::promise<void> promise;
q.submit([&](sycl::handler& cgh) {
    cgh.host_task([&] { promise.get_future().wait(); });
});
q.khr_flush();
promise.set_value();
q.wait();
```
It needs to be `unspecified host thread providing concurrent forward progress guarantees `. If it's the host thread that blocks, it doesn't work.


The interpretation, subject to one thread implementation, is debatable. But I don't see how it impact this sentence.

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
etomzak: > What it's a bad job reimplmentation of wait?

As we've discussed, the user wants the application to spin there until the event enters a `complete` state. SYCL already provides a facility to achieve the equivalent of this: `event::wait()`. Instead of using the facility provided by SYCL, the user is trying to create their own equivalent. The user's code turns out not to work as intended, because the user doesn't understand SYCL semantics. I call this a failed (= bad) attempt to re-implement what SYCL already provides (`event::wait()`).

> Are you arguing against flush, against the example, or against cuda developer?

Yes to all(?)

Re: flush: As it currently stands, IMO flush causes more problems than it solves. What's being asked for, IIUC, is a way to compel a SYCL runtime to be eager if it's lazy by default. Some sort of device or queue property would be a clearer way to achieve that, I think. Or is the desire to have a queue that's sometimes eager and sometimes lazy?

Re: the example: Currently, it looks like the motivating example for this extension is to allow for the creation of a spin lock. I don't find it particularly compelling, since spin locks are often considered bad practice. A different motivating example might be better for communicating (to me and others) why this extension is needed.

Re: CUDA: I don't think there's any inherent need for SYCL to support any given CUDA idiom. The SYCL WG can choose to support some specific CUDA idioms if it wants to, but "it's a CUDA idiom" shouldn't guarantee inclusion in SYCL. Someone might want Vulkan-style pipeline caches or Java-style garbage collection in SYCL, but they'll probably not get those things, because those things don't mesh well with the SYCL paradigm. SYCL is SYCL; SYCL is not CUDA, and part of my argument here is that flush is awkward in the context of the rest of the SYCL paradigm.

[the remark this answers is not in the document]
TApplencourt: Ok, then good thing it's a KHR and not part of the main spec :)

TApplencourt:
Added KHR tests (https://github.com/KhronosGroup/SYCL-CTS/pull/1094)

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
etomzak: I think I see what you mean.

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
etomzak: > We can say "The code is UB without the flush. The while-loop exit may never exits" if it's clearer.

That's sort-of what I suggested. Some wording like that would be helpful.

Can we just be clear about terminology, though. AIUI, in a current implementation, whether `e.get_info<sycl::info::event::command_execution_status>()` returns `complete` in this specific example is *implementation-defined* (not undefined) behavior. If this implementation-defined behavior causes an infinite loop, then and only then is it undefined behavior because infinite loops are undefined behavior in C++.

(I'm not sure if P2809 in C++26 affects this interpretation, but that's beside the point.)

Or is there another instance of undefined behavior in this example other than the infinite loop that I'm missing?

adoc/extensions/sycl_khr_queue_flush.adoc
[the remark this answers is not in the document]
etomzak: > Ok, then good thing it's a KHR and not part of the main spec :)

I disagree. A KHR should represent a high bar (as opposed to a vendor extension or an implementation-defined feature). Currently, I don't think this PR has a sufficiently strong rationale behind it to be a candidate for a KHR.

I spent a fair bit of time thinking about this extension, so I'd like to propose what I think is a better problem statement and a better solution.

**The problem**, AIUI, is actually with the interaction between implementation-defined kernel dispatch behavior and `aspect::usm_atomic_shared_allocations` --- SYCL allows for a spectrum of dispatch behaviors ranging from very eager to very lazy. However, some lazy behaviors, while otherwise allowed by the spec, can cause deadlocks if (and only if) there are host-device shared atomics.

The appropriate **solution**, IMO, is that implementations that support `aspect::usm_atomic_shared_allocations` must implement kernel dispatch behavior that is compatible with host-device shared atomics. It's as simple as that. If an implementation advertises that it supports `aspect::usm_atomic_shared_allocations` but implements kernel dispatch in a way that deadlocks when the user uses shared atomics, then this is an implementation bug.

IMO, this solution is already implied by the spec. However, it might be appropriate to introduce a non-normative implementation note into the core spec saying something like, "If you support `aspect::usm_atomic_shared_allocations` then you must ensure that you have sufficient forward progress guarantees for your kernels to make shared atomics actully work." (But phrased better.)

If I've understood the problem correctly, then I don't think that this (or any) extension is an appropriate solution for a few different reasons:

* `aspect::usm_atomic_shared_allocations` is in the core spec, so an extension shouldn't be required to make shared atomics work correctly.
* Shared atomics should just work. Users shouldn't be required to remember to call `flush` to make them work.
* If I'm wrong about my solution already being implied by the spec, then we have a core spec bug. An extension doesn't fix the core spec. We should fix the core spec instead.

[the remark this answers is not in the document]
TApplencourt: > The problem, AIUI, is actually with the interaction between implementation-defined kernel dispatch behavior and aspect::usm_atomic_shared_allocations

This is not the problem my user asked me to solve. Please open a new KHR to solve this issue.
Can we focus back the discussion on the flush (the property that both OpenCL and CUDA provide)?

nliber:
To me, this KHR is a simple addition (both to specify and to implement) which helps folks coming/porting from CUDA.  It is not a commitment to support every CUDA idiom.  While the user can already do the equivalent of `flush`, this gives them a documented, supported and more convenient way of doing so.

I’m not sure where “the impression that freestanding C++ gets barely any attention and is basically unusable” is coming from.  Since 2017 WG21 has been spending time on improving freestanding, and there have been changes to C++20, C++23 and the upcoming C++26 in that regard.

Sidebar:  I think most of what is freestanding in C++ could be adapted to SYCL fairly easily, but I haven’t quite flushed it out in my head to turn it into a proposal. What I’m wrestling with: the guarantee that people seem to want is that the implementation doesn’t allocate memory nor throw an exception. Specifications don’t tend to put those kinds of constraints on implementations, since they talk more about the interface boundary rather than what happens behind the interface.  Implementers are smart and know this is what their customers expect as well, but that isn't the same as a guarantee.  But I digress…

As for what a KHR should represent, as I understand it from our discussions, it is an extension that we are hoping multiple vendors adopt, and after sufficient positive experience we add it to the SYCL specification.  This KHR meets that bar.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -32,6 +32,10 @@ below.
 This extension adds the following function to the [code]#sycl::queue# class,
 which provides information about the emptiness of the queue.
 
+{note} This khr is more usefull when used in conjuction with
```
gmlueck, line 35:
```suggestion
{note} This khr is more useful when used in conjunction with
```

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -32,6 +32,10 @@ below.
...
+[api]#queue::khr_flush()#.
```
gmlueck, line 36:
```suggestion
[api]#queue::khr_flush#.
```

This must exactly match the `[apidef]`.  Also, I'm moving to a style where we refer to function names without the `()`.  This is the same style as cppreference.

I'm surprised this does not result in a build error, I thought the scripting used to generate an error if an `[api]` did not have a matching `[apidef]`.  I wonder if that error checking logic got broken somehow.

TApplencourt: Look like the implementation is here`api_xrefs/extension.rb`
And indeed, without looking to closely I was expecting
```
        # Make sure there was a definition for this API.  Diagnose an error if
        # not.
        if not api_to_id.key?(api)
          logger.error "[api]##{api}#: Not defined"
 ```
 To fail

File: adoc/extensions/index.adoc [hunk 13]
```diff
@@ -11,4 +11,5 @@ specification, but their design is subject to change.
 // include::sycl_khr_extension_name.adoc[leveloffset=2]
 
 include::sycl_khr_default_context.adoc[leveloffset=2]
+include::sycl_khr_queue_flush.adoc[leveloffset=2]
```
gmlueck, line 14: Please move this to the end.  New extensions should be added after ones that are already published, so we don't change the section number of the already-published extensions.

adoc/extensions/sycl_khr_queue_empty_query.adoc
gmlueck: > But I think it's because `()` doesn't match the regex (`[\w:]+`), so they were not considered API at all. I think that `khr_flush2` would have triggered an error

Indeed.  Looks like this has been a bug since I first added this extension.  I'll try to take a look later today.

adoc/extensions/index.adoc
TApplencourt: Fixed with d08148e

TApplencourt:
What about
```
In the absence of cross-queue dependencies, ensures that device code is able to make forward progress, as if an
unspecified host thread providing concurrent forward progress guarantees called
[api]#queue::wait#. Otherwise, behavior is implementation-defined.

{note} Calling this function is an implementation-independent way to ensure that
pending commands issued to the queue can start executing on the device without blocking the calling
thread.

Exactly when pending commands start executing is unspecified, as it depends on
the scheduling behavior(s) of implementations and individual devices.

In the absence of cross-queue dependencies,
if a call to [api]#queue::wait# would have been unblocked in a finite
amount of time, a call to [api]#queue::khr_flush# guarantees that pending
commands on that queue will execute in a finite amount of time.

When queues have interdependencies, users may need to call [api]#queue::khr_flush# on each involved queue to ensure forward progress.

For greedy implementation, [api]#queue::khr_flush# may be implemented as a no-op.
```

gmlueck:
It's interesting to note that SYCL's `queue::wait` has a stronger guarantee than OpenCL's `clFinish`.  Therefore, a SYCL implementation on top of OpenCL would need to call `clFlush` on other command lists when there are cross-queue dependencies.  For example:

```
e1 = q1.submit();  // Command C1 is Queued on underlying OpenCL
e2 = q2.submit();  // Command C2 is Queued on underlying OpenCL
q2.wait();         // Command C2 becomes Submitted on underlying OpenCL
                   // Command C1 remains Queued on underlying OpenCL unless SYCL implementation calls clFlush on "q1"
```

TApplencourt:
You have multiple algorithms: You can topologically sort all the dependencies and call `finish` on the correct order -- or maybe just execute finish on the user-submit order) . Or keep the queue, and call flush in any order, as you noticed. Less work.

But yes, it's the @Pennycook point. `sycl::wait` has far more guarantee than `clFinish`. And now `flush` has less than `wait`. 

But I'm ok with that if that help implementer (even if for OpenCL just keeping a list of used `queue`  doesn't seem the end of the world, as you already need to do that for `queue::wait`,  and for CUDA / L0 it's a no-op)

Pennycook:
> But yes, it's the @Pennycook point. `sycl::wait` has far more guarantee than `clFinish`. And now `flush` has less than `wait`. 
> 
> But I'm ok with that if that help implementer (even if for OpenCL just keeping a list of used `queue`  doesn't seem the end of the world, as you already need to do that for `queue::wait`;  and for CUDA / L0 it's a no-op)

I still think it's really weird to let OpenCL dictate how flush works when we didn't consider OpenCL's behaviour when defining wait.  Like you said, implementers *already* need to pay attention to this discrepancy, so I'm not convinced that adding restrictions to flush actually makes things that much easier for implementers.

I also realised that there might be a simpler implementation possible here.  Instead of back-tracking or keeping track of all queues, couldn't an implementation just call `clFlush` immediately when a cross-queue dependency is detected?  That would ensure that all commands in the first queue become Submitted, and then future flush/wait calls on the dependent queue would do the right thing.

TApplencourt:
>I also realised that there might be a simpler implementation possible here. Instead of back-tracking or keeping track of all queues, couldn't an implementation just call clFlush immediately when a cross-queue dependency is detected?

Yes, it will transforms a lazy implementation into a greedy one. Hence, `sycl::flush` will be a no-op. Lot of possible implementation indeed.

TApplencourt:
Oh, I never pushed my old change with the `multiple-queue` restriction. So we are good to go! Just waiting for implementers :)

etomzak:
I'd like to try once more to understand what specific problem this extension is intended to solve. Is this extension intended to give users the ability to load-balance across `sycl::queue`s?

adoc/extensions/sycl_khr_queue_flush.adoc
```diff
@@ -0,0 +1,87 @@
...
+----
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue Q;
+  auto e = Q.single_task([] {});
+  Q.khr_flush();
+  // Note: there's no wait.
+  // Without flushing, whether the event is marked as complete is implementation-defined.
+  while (e.get_info<sycl::info::event::command_execution_status>() !=
```
etomzak, line 83: I've been wondering if this code is guaranteed to work as intended even if `khr_flush()` were implemented as specified. I've been reading the spec, and I can't tell if spinning on a `sycl::event` status is something that's intended to work. Is a SYCL implementation required to update the status of a `sycl::event` asynchronously to a user thread? The event status definitely gets updated on `event::wait()` because `wait()` is a synchronization point. What happens when you use `krh_flush()` and get rid of that synchronization point?

I imagine that current implementations will probably just check a backend event every time `event::get_info()` is called, so this code should work. I'm less sure if implementations are actually required to do that. For some implementations and backends, it could be infeasible.

TApplencourt: Oh, I see. Good point.

As now defined, `flush` behaves as if `wait` has been called in another thread. So the synchronization point will exist in the future. You are just spin-locking to wait for the synchronization point to be reached. When reached, it will be visible to all threads, hence no deadlock and forward-progress guarantee.

Pennycook: In case it helps: I think this is precisely what the clarification in https://github.com/KhronosGroup/SYCL-Docs/pull/808 is for.

Spinning on an `event` status doesn't force a command to make forward progress, but reading `event::get_info()` _is_ a synchronization point if it returns `complete`.  Without `flush`, there's no guarantee that an `event` will ever become `complete`; spinning on an `event` may work on some implementations, but there's no guarantee.

Adding `flush` makes spinning on an `event` portable, because calling `flush` guarantees that the `event` the user is spinning on will eventually become `complete`.

etomzak: > In case it helps: I think this is precisely what the clarification in https://github.com/KhronosGroup/SYCL-Docs/pull/808 is for.

I was looking at #808 earlier, and my reading was that if `e.get_info()` returns `event_command_status::complete`, it is as-if the user had called `e.wait()` earlier (even if the user hadn't called `e.wait()`). This makes sense and follows necessarily from existing behavior --- the user could have waited on a later event in a chain or on the queue. Hence why #808 is tagged as a `clarification`.

If the goal of #808 was, in fact, to enable spinning on an event, then I don't think the wording in #808 is sufficient. If #808 was intended to enable spinning, then I'm also concerned that it wasn't a clarification but a feature change. I don't think overloading a `get_info()` function to occasionally be a synchronization point is good API design.

> Spinning on an `event` status doesn't force a command to make forward progress, but reading `event::get_info()` is a synchronization point if it returns `complete`. Without `flush`, there's no guarantee that an `event` will ever become `complete`; spinning on an `event` may work on some implementations, but there's no guarantee.
> 
> Adding `flush` makes spinning on an `event` portable, because calling `flush` guarantees that the `event` the user is spinning on will eventually become `complete`.

I understand that this is how @TApplencourt intends for this code to work. What I'm questioning is whether the spec text supports this interpretation. I'm also wondering about the implementation machinery needed to make all events on all backends have this dynamic behavior all the time.

The more I think about this, the more it looks to me that this PR is trying to retcon a new interpretation onto `event::get_info()` in the core spec to make this KHR work. This is bad. If this KHR needs a way to spin on an event, then it should provide a way to spin on an event.

Pennycook: > I was looking at #808 earlier, and my reading was that if `e.get_info()` returns `event_command_status::complete`, it is as-if the user had called `e.wait()` earlier (even if the user hadn't called `e.wait()`). This makes sense and follows necessarily from existing behavior --- the user could have waited on a later event in a chain or on the queue. Hence why #808 is tagged as a `clarification`.

Your reading is correct.

But I disagree with the last bit about "a user could have waited on a later event or on the queue".  SYCL guarantees progress if you call `wait`, but it doesn't guarantee a lack of progress if you don't call `wait`.  Implementations are free to run kernels as soon as their dependencies are satisfied (as long as doing so doesn't block the host thread), and if such an implementation flags via `e.get_info()` that an `event` is `complete`, that implementation must _also_ guarantee that the thread which read `e.get_info()` can see all the side effects of any kernels that must have finished.  

The user doesn't (and can't) know why `e.get_info()` returns `complete`, so this is a user-friendly solution.  If seeing that an `event` is `complete` isn't sufficient to meaningfully do anything, why have the query at all?

> If the goal of #808 was, in fact, to enable spinning on an event, then I don't think the wording in #808 is sufficient. If #808 was intended to enable spinning, then I'm also concerned that it wasn't a clarification but a feature change. I don't think overloading a `get_info()` function to occasionally be a synchronization point is good API design.

It wasn't to enable spinning, but we spotted the gap in the wording when considering this use-case.  #808 alone isn't sufficient to enable spinning, but it is when combined with `flush` (which is a new feature).

`get_info()` occasionally being a synchronization point has a precedent in multiple other APIs.  CUDA's [`cudaEventQuery`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g2bf738909b4a059023537eaa29d8a5b7) works this way, for example.  In ISO C++, an atomic read of a variable synchronizes-with earlier writes of the same variable -- a spin loop that atomically reads a value until it sees a specific value has exactly the same behavior of only synchronizing on success that we are talking about here.

> I understand that this is how @TApplencourt intends for this code to work. What I'm questioning is whether the spec text supports this interpretation. I'm also wondering about the implementation machinery needed to make all events on all backends have this dynamic behavior all the time.

What part of the spec text _doesn't_ support this interpretation?  If you agree that calling `wait` is sufficient to synchronize the host and device while putting all completed events in the `complete` state, how does moving the caller of `wait` to another host thread change anything?

> 
> The more I think about this, the more it looks to me that this PR is trying to retcon a new interpretation onto `event::get_info()` in the core spec to make this KHR work. This is bad. If this KHR needs a way to spin on an event, then it should provide a way to spin on an event.

I think we have to view these things as separate.  Let's close on whether #808 is a clarification in #808, first.

etomzak: > I think we have to view these things as separate. Let's close on whether https://github.com/KhronosGroup/SYCL-Docs/pull/808 is a clarification in https://github.com/KhronosGroup/SYCL-Docs/pull/808, first.

I agree clarity on #808 needs to come first. I'll comment here for now, but if we don't converge in a few more comments, we should probably open a separate issue.

> SYCL guarantees progress if you call `wait`, but it doesn't guarantee a lack of progress if you don't call `wait`. Implementations are free to run kernels as soon as their dependencies are satisfied (as long as doing so doesn't block the host thread)

Agreed

> and if such an implementation flags via `e.get_info()` that an event is complete, that implementation must also guarantee that the thread which read `e.get_info()` can see all the side effects of any kernels that must have finished.

I'm cautiously in agreement with this. The key question is the other way around, though: If an implementation runs a kernel to completion before any call to `wait()` (either because the implementation is sufficiently eager or because of a call to flush), is `e.get_info()` required to eventually return `complete` (even if `wait()` is never called)? I think you'd answer *yes*, but I'm less sure. Your comments re: atomics are exactly what I had in mind. What I'd like to get clarity on is whether #808 has quietly created a requirement for `event::get_info()` to have some atomic-like behavior (e.g., it might need to be implemented using an atomic) where that requirement wasn't there before.

As an aside, when you say, "... that implementation must also guarantee that the thread which read `e.get_info()` can see all the side effects of any kernels that must have finished," I suspect that this statement could be unimplemented when USM shared atomics are involved. There's no way to guarantee that the host view of USM shared atomics and whatever atomic `e.get_info()` is backed by are updated in a specific order. ATM I don't think this materially impacts the question about #808.

TL;DR: I'm not convinced that `event::get_info()` was intended to provide a "running commentary" on what's happening asynchronously in the backend or on device prior to #808. I'm not convinced that using `event` in this way would be a good feature. But if this *is* the intended behavior, I don't think the wording in #808 adequately captures it.

Pennycook: > I'm cautiously in agreement with this. The key question is the other way around, though: If an implementation runs a kernel to completion before any call to `wait()` (either because the implementation is sufficiently eager or because of a call to flush), is `e.get_info()` required to eventually return `complete` (even if `wait()` is never called)? I think you'd answer _yes_, but I'm less sure. Your comments re: atomics are exactly what I had in mind. What I'd like to get clarity on is whether #808 has quietly created a requirement for `event::get_info()` to have some atomic-like behavior (e.g., it might need to be implemented using an atomic) where that requirement wasn't there before.

I would definitely answer _yes_ if `flush()` had been called.  If neither `flush()` nor `wait()` had been called, I think strictly speaking the implementation wouldn't have to return `complete` until the program entered a state where the developer could infer that the kernel had completed (e.g., because a dependent kernel completed, or the host and device synchronized via a system-wide USM atomic).

I think implementations _should_ return `complete` when a kernel has completed, because the potential to end up in a situation where one thread concludes that something has happened (e.g., because of the state of an atomic) while another thread concludes it hasn't happened yet (e.g., because `event.get_info()` doesn't return `complete`) seems really dangerous and confusing.

> As an aside, when you say, "... that implementation must also guarantee that the thread which read `e.get_info()` can see all the side effects of any kernels that must have finished," I suspect that this statement could be unimplemented when USM shared atomics are involved. There's no way to guarantee that the host view of USM shared atomics and whatever atomic `e.get_info()` is backed by are updated in a specific order. ATM I don't think this materially impacts the question about #808.

I disagree: we need to provide that guarantee.  Perhaps we need further clarifications, and may be missing some _Synchronization_ paragraphs on relevant APIs (e.g., `wait`).

`wait` _must_ operate in a way that makes it possible to reason about whether atomic operations happened before or after the `wait`.  I can't find where in the specification we say this, but for USM to work at all, all of the operations submitted before `wait` must happen before the host thread which calls `wait` is unblocked.

**EDIT**: I found it; [3.9.8.1. Host-device coordination](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_host_device_coordination) defines the happens-before behavior for any device-host interaction involving a blocking function (which includes `wait`).

> TL;DR: I'm not convinced that `event::get_info()` was intended to provide a "running commentary" on what's happening asynchronously in the backend or on device prior to #808. I'm not convinced that using `event` in this way would be a good feature. But if this _is_ the intended behavior, I don't think the wording in #808 adequately captures it.

Okay.  What would you add?

etomzak: > I would definitely answer yes if flush() had been called. If neither flush() nor wait() had been called, I think strictly speaking the implementation wouldn't have to return complete until the program entered a state where the developer could infer that the kernel had completed (e.g., because a dependent kernel completed, or the host and device synchronized via a system-wide USM atomic).

Agreed. The way I think about it is that an implementation is allowed to be eager or lazy (or some combination), and for almost all practical purposes[1], it's impossible for the user to observe the difference. It follows that an implementation could secretly be eager but could pretend to be lazy. Even if we make it clear in the spec that the output of `event::get_info()` must transition from `submitted` to `complete` on an eager implementation without the user calling `wait()`, the implementation could just pretend to be lazy and not report `complete` until `wait()` has been called. So I see very little value in what #808 is trying to achieve.

`flush()` substantially changes the considerations. `flush()` forces some level of eagerness, which implies that an implementation is no longer allowed to pretend to be completely lazy after the user calls `flush()`. This is why I said that the spinning-on-`event::get_info()` behavior should be part of the flush KHR --- the functionalities are tightly coupled and belong together.

[1] I think we're both finding that some USM shared atomics (and only USM shared atomics) are creating host-observable kernel side effects in a way that nothing else in SYCL does. A lot of the reasoning about forward progress and synchronization is fairly straightforward until shared atomics are added into the mix. Given that USM as a whole is optional, and shared atomics are only available on some hardware that supports USM, I think we should first figure out what the wording should be for the required SYCL feature set, and then identify what additional requirements the optional features impose.

> What would you add?

You're asking how I would specify a feature that I don't think should exist? Is this a trap? The big problem ATM is that `complete` is defined in a sort-of logically contradictory way. `complete` is defined as, "what `wait()` does, except with visibility into eagerness," and `wait()` is defined as, "Deliberately designed to hide laziness vs. eagerness." That's the source of the lying problem I mentioned above. I think what we need is a forward progress definition for `flush()` (instead of defining `flush()` in terms of `wait()`) (or, alternatively, a `force_eager` queue property). Then the definition of `event::get_info()` would just say something like, "**must** eventually return `complete`".

> I disagree: we need to provide that guarantee. Perhaps we need further clarifications, and may be missing some Synchronization paragraphs on relevant APIs (e.g., `wait`).
>
> `wait` must operate in a way that makes it possible to reason about whether atomic operations happened before or after the `wait`. I can't find where in the specification we say this, but for USM to work at all, all of the operations submitted before `wait` must happen before the host thread which calls `wait` is unblocked.

I'm bumping up against the limits of my understanding on this one, but I'm pretty sure what you're describing has dramatic performance implications. (If anyone knows any better, please correct me.) On a hardware level, there are really just two ways of ensuring consistency: Atomics, where you work with data that fits into one instruction (e.g. 128 bytes), or fences, where you flush data to memory. Confusingly, `atomic_ref` is allowed to use locking (i.e., fences) to achieve atomicity [[cppreference](https://en.cppreference.com/w/cpp/atomic/atomic_ref/is_lock_free.html)]. In an asynchronous environment, there are no guarantees about the order that writes become visible to two independent atomic objects, except if there are memory fences involved.

> all of the operations submitted before `wait` must happen before the host thread which calls `wait` is unblocked

The way to achieve this would be for `wait()` to contain a memory fence that covers shared allocations. I have no idea if existing implementations do this.

A corollary to this is that when #808 defines `event::get_info()` WRT `wait()`, it's essentially asking that `event::get_info()` read all shared allocations back into CPU memory. You can imagine what spinning on this function will do for performance.

TApplencourt: > I think what we need is a forward progress definition for flush() (instead of defining flush() in terms of wait()) (or, alternatively, a force_eager queue property). Then the definition of event::get_info() would just say something like, "must eventually return complete".

Good idea! Please provide a wording.

Regarding the `or, alternatively, a force_eager queue property`, we already discussed it. And... we rejected it. Because it will force implementers too duplicate  their implementation effort twice by providing both an eager and lazy queue implementation. And also you can imagine, lot of implementation "lazy with timeout", or "greedy if low volume of request". It seem really hard to specified implementation details, but have fun if you are interested to go this way!  We can have two competing KHR and see the one who have more adoptions :) 

Also, if you go this way,  please provide a CTS tests to verify that the implementation is indeed lazy. It will be interesting! Because if something is not-observable, it's useless to put it in the spec in a normative way. 

---
I noticed that, you put back `atomic` and `shared memory` to the picture. I totally agree that those area need improvement. But this have nothing to do with the PR. This PR is just about "async" wait.  We don't change anything else.  We just put a name on `std::async(std::launch::async, [&] {this->wait();});`. Can we at least agree on that?

We still live with  current spec and all their dark alley where their is dragon.  I agree, that thinking about `wait`, may make your spirit go wild and open endless possibility of intellectual conundrums. But this is not the correct place.

I will be more than happy to review your PR who clarify shared memory and atomic! 

I hope this message helped clarifying a little bit your ideas,

etomzak: > And... we rejected it. Because it will force implementers too duplicate their implementation effort twice by providing both an eager and lazy queue implementation.

The WG rejected creating new queue properties that would require implementers to support new eager/lazy behaviors, as you've correctly noted. I'm not suggesting that implementers implement new behaviors. AIUI, DPC++ doesn't actually need `flush()`, because it already implements the forward progress your use case requires. AIUI, an upcoming release of AdaptiveCpp will support the forward progress your use case requires, so again, `flush()` isn't needed.

Therefore, I'm suggesting that there is an alternative to the `khr_flush` extension: Specify a queue property extension that corresponds to the behavior in DPC++ and AdaptiveCpp. Existing implementations can make this their default, so no user code needs to change. If a new SYCL implementation comes along, it can choose to support this particular implementation-defined behavior, or do something else.

I haven't completely thought through the pros and cons of `khr_flush` vs. `khr_queue_eager_like_dpc`, but I think it should be considered.

> I noticed that, you put back `atomic` and `shared memory` to the picture. I totally agree that those area need improvement. But this have nothing to do with the PR.

The reason these topics keep coming up is because we need to answer the question, "What are the observable effects of the `khr_flush` extension?" We can't just ratify an extension without understanding what it does.

There's an interpretation that says that `flush()` has no (functionally) observable effects. That was my initial understanding, which is why I asked if it's meant to be a performance hint.

There's an interpretation that says that `flush()` will have observable effects via `event::get_info()`. See my comments above for why I think this is problematic.

The most interesting observable effect of `flush()` (and also the most potentially useful, and also the most foot-gun-y) is in the interaction with shared atomics. As I said above, I'd prefer to first figure out what this extension and #808 mean when shared atomics are ignored, and then figure out the implications on shared atomics. But I don't think this extension can be ratified without understanding the interaction with shared atomics.

TApplencourt: Funny that you didn't reply to only my question:

>We just put a name on std::async(std::launch::async, [&] {this->wait();});. Can we at least agree on that?

Please, can we try to set some baseline of understanding / truth so we can also make forward progress on this discussion (ohoh)? Just reply "yes" or "no"

If you agree on the previous sentence, you will realize that this KHR, doesn't add any new observable effect. None.  If problems exist with `wait`, they are just "exposed" here. So if you have problem with the obervability of `wait`, as I said early, please open a new issue.

>There's an interpretation [...]

Please. If you are agree with the previous question, there is no "interpretation" of this KHR, as if some stuff are unclear. This KHR is crystal clear. 

Their is just confusion on your side about wait in a  multi-threaded environment; and confusion in the general spec about atomic and shared memory with wait. Nothing related to "flush" or this "KHR"

#### Your proposal for a new KHR 

>Specify a queue property extension that corresponds to the behavior in DPC++ and AdaptiveCpp.

Amazing! At least we seem to agree on our first point : You agree that we should add something to the spec to give more control over the execution to the user. I pleasantly surprised!

Just to be sure. You propose a "Implementation defined behavior" and a "eagger" one? So people will not be able to opt-in for "lazy"? Why ?  C++ provide both (https://en.cppreference.com/w/cpp/thread/launch.html) 

I found `flush` far more general. "Just write a portable code, provide hight level semantic information to your runtime, and let the implementation optimize it". I suppose it's a "functional" versus "imperative" mind set.  The "property" is far more imperative.

But we can discuss that on your PR with you new proposal.

illuhad: > . AIUI, DPC++ doesn't actually need flush(), because it already implements the forward progress your use case requires. AIUI, an upcoming release of AdaptiveCpp will support the forward progress your use case requires, so again, flush() isn't needed.

I don't think this is necessarily correct. DPC++ and AdaptiveCpp control the submission behavior on the SYCL level. But if a backend internally uses lazy submission, you might still not get the right behavior.
AdaptiveCpp has now implemented eager submission on the SYCL-level as you mention, but that does for example not help you if you have a lazily executing OpenCL implementation (the other backends are greedy internally, so no issues there). I suspect that DPC++ will probably have the same issue on the OpenCL side (?).

With that said, when it comes to implementing `khr_flush()`, I would probably rather look into forcing the OpenCL backend to always execute greedily (e.g. regularly calling `clFlush()`) to align it with the other backends. This would avoid the complexities with cross-queue dependencies as discussed in the WG call.

If it is possible to force OpenCL into eager submission without performance drawbacks, then it won't matter whether there is `khr_flush` or a property implementation-wise, since both will be a no-op.

adoc/extensions/sycl_khr_queue_empty_query.adoc
```diff
@@ -32,6 +32,10 @@ below.
 This extension adds the following function to the [code]#sycl::queue# class,
 which provides information about the emptiness of the queue.
 
+{note} This khr is more useful when used in conjunction with
```
Pennycook: Nit: Outside of the working group people might struggle to understand what "khr" means in this context.

```suggestion
{note} This feature is most useful when used in conjunction with
```

keryell (approved):
Thanks!

gmlueck (approved):

PR #810, opened 2025-05-08
Remove spurious line from device info descriptor code

etomzak:
I'm pretty sure this line shouldn't be here.


gmlueck (approved):
Good catch!

gmlueck:
Merging as editorial.

PR #811, opened 2025-05-08
Let implementations elect not to use secondary queues

steffenlarsen:
This change changes the wording of secondary queues and their exception conditions to specify that an implementation may elect to not use the specified secondary queue.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5534,9 +5534,10 @@ the implementation to throw an [code]#exception# with the [code]#errc::invalid#
 error code from the [code]#accessor# constructor (if the accessor is not a
 placeholder) or from [code]#handler::require()# (if the accessor is a
 placeholder).  If the accessor is bound to a <<command-group>> with a secondary
-queue, the sub-buffer's alignment must be compatible with both the primary
-queue's device and the secondary queue's device, otherwise this exception is
-thrown.
+queue that the implementation intends on using in case of failure during
+enqueuing the <<command-group>> to the primary queue, the sub-buffer's alignment
+must be compatible with both the primary queue's device and the secondary
+queue's device, otherwise this exception is thrown.
 
 Must throw an [code]#exception# with the [code]#errc::invalid# error code if
 [code]#b# is a sub-buffer.
@@ -14288,7 +14289,7 @@ A user can therefore supply a secondary queue when submitting a command group to
 the primary queue.
 If the <<sycl-runtime>> fails to enqueue or execute a command group on a primary
 queue, it can attempt to run the command group on the secondary queue.
-The circumstances in which it is, or is not, possible for a <<sycl-runtime>> to
+The circumstances in which the <<sycl-runtime>> may elect to
 fall-back from primary to secondary queue are unspecified in the specification.
 Even if a command group is run on the secondary queue, the requirement that host
 code within the command group is executed exactly once remains, regardless of
@@ -14598,9 +14599,13 @@ This invocation function ignores any [code]#kernel_bundle# that was bound to
 this command group handler via [code]#handler::use_kernel_bundle()# and instead
 implicitly uses the kernel bundle that contains the [code]#kernelObject#.
 Throws an [code]#exception# with the [code]#errc::kernel_not_supported# error
-code if the [code]#kernelObject# is not compatible with either the device
-associated with the primary queue of the <<command-group>> or with the device
-associated with the secondary queue (if specified).
+code if the [code]#kernelObject# is not compatible with the device
+associated with the primary queue of the <<command-group>>.
+Throws an [code]#exception# with the [code]#errc::kernel_not_supported# error
+code if the <<command-group>> has a secondary queue associated that the
+implementation intends on using in the case that the <<command-group>> fails to
+enqueue on the primary queue, and the [code]#kernelObject# is not compatible
+with the device associated with the secondary queue of the <<command-group>>.
 
 a@
 [source]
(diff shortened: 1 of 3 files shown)


steffenlarsen:
Alternative wording: https://github.com/KhronosGroup/SYCL-Docs/pull/812

adoc/chapters/programming_interface.adoc
```diff
@@ -14598,9 +14599,13 @@ This invocation function ignores any [code]#kernel_bundle# that was bound to
...
-code if the [code]#kernelObject# is not compatible with either the device
-associated with the primary queue of the <<command-group>> or with the device
-associated with the secondary queue (if specified).
+code if the [code]#kernelObject# is not compatible with the device
+associated with the primary queue of the <<command-group>>.
+Throws an [code]#exception# with the [code]#errc::kernel_not_supported# error
+code if the <<command-group>> has a secondary queue associated that the
+implementation intends on using in the case that the <<command-group>> fails to
+enqueue on the primary queue, and the [code]#kernelObject# is not compatible
+with the device associated with the secondary queue of the <<command-group>>.
```
gmlueck, line 14608: For all these kernel invocation commands, we can shorten the description by using the new terminology:

> Throws an `exception` with the `errc::kernel_not_supported` error code if the \<\<command-group\>\> has a secondary queue, the implementation supports the secondary queue fallback, and the `kernelObject` is not compatible with the device associated with the secondary queue.

gmlueck (commented):
I prefer this wording over #812.  However, I think we can make it a bit more concise by introducing some new terminology. See comments below.

steffenlarsen:
Thanks, @gmlueck! I've applied the suggested rewordings as well as some related ones.

TApplencourt (approved):

TApplencourt:
LGTM. Thanks! And good luck fixing the
> asciidoctor: ERROR: the table cell on page 138 has been truncated; Asciidoctor PDF does not support table cell content that exceeds the height of a single page

:)

gmlueck:
> > asciidoctor: ERROR: the table cell on page 138 has been truncated; Asciidoctor PDF does not support table cell content that exceeds the height of a single page

Yuck!  Here's some info that may help.  As you can see, the asciidoctor error message isn't very helpful because it doesn't tell you the source line of the input *.adoc file that causes the problem.  The "page 138" is only marginally useful because asciidoctor doesn't create an output PDF when an error occurs.  However, you can use the fact that you didn't change the document very much.

Look at the published [PDF spec](https://registry.khronos.org/SYCL/specs/sycl-2020/pdf/sycl-2020.pdf) and find page 138.  You can see that this corresponds to the table cell defining the `buffer` constructor that takes a `subRange`.  You can also see that this table cell is already almost an entire page high.  Thus, any small increase in the text will cause it to overflow the limit.

Your PR did change the text describing this constructor, so that is the culprit.  The easiest fix may be to try to shorten the description of this descriptor somehow.  It can be any part of the description, not necessarily the new text you added.  Another possible option is to split the description into two rows of the table.  I don't think we've done this before, though, so it would be better to shorten the description if possible.

The *real* fix is to restructure this part of the spec to use our new style, which does not cram everything into tables.  This has many advantages, one of which is that we avoid this sort of error with overflowing table cells.

adoc/chapters/programming_interface.adoc
```diff
@@ -14598,9 +14599,12 @@ This invocation function ignores any [code]#kernel_bundle# that was bound to
...
 implicitly uses the kernel bundle that contains the [code]#kernelObject#.
 Throws an [code]#exception# with the [code]#errc::kernel_not_supported# error
-code if the [code]#kernelObject# is not compatible with either the device
-associated with the primary queue of the <<command-group>> or with the device
-associated with the secondary queue (if specified).
+code if the [code]#kernelObject# is not compatible with the device
+associated with the primary queue of the <<command-group>>.
+Throws an [code]#exception# with the [code]#errc::kernel_not_supported# error
+code if the <<command-group>> has a secondary queue, the implementation supports
+the secondary queue fallback, and the [code]#kernelObject# is not compatible
```
Pennycook, line 14606:
```suggestion
secondary queue fallback, and the [code]#kernelObject# is not compatible
```

Same comment about wording as above, just flagging this as a place to change.

adoc/chapters/programming_interface.adoc
```diff
@@ -5535,8 +5535,9 @@ error code from the [code]#accessor# constructor (if the accessor is not a
 placeholder) or from [code]#handler::require()# (if the accessor is a
 placeholder).  If the accessor is bound to a <<command-group>> with a secondary
 queue, the sub-buffer's alignment must be compatible with both the primary
-queue's device and the secondary queue's device, otherwise this exception is
-thrown.
+queue's device and the secondary queue's device. If the implementation supports
+the secondary queue fallback, it also throws this exception if the sub-buffer's
```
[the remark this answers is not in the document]
gmlueck: The intent is definitely (1).  We want to give implementations the freedom to decide whether the secondary queue is ever used at all.  We expect DPC++ to not support secondary queue fallback for any queue.

Note that the secondary queue fallback feature is being deprecated in SYCL-Next, so it will likely be removed in the release after that.

tomdeakin:
One more review, and needs a CTS change

keryell (approved):
@Pennycook suggestions look good.
Thanks.

tomdeakin:
Blocked by https://github.com/KhronosGroup/SYCL-Docs/pull/824

gmlueck:
@steffenlarsen I think this PR is now unblocked because #824 has been merged.  I think the following has to happen first:

* I just realized that this PR is targeting the "sycl-2020" branch.  Please retarget it to the "main" branch instead.  Our process is to target all PRs to the "main" branch and then cherry-pick (if appropriate) to "sycl-2020".

* Since we merged #824, there are now merge conflicts with this PR that must be resolved.  Doing this should also fix the CI build error.

* I see that @Pennycook made a few comments about editorial changes.  Could you address these?  Personally, I think his suggestions look reasonable.

steffenlarsen:
Thanks, @gmlueck ! It has been rebased, re-targeted and I should have addressed all @Pennycook's comments. 🚀

Pennycook (approved):

gmlueck:
@steffenlarsen thanks for making these updates.

Unfortunately, it seems like the CI test fails when I run it manually.  I think the CI workflow didn't run automatically in this PR when you did the "retarget", which is why it didn't fail here.  Could you run "./adoc/scripts/verify_reflow_conformance.sh" manually, and then fix the errors?  Or, you could probably push an empty commit to cause the CI to run, and then fix the errors.

Thanks!

PR #812, opened 2025-05-08
Specify that implementations may choose to not use secondary queues

steffenlarsen:
This commit changes the specification for the submit function taking a secondary queue, specifying that implementations may choose to not associate the secondary queue with the command-groups. In tandem with this, the changes use the term "associated" in more places to make it clearer that cases requiring exceptions for secondary queues are only if the implementations elect to associate the secondary queues with the command-groups.


steffenlarsen:
Wording in https://github.com/KhronosGroup/SYCL-Docs/pull/811 is preferred. Closing this.

PR #823, opened 2025-05-20
Remove references to "inter-device copy"

### LINKED ISSUE #820: What is an "inter-device copy" and when does it happen?
### Specification Version

SYCL 2020 (Source)

### Section Number(s)

- Section 3.13.1. Device copyable, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable

### Issue Description

The description of "device copyable" seems to imply that it is only legal for an implementation to treat a type as bitwise-copyable when performing an "inter-device copy".  Specifically, see the following statements:

> The SYCL implementation may need to copy data between the host and a device or between two devices.

> Although these view types are device copyable, the implementation copies just the view and not the contained data when doing an inter-device copy.

> It is unspecified whether the implementation actually calls the copy constructor, move constructor, copy assignment operator, or move assignment operator of a class declared as is_device_copyable_v when doing an inter-device copy. Since these operations must all be the same as a bitwise copy, the implementation may simply copy the memory where the object resides.

However, SYCL doesn't actually define what constitutes an "inter-device" copy, or when it happens.  The current text also makes allowances for types that are not *actually* bitwise copyable, saying that:

> Likewise, it is unspecified whether the implementation actually calls the destructor for such a class on the device since the destructor must have no effect on the device.

...which seems to imply that if a copy is created on the host, it should be expected that the destructor *does* run.

These descriptions are problematic because there is an unclear contract between the developer and the implementation.  As written, I don't think it is possible to answer these questions:

1. If a kernel function object (or lambda) is passed to `parallel_for`, can the implementation create a bitwise copy of it?  Doing so would create a copy on the host, but bypass the copy constructor and destructor.

2. Similarly: Can the implementation create a bitwise copy of any value passed to `set_arg` or `set_args`? 

3. If a kernel function object (or lambda) is passed to `parallel_for`, can the implementation create a regular copy of it?  The copy constructor and destructor would both run, even though the developer might have expected the call to `parallel_for` to be the point at which the arguments are transferred to the device.

4. Similarly: Can the implementation create a regular copy of any value passed to `set_arg` or `set_args`?

In order to be able to answer questions like these, the specification needs to be rewritten to define the behavior of device-copyable types in terms of their interaction with specific functions from the SYCL API.

### Code Example (Optional)

How many copies and how many destructors are run in the example below?  How many are guaranteed to run?

```c++
#include <sycl/sycl.hpp>
#include <atomic>

static std::atomic<int> ndestructors = 0;

struct MyArgument {
  MyArgument() {}
  MyArgument(const MyArgument& other) { this->ncopies = other.ncopies + 1; }
  ~MyArgument() {
#ifndef __SYCL_DEVICE_ONLY__
    ndestructors++;
#endif
  }
  int ncopies = 0;
};

template<>
struct sycl::is_device_copyable<MyArgument> : std::true_type {};

int main() {

    sycl::queue q;
    int* output = sycl::malloc_shared<int>(1, q);

    MyArgument arg;
    q.single_task([=]() {
        *output = arg.ncopies;
    }).wait();

    std::cout << "Copies: " << *output << std::endl;
    std::cout << "Destructors: " << ndestructors << std::endl;

    sycl::free(output, q);
}
```

Pennycook:
This PR clarifies the how device-copyable types work by removing vague references to "inter-device" copies.  Specifically, it makes the following changes:
- Clarifies that copying views _always_ requires some thought about where the underlying data is allocated.
- Clarifies that the effect of copying a device-copyable type must always be the same as a bitwise copy.
- Clarifies that the implementation can therefore assume that a device-copyable type is bitwise copyable.
- Clarifies that declaring a type is device-copyable when it isn't results in undefined behavior.

I also fixed a few typos and reworked some of the surrounding text to make it flow better around my changes.

Closes #820.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1838,15 +1838,13 @@ device copyable.
 
 [NOTE]
 ====
-The types [code]#std::basic_string_view<CharT, Traits># and
-[code]#std::span<ElementType, Extent># are both view types, which reference
-underlying data that is not contained within their type.
-Although these view types are device copyable, the implementation copies just
-the view and not the contained data when doing an inter-device copy.
-In order to reference the contained data after such a copy, the application must
-allocate the contained data in unified shared memory (USM) that is accessible on
-both the host and device (or on both devices in the case of a device-to-device
-copy).
+Types such as [code]#std::basic_string_view<CharT, Traits># and
+[code]#std::span<ElementType, Extent># are view types, which reference
+underlying data that they do not own.
+Copying such a type only copies the view and not the referenced data.
+If a view is copied between the host and device or between two devices, it is
+the application's responsibility to ensure that the referenced data is allocated
+in memory that can be accessed by the recipient (see <<sec:usm>>).
 ====
 
 In addition, the implementation may allow the application to explicitly declare
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1864,7 +1864,7 @@ copyable if all of the following statements are true:
   * The effect of each eligible copy constructor, move constructor, copy
     assignment operator, and move assignment operator is the same as a bitwise
     copy of the object;
-  * Type [code]#T# has a [code]#public# non-deleted destructor;
+  * Type [code]#T# has a [code]#public# non-deleted destructor; and
   * The destructor has no effect when executed on the device.
 
 When the application explicitly declares a class type to be device copyable,
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1861,10 +1861,9 @@ copyable if all of the following statements are true:
     copy assignment operator, or move assignment operator;
   * Each eligible copy constructor, move constructor, copy assignment operator,
     and move assignment operator is [code]#public#;
-  * When doing an inter-device transfer of an object of type [code]#T#, the
-    effect of each eligible copy constructor, move constructor, copy assignment
-    operator, and move assignment operator is the same as a bitwise copy of the
-    object;
+  * The effect of each eligible copy constructor, move constructor, copy
+    assignment operator, and move assignment operator is the same as a bitwise
+    copy of the object;
   * Type [code]#T# has a [code]#public# non-deleted destructor;
   * The destructor has no effect when executed on the device.
 
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1852,11 +1852,11 @@ certain class types as device copyable.
 If the implementation has this support, it must predefine the preprocessor macro
 [code]#SYCL_DEVICE_COPYABLE# to [code]#1#, and it must not predefine this
 preprocessor macro if it does not have this support.
-When the implementation has this support, a class type [code]#T# is device
-copyable if all of the following statements are true:
+When the implementation has this support, an application may declare that a
+class type [code]#T# is device copyable by defining the trait
+[code]#is_device_copyable_v<T># to [code]#true# if all of the following
+statements are true:
 
-  * The application defines the trait [code]#is_device_copyable_v<T># to
-    [code]#true#;
   * Type [code]#T# has at least one eligible copy constructor, move constructor,
     copy assignment operator, or move assignment operator;
   * Each eligible copy constructor, move constructor, copy assignment operator,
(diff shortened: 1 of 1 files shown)


adoc/chapters/architecture.adoc
```diff
@@ -1838,57 +1841,43 @@ device copyable.
...
-  * When doing an inter-device transfer of an object of type [code]#T#, the
-    effect of each eligible copy constructor, move constructor, copy assignment
-    operator, and move assignment operator is the same as a bitwise copy of the
-    object;
-  * Type [code]#T# has a [code]#public# non-deleted destructor;
+  * The effect of each eligible copy constructor, move constructor, copy
+    assignment operator, and move assignment operator is the same as a bitwise
+    copy of the object;
+  * Type [code]#T# has a [code]#public# non-deleted destructor; and
   * The destructor has no effect when executed on the device.
```
gmlueck, line 1871: It seems like we also need to change this statement to remove the part about "when executed on the device".  We say earlier that an implementation may assume that it is always safe to perform bitwise copies of a device-copyable type (even copies on the host).  It seems inconsistent to allow an implementation to elide the type's copy-constructor but still execute the type's destructor.

It also raises questions about what an implementation is required to do regarding the destructor.  Consider a case where the implementation copies an object on the host using bitwise copy into a temporary buffer of bytes.  That temporary buffer is eventually used to call some backend API that sets the argument's value on the device.  What must the implementation do when it destroys the temporary buffer?  Must it cast the buffer back to the object's type and then execute the destructor?  The current wording implies that this might be required since the buffer is on the host.  If we change the spec to say that the buffer must always have no effect, then this issue goes away, and it's clear that the implementation is not required to execute the destructor on the temporary buffer.

Pennycook: > It seems like we also need to change this statement to remove the part about "when executed on the device". We say earlier that an implementation may assume that it is always safe to perform bitwise copies of a device-copyable type (even copies on the host). It seems inconsistent to allow an implementation to elide the type's copy-constructor but still execute the type's destructor.

I wasn't sure about this, which is why I left it in.  Unlike with the copy constructor, the specification _does_ provide a way to ensure that the destructor is defined differently for host and device (via `__SYCL_DEVICE_ONLY__`).  Somebody might be relying on this, and so removing this language might break existing code.

I don't like this behavior and I think it might have been a mistake to specify things this way, but I didn't want to change the behavior without discussing it first.

@nliber, @TApplencourt - Are any of the Kokkos or Argonne codes relying on this destructor behavior?

> It also raises questions about what an implementation is required to do regarding the destructor. Consider a case where the implementation copies an object on the host using bitwise copy into a temporary buffer of bytes. That temporary buffer is eventually used to call some backend API that sets the argument's value on the device. What must the implementation do when it destroys the temporary buffer? Must it cast the buffer back to the object's type and then execute the destructor? The current wording implies that this might be required since the buffer is on the host. If we change the spec to say that the destructor must always have no effect, then this issue goes away, and it's clear that the implementation is not required to execute the destructor on the temporary buffer.

I agree it would be odd to require implementations to execute the destructor manually.  I don't want a solution that mandates that.

Saying that the destructor must always have no effect would make sense, and I'm not opposed to it.  I'm just reluctant because of the reasons I gave above.

Another alternative would be to say that an implementation must either: 1) not make any copies on the host; or 2) only bitwise copy into a temporary buffer of bytes on the host.  Then the user would have a guarantee that the destructor is only called on the host when the kernel function object is destroyed.  I'm not sure this is really better, though, because guaranteeing that an implementation makes no copies would be a new restriction for implementers.

masterleinad: > @nliber, @TApplencourt - Are any of the Kokkos or Argonne codes relying on this destructor behavior?

`Kokkos` wraps all non-trivially-copyable functors with a wrapper that forces copies to be memcpys and destructors to be empty. We don't diagnose if that's what the user's functors are actually doing or not.
Objects constructed on device code might, of course, have different behavior depending on `__SYCL_DEVICE_ONLY__`.

Pennycook: Thanks, @masterleinad. 

Nothing we're considering changing here would affect the case where an object is constructed and destroyed in device code, so it sounds like we're good.  I'll make the wording change to simplify the destructor case.

TApplencourt:
@masterleinad is happy, I'm happy!

etomzak:
LGTM

AerialMantis (approved):

PR #824, opened 2025-05-20
Change buffer API descriptions to use new style

steffenlarsen:
This commit changes the buffer API descriptions to use the new style of API documentation. This should break up some of the larger tables in the specification.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5217,10 +5217,10 @@ _Effects (2):_ Equivalent to `buffer(bufferRange, AllocatorT{}, propList)`.
 .[apititle]#Construct with host memory#
 [source,role=synopsis,id=api:buffer-ctor-host-mem]
 ----
-buffer(T* hostData, const range<Dimensions>& bufferRange,          (2)
+buffer(T* hostData, const range<Dimensions>& bufferRange,          (1)
        AllocatorT allocator, const property_list& propList = {});
 
-buffer(T* hostData, const range<Dimensions>& bufferRange,          (1)
+buffer(T* hostData, const range<Dimensions>& bufferRange,          (2)
        const property_list& propList = {});
 
 buffer(const T* hostData, const range<Dimensions>& bufferRange,    (3)
@@ -5405,6 +5405,7 @@ later.
 [code]#baseIndex# specifies the origin of the sub-buffer inside the buffer
 [code]#b#.
 [code]#subRange# specifies the size of the sub-buffer.
+
 The origin (based on [code]#baseIndex#) of the sub-buffer being constructed must
 be a multiple of the memory base address alignment of each SYCL [code]#device#
 which accesses data from the buffer.
@@ -5423,12 +5424,13 @@ _Throws:_
 * An [code]#exception# with the [code]#errc::invalid# error code if [code]#b# is
 a sub-buffer.
 
-* An [code]#exception# with the [code]#errc::invalid# error code if
-[code]#subRange# corresponds to a a non-contiguous region of [code]#b#.
+* An [code]#exception# with the [code]#errc::invalid# error code if the subrange
+formed by [code]#baseIndex# and [code]#subRange# is not a contiguous region of
+[code]#b#.
 
 * An [code]#exception# with the [code]#errc::invalid# error code if the sum of
 [code]#baseIndex# and [code]#subRange# in any dimension exceeds the parent
-buffer [code]#b# size [code]#bufferRange# in that dimension.
+buffer [code]#b# size ([code]#bufferRange#) in that dimension.
 
 '''
 
@@ -5552,10 +5554,10 @@ Deprecated in SYCL 2020.
 Use [code]#get_host_access()# instead.
 
 _Returns (1):_ A valid host [code]#accessor# to the buffer with the specified
-access mode and target.
+access mode.
 
 _Returns (2):_ A valid host [code]#accessor# to the buffer with the specified
-access mode and target.
+access mode.
 The accessor is a <<ranged-accessor>>, where the range starts at the given
 offset from the beginning of the buffer.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5728,7 +5728,7 @@ use_mutex(std::mutex& mutexRef);
 ----
 
 _Effects:_ Constructs a SYCL [code]#use_mutex# property instance with a
-reference to [code]#mutexRef# parameter provided.
+reference to [code]#mutexRef#.
 
 '''
 
@@ -5738,8 +5738,8 @@ reference to [code]#mutexRef# parameter provided.
 std::mutex* get_mutex_ptr() const;
 ----
 
-_Returns:_ The [code]#std::mutex# which was specified when constructing this
-SYCL [code]#use_mutex# property.
+_Returns:_ A pointer to the [code]#std::mutex# provided when constructing this
+property.
 
 '''
 
@@ -5779,8 +5779,7 @@ copy of a SYCL [code]#context#.
 context get_context() const;
 ----
 
-_Returns:_ The [code]#context# which was specified when constructing this SYCL
-[code]#context_bound# property.
+_Returns:_ The [code]#context# provided when constructing this property.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5279,7 +5279,7 @@ buffer(Container& container, AllocatorT allocator,
 template <typename Container>
 buffer(Container& container, const property_list& propList = {}); (2)
 ----
-Constraints:_ Available only if [code]#Container# is a contiguous container,
+_Constraints:_ Available only if [code]#Container# is a contiguous container,
 that is the following requirements must be met:
 
 * [code]#std::data(container)# and [code]#std::size(container)# are well formed;
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -5207,14 +5207,16 @@ buffer(const range<Dimensions>& bufferRange, AllocatorT allocator,
 _Effects (1):_ Equivalent to `buffer(bufferRange, AllocatorT{}, propList)`.
 
 _Effects (2):_ Construct a SYCL [code]#buffer# instance with uninitialized
-memory. The constructed SYCL [code]#buffer# will use the [code]#allocator#
-parameter provided when allocating memory on the host. The range of the
-constructed SYCL [code]#buffer# is specified by the [code]#bufferRange#
-parameter provided. Data is not written back to the host on destruction of the
-[code]#buffer# unless the [code]#buffer# has a valid non-null pointer specified
-via the member function [code]#set_final_data()#. Zero or more properties can be
-provided to the constructed SYCL [code]#buffer# via an instance of
-[code]#property_list#.
+memory.
+The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
+provided when allocating memory on the host.
+The range of the constructed SYCL [code]#buffer# is specified by the
+[code]#bufferRange# parameter provided.
+Data is not written back to the host on destruction of the [code]#buffer# unless
+the [code]#buffer# has a valid non-null pointer specified via the member
+function [code]#set_final_data()#.
+Zero or more properties can be provided to the constructed SYCL [code]#buffer#
+via an instance of [code]#property_list#.
 '''
 
 .[apititle]#Construct with host memory#
@@ -5232,35 +5234,39 @@ buffer(const T* hostData, const range<Dimensions>& bufferRange,
 buffer(const T* hostData, const range<Dimensions>& bufferRange,
        AllocatorT allocator, const property_list& propList = {}); (4)
 ----
-_Effects (1):_ Equivalent to
-`buffer(hostData, bufferRange, AllocatorT{}, propList)`.
+_Effects (1):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
 
 _Effects (2):_ Construct a SYCL [code]#buffer# instance with the
-[code]#hostData# parameter provided. The buffer is initialized with the memory
-specified by [code]#hostData#, and the buffer assumes exclusive access to this
-memory for the duration of its lifetime.  The constructed SYCL [code]#buffer#
-will use the [code]#allocator# parameter provided when allocating memory on the
-host.  The range of the constructed SYCL [code]#buffer# is specified by the
-[code]#bufferRange# parameter provided.  Zero or more properties can be provided
-to the constructed SYCL [code]#buffer# via an instance of [code]#property_list#.
+[code]#hostData# parameter provided.
+The buffer is initialized with the memory specified by [code]#hostData#, and the
+buffer assumes exclusive access to this memory for the duration of its lifetime.
+The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
+provided when allocating memory on the host.
+The range of the constructed SYCL [code]#buffer# is specified by the
+[code]#bufferRange# parameter provided.
+Zero or more properties can be provided to the constructed SYCL [code]#buffer#
+via an instance of [code]#property_list#.
 
-_Effects (3):_ Equivalent to
-`buffer(hostData, bufferRange, AllocatorT{}, propList)`.
+_Effects (3):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
 
 _Effects (4):_ Construct a SYCL [code]#buffer# instance with the
-[code]#hostData# parameter provided.  The buffer assumes exclusive access to
-this memory for the duration of its lifetime.
+[code]#hostData# parameter provided.
+The buffer assumes exclusive access to this memory for the duration of its
+lifetime.
 
 The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
 provided when allocating memory on the host.
 
 The host address is [code]#const T#, so the host accesses can be read-only.
 However, the [code]#typename T# is not const so the device accesses can be both
-read and write accesses. Since, the [code]#hostData# is const, this buffer is
-only initialized with this memory and there is no write back after its
-destruction, unless the [code]#buffer# has another valid non-null final data
-address specified via the member function [code]#set_final_data()# after
-construction of the [code]#buffer#.
+read and write accesses.
+Since, the [code]#hostData# is const, this buffer is only initialized with this
+memory and there is no write back after its destruction, unless the
+[code]#buffer# has another valid non-null final data address specified via the
+member function [code]#set_final_data()# after construction of the
+[code]#buffer#.
 
 The range of the constructed SYCL [code]#buffer# is specified by the
 [code]#bufferRange# parameter provided.
@@ -5293,13 +5299,13 @@ _Effects (1):_ Equivalent to `buffer(container, AllocatorT{}, propList)`.
 
 _Effects (2):_ Construct a one dimensional SYCL [code]#buffer# instance from the
 elements starting at [code]#std::data(container)# and containing
-[code]#std::size(container)# number of elements. The buffer is initialized with
-the contents of [code]#container#, and the buffer assumes exclusive access to
-[code]#container# for the duration of its lifetime.
+[code]#std::size(container)# number of elements.
+The buffer is initialized with the contents of [code]#container#, and the buffer
+assumes exclusive access to [code]#container# for the duration of its lifetime.
 
 Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+[code]#buffer# destruction if the return type of [code]#std::data(container)# is
+not [code]#const#.
 
 The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
 provided when allocating memory on the host.
@@ -5327,41 +5333,47 @@ via an instance of [code]#property_list#.
          const range<Dimensions>& bufferRange,
          const property_list& propList = {});
 ----
-_Effects (1):_ Equivalent to
-`buffer(hostData, bufferRange, AllocatorT{}, propList)`.
+_Effects (1):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
 
 _Effects (2):_ When [code]#hostData# is not empty, construct a SYCL buffer with
-the contents of its stored pointer.  The buffer assumes exclusive access to this
-memory for the duration of its lifetime.  The buffer also creates its own
-internal copy of the [code]#shared_ptr# that shares ownership of the
-[code]#hostData# memory, which means the application can safely release
-ownership of this [code]#shared_ptr# when the constructor returns.
+the contents of its stored pointer.
+The buffer assumes exclusive access to this memory for the duration of its
+lifetime.
+The buffer also creates its own internal copy of the [code]#shared_ptr# that
+shares ownership of the [code]#hostData# memory, which means the application can
+safely release ownership of this [code]#shared_ptr# when the constructor
+returns.
 
 When [code]#hostData# is empty, construct a SYCL buffer with uninitialized
 memory.
 
 The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
-provided when allocating memory on the host.  The range of the constructed SYCL
-[code]#buffer# is specified by the [code]#bufferRange# parameter provided. Zero
-or more properties can be provided to the constructed SYCL [code]#buffer# via an
-instance of [code]#property_list#.
+provided when allocating memory on the host.
+The range of the constructed SYCL [code]#buffer# is specified by the
+[code]#bufferRange# parameter provided.
+Zero or more properties can be provided to the constructed SYCL [code]#buffer#
+via an instance of [code]#property_list#.
 
-_Effects (3):_ Equivalent to
-`buffer(hostData, bufferRange, AllocatorT{}, propList)`.
+_Effects (3):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
 
 _Effects (4):_ When [code]#hostData# is not empty, construct a SYCL buffer with
-the contents of its stored pointer.  The buffer assumes exclusive access to this
-memory for the duration of its lifetime.  The buffer also creates its own
-internal copy of the [code]#shared_ptr# that shares ownership of the
-[code]#hostData# memory, which means the application can safely release
-ownership of this [code]#shared_ptr# when the constructor returns.
+the contents of its stored pointer.
+The buffer assumes exclusive access to this memory for the duration of its
+lifetime.
+The buffer also creates its own internal copy of the [code]#shared_ptr# that
+shares ownership of the [code]#hostData# memory, which means the application can
+safely release ownership of this [code]#shared_ptr# when the constructor
+returns.
 
 When [code]#hostData# is empty, construct a SYCL buffer with uninitialized
 memory.
 
 The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
-provided when allocating memory on the host.  The range of the constructed SYCL
-[code]#buffer# is specified by the [code]#bufferRange# parameter provided.
+provided when allocating memory on the host.
+The range of the constructed SYCL [code]#buffer# is specified by the
+[code]#bufferRange# parameter provided.
 Zero or more properties can be provided to the constructed SYCL [code]#buffer#
 via an instance of [code]#property_list#.
 '''
@@ -5378,15 +5390,16 @@ buffer<T, 1>(InputIterator first, InputIterator last,
              const property_list& propList = {});
 ----
 _Effects (1):_ Create a new allocated 1D buffer initialized from the given
-elements ranging from [code]#first# up to one before [code]#last#. The data is
-copied to an intermediate memory position by the runtime. Data is not written
-back to the same iterator set provided. However, if the [code]#buffer# has a
-valid non-const iterator specified via the member function
-[code]#set_final_data()#, data will be copied back to that iterator. The
-constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
-provided when allocating memory on the host. Zero or more properties can be
-provided to the constructed SYCL [code]#buffer# via an instance of
-[code]#property_list#.
+elements ranging from [code]#first# up to one before [code]#last#.
+The data is copied to an intermediate memory position by the runtime.
+Data is not written back to the same iterator set provided.
+However, if the [code]#buffer# has a valid non-const iterator specified via the
+member function [code]#set_final_data()#, data will be copied back to that
+iterator.
+The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
+provided when allocating memory on the host.
+Zero or more properties can be provided to the constructed SYCL [code]#buffer#
+via an instance of [code]#property_list#.
 
 _Effects (2):_ Equivalent to `buffer(first, last, AllocatorT{}, propList)`.
 '''
@@ -5398,33 +5411,35 @@ buffer(buffer& b, const id<Dimensions>& baseIndex,
        const range<Dimensions>& subRange);
 ----
 _Effects:_ Create a new sub-buffer without allocation to have separate accessors
-later. [code]#b# is the buffer with the real data. [code]#baseIndex# specifies
-the origin of the sub-buffer inside the buffer [code]#b#. [code]#subRange#
-specifies the size of the sub-buffer.
+later.
+[code]#b# is the buffer with the real data.
+[code]#baseIndex# specifies the origin of the sub-buffer inside the buffer
+[code]#b#.
+[code]#subRange# specifies the size of the sub-buffer.
 
-The origin (based on [code]#baseIndex#) of the sub-buffer being constructed
-must be a multiple of the memory base address alignment of each SYCL
-[code]#device# which accesses data from the buffer.  This value is retrievable
-via the SYCL [code]#device# class info query
-[code]#info::device::mem_base_addr_align#.  Violating this requirement causes
-the implementation to throw an [code]#exception# with the [code]#errc::invalid#
-error code from the [code]#accessor# constructor (if the accessor is not a
-placeholder) or from [code]#handler::require()# (if the accessor is a
-placeholder).  If the accessor is bound to a <<command-group>> with a secondary
-queue, the sub-buffer's alignment must be compatible with both the primary
-queue's device and the secondary queue's device, otherwise this exception is
-thrown.
+The origin (based on [code]#baseIndex#) of the sub-buffer being constructed must
+be a multiple of the memory base address alignment of each SYCL [code]#device#
+which accesses data from the buffer.
+This value is retrievable via the SYCL [code]#device# class info query
+[code]#info::device::mem_base_addr_align#.
+Violating this requirement causes the implementation to throw an
+[code]#exception# with the [code]#errc::invalid# error code from the
+[code]#accessor# constructor (if the accessor is not a placeholder) or from
+[code]#handler::require()# (if the accessor is a placeholder).
+If the accessor is bound to a <<command-group>> with a secondary queue, the
+sub-buffer's alignment must be compatible with both the primary queue's device
+and the secondary queue's device, otherwise this exception is thrown.
 
 _Throws:_
 
-* An [code]#exception# with the [code]#errc::invalid# error code if [code]#b#
-is a sub-buffer.
+* An [code]#exception# with the [code]#errc::invalid# error code if [code]#b# is
+a sub-buffer.
 
 * An [code]#exception# with the [code]#errc::invalid# error code if
 [code]#subRange# corresponds to a a non-contiguous region of [code]#b#.
 
-- An [code]#exception# with the [code]#errc::invalid# error code if the sum
-of [code]#baseIndex# and [code]#subRange# in any dimension exceeds the parent
+* An [code]#exception# with the [code]#errc::invalid# error code if the sum of
+[code]#baseIndex# and [code]#subRange# in any dimension exceeds the parent
 buffer [code]#b# size [code]#bufferRange# in that dimension.
 '''
 
@@ -5446,8 +5461,8 @@ of elements in each dimension as passed to the constructor.
 ----
 size_t byte_size() const noexcept;
 ----
-_Returns:_ The size of the buffer storage in bytes. Equal to
-[code]#size()*sizeof(T)#.
+_Returns:_ The size of the buffer storage in bytes.
+Equal to [code]#size()*sizeof(T)#.
 '''
 
 .[apititle]#buffer::size#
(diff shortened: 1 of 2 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -5190,281 +5190,169 @@ include::{header_dir}/buffer.h[lines=4..-1]
...
+buffer(Container& container, const property_list& propList = {}); (2)
 ----
-   a@ Construct a one dimensional SYCL [code]#buffer# instance
-      from the elements starting at [code]#std::data(container)#
-      and containing [code]#std::size(container)# number of elements.
-      The buffer is initialized with the contents of [code]#container#,
-      and the buffer assumes exclusive access to [code]#container# for
-      the duration of its lifetime.
+Constraints:_ Available only if [code]#Container# must be a contiguous
+container, that is the following requirements must be met:
```
Pennycook, line 5281:
```suggestion
Constraints:_ Available only if [code]#Container# is a contiguous
container, that is the following requirements must be met:
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5474,58 +5362,53 @@ provided when allocating memory on the host.  The range of the constructed SYCL
...
-buffer(InputIterator first, InputIterator last,
-       const property_list& propList = {})
-----
-   a@ Create a new allocated 1D buffer initialized from the given elements
-      ranging from [code]#first# up to one before [code]#last#.
-      The data is copied to an intermediate memory position by the runtime.
-      Data is not written back to the same iterator set provided. However, if the [code]#buffer# has a valid non-const iterator specified via the member function [code]#set_final_data()#, data will be copied back to that iterator.
-      The constructed SYCL [code]#buffer# will use a default constructed [code]#AllocatorT# when allocating memory on the host.
-      Zero or more properties can be provided to the constructed SYCL [code]#buffer# via an instance of [code]#property_list#.
+template <class InputIterator>
```
Pennycook, line 5370: For these functions accepting `InputIterator`, should there be _Constraints_?  They definitely need to be iterators.  Do they need to be contiguous/random access/whatever?  Does the `value_type` need to be convertible to the buffer type?

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
+* An [code]#exception# with the [code]#errc::invalid# error code if [code]#b#
+is a sub-buffer.
 
+* An [code]#exception# with the [code]#errc::invalid# error code if
+[code]#subRange# corresponds to a a non-contiguous region of [code]#b#.
 
+- An [code]#exception# with the [code]#errc::invalid# error code if the sum
+of [code]#baseIndex# and [code]#subRange# in any dimension exceeds the parent
+buffer [code]#b# size [code]#bufferRange# in that dimension.
+'''
```
Pennycook, line 5437: This contradicts the _Preconditions_.  If something is a _Precondition_ then violating it results in undefined behavior.  Each of these conditions should either be a precondition or be a condition in which an exception is thrown.

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
-size_t byte_size() const noexcept
+size_t get_count() const;
 ----
-   a@ Returns the size of the buffer storage in bytes.
-      Equal to [code]#size()*sizeof(T)#.
-a@
-[source]
+Deprecated by SYCL 2020.
+
+_Returns:_ The same value as [code]#size()#.
```
Pennycook, line 5477:
```suggestion
Effects:_ Equivalent to [code]#return size()#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
 ----
-size_t get_size() const
+size_t get_size() const;
 ----
-   a@ Returns the same value as [code]#byte_size()#. Deprecated.
-a@
-[source]
+Deprecated by SYCL 2020.
+
+_Returns:_ The same value as [code]#byte_size()#.
```
Pennycook, line 5487:
```suggestion
_Effects:_ Equivalent to [code]#return byte_size()#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
+<<ranged-accessor>>, where the range starts at the given offset from the
+beginning of the buffer.  The value of target can be [code]#target::device#,
+[code]#target::constant_buffer# or [code]#target::host_task#.
+
+_Returns (3):_ A  valid [code]#accessor# as if constructed via passing the
+buffer and all provided arguments to the [code]#accessor# constructor.
+
+Possible implementation:
+
+[code]#+return accessor{*this, args...};+#
```
Pennycook, line 5531:
```suggestion
```

I don't think we need possible implementations in the spec.

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
+
+_Returns (4):_ A valid [code]#host_accessor# as if constructed via passing the
+buffer and all provided arguments to the [code]#host_accessor# constructor.
+
+Possible implementation:
+
+[code]#+return host_accessor{*this, args...};+#
```
Pennycook, line 5538:
```suggestion
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
-      The buffer object being reinterpreted can be a SYCL sub-buffer
-      that was created from a SYCL [code]#buffer#.  The implementation must
-      throw an [code]#exception# with the [code]#errc::invalid# error code
-      if the total size in bytes represented by this SYCL [code]#buffer# (or
-      sub-buffer) is not evenly divisible by [code]#sizeof(ReinterpretT)#.
-      Reinterpreting a sub-buffer provides a reinterpreted view
-      of the sub-buffer only,
-      and does not change the offset or size of the sub-buffer view (in bytes)
-      relative to the parent [code]#buffer#.
+_Effects:_ Creates a reinterpreted SYCL [code]#buffer#. The buffer object being
```
Pennycook, line 5627: I don't think we should have both an _Effects_ and a _Returns_ for this.  _Returns_ seems to be sufficient.

adoc/chapters/programming_interface.adoc
steffenlarsen: Good question. I think I agree, but I didn't want to change the current descriptions too much. I'll happily remove it if people agree though.

adoc/chapters/programming_interface.adoc
```diff
@@ -5540,358 +5423,318 @@ queue, the sub-buffer's alignment must be compatible with both the primary
...
+unlocked otherwise. The contents of [code]#hostData# are guaranteed to reflect
+the contents of the buffer when the [code]#std::mutex# is unlocked by the
+runtime.
 
-|====
+_Effects (1):_ Constructs a SYCL [code]#use_mutex# property instance with a
+reference to [code]#mutexRef# parameter provided.
 
+_Returns (2):_ The [code]#std::mutex# which was specified when constructing this
+SYCL [code]#use_mutex# property.
```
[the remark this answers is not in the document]
Pennycook: I think you should just treat it like another class, like you did for `buffer`.  There should be a source block showing the whole class, then separate `apidef` blocks for the constructor and any member functions.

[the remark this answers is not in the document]
steffenlarsen: I've split up the two properties with values. Let me know what you think!

adoc/chapters/programming_interface.adoc
steffenlarsen: I think the benefit of having Effects as well is that I could document some of the common (mostly sub-buffer) behavior.

adoc/chapters/programming_interface.adoc
```diff
@@ -5190,708 +5190,597 @@ include::{header_dir}/buffer.h[lines=4..-1]
...
 ----
-property::buffer::use_host_ptr::use_host_ptr()
+use_mutex(std::mutex& mutexRef);
 ----
-   a@ Constructs a SYCL [code]#use_host_ptr# property instance.
 
-a@
-[source]
+_Effects:_ Constructs a SYCL [code]#use_mutex# property instance with a
+reference to [code]#mutexRef# parameter provided.
```
Pennycook, line 5729:
```suggestion
reference to [code]#mutexRef#.
```

We could also consider just calling this something like `m`, but I'm not sure.

adoc/chapters/programming_interface.adoc
```diff
@@ -5190,708 +5190,597 @@ include::{header_dir}/buffer.h[lines=4..-1]
...
+std::mutex* get_mutex_ptr() const;
 ----
-   a@ Constructs a SYCL [code]#use_mutex# property instance with a reference to [code]#mutexRef# parameter provided.
 
-a@
-[source]
-----
-property::buffer::context_bound::context_bound(context boundContext)
+_Returns:_ The [code]#std::mutex# which was specified when constructing this
+SYCL [code]#use_mutex# property.
```
Pennycook, line 5740:
```suggestion
_Returns:_ A pointer to the [code]#std::mutex# provided when constructing this property.
```

Kind of a nit, but I like when we can avoid being verbose.  I considered "provided at construction" too, but I don't know if that's clear enough.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
Pennycook: This looks much better to me.  Thanks!

gmlueck:
Thanks for doing this!

I need to read through this in more depth, but here are some comments from a quick review.  In general, see section 4.6.5 "Queue class" as a model to follow for the new style.

* We no longer have the "interface" sub-section in the new style.  Instead, merge the contents of "Buffer interface" into the introductory paragraphs of the "Buffer" section.
* Use `[source,role=synopsis]` for the top-level synopsis of the `buffer` class.
* Remove the properties from the top-level synopsis.  You have the synopses for these properties under the "Properties" section, and there is no need to duplicate them.
* Pay attention to the callouts in the synopses.  There are cases where they are missing from the synopsis, but you reference them from the text.  The callout should be on the first line of the function definition.  (This is sometimes the line with `template` for a function template.)  Make sure the callouts are aligned in any single synopsis.  The alignment should be such that the callouts are two spaces to the right of the longest line in the synopsis.
* The section name "Buffer properties" should be just "Properties".  Probably the same for "Buffer destruction rules" -> "Destruction rules".
* The horizonal lines are not rendering properly.  I think this is because you need a blank line before and after the `'''`.

steffenlarsen:
Thanks, @gmlueck! I believe I've addressed most of the initial comments, but I'm still a little fuzzy on the placement of the callouts.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,636 @@ If no template argument is provided, then the default allocator for the SYCL
...
-buffer(Container& container,
-       const property_list& propList = {})
-----
-   a@ Construct a one dimensional SYCL [code]#buffer# instance
-      from the elements starting at [code]#std::data(container)#
-      and containing [code]#std::size(container)# number of elements.
-      The buffer is initialized with the contents of [code]#container#,
-      and the buffer assumes exclusive access to [code]#container# for
-      the duration of its lifetime.
+buffer(Container& container, AllocatorT allocator,                 (1)
```
gmlueck, line 5278: > Thanks, @gmlueck! I believe I've addressed most of the initial comments, but I'm still a little fuzzy on the placement of the callouts.

The callouts mostly look good, except they should be moved to the first line of each function definition.  Using this one as an example, move the "(1)" one line earlier, so that it's on the same line as the `template` keyword.  The justification should be the same as what you have now.  For example, this "(1)" starts in column 68.  It should still be in column 68 when you move it one line earlier.

gmlueck (commented):
I did finish reviewing, but here are my comments so far.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,636 @@ If no template argument is provided, then the default allocator for the SYCL
...
+       const property_list& propList = {});
 
-Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+template <typename Container>                                      (2)
+buffer(Container& container, const property_list& propList = {});
+----
+_Constraints:_ Available only if [code]#Container# is a contiguous container,
+that is the following requirements must be met:
```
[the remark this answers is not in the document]
steffenlarsen: This was taken from the comments in `buffer.h`.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,636 @@ If no template argument is provided, then the default allocator for the SYCL
...
 ----
-   a@ Construct a one dimensional SYCL [code]#buffer# instance
-      from the elements starting at [code]#std::data(container)#
-      and containing [code]#std::size(container)# number of elements.
-      The buffer is initialized with the contents of [code]#container#,
-      and the buffer assumes exclusive access to [code]#container# for
-      the duration of its lifetime.
+template <typename Container>                                      (1)
+buffer(Container& container, AllocatorT allocator,
+       const property_list& propList = {});
```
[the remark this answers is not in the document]
steffenlarsen: I agree, it is odd. I was hoping to avoid touching `buffer.h`, but it had it in this order. I have reordered the ctors to make it consistent.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,636 @@ If no template argument is provided, then the default allocator for the SYCL
...
-a@
-[source]
+.[apititle]#Construct from iterators#
+[source,role=synopsis,id=api:buffer-ctor-iterator]
 ----
-template <typename InputIterator>
-buffer(InputIterator first, InputIterator last,
-       AllocatorT allocator = {},
-       const property_list& propList = {})
+template <class InputIterator>                                               (1)
```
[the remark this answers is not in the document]
steffenlarsen: Agreed. It has been changed here and in buffer.h. Do we want to change it for the deduction guides too?

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
gmlueck: I see.  I guess `std::data(container)` is normally defined only for a contiguous container.  Perhaps we should say:

> _Preconditions:_ `container` is a contiguous container.
> _Constraints:_ Available only when:
> * `std::data(container)` and `std::size(container)` are well-formed;
> * The return type of `std::data(container)` is convertible to `T*`; and
> * `Dimensions == 1`.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
gmlueck: Yes.

adoc/chapters/programming_interface.adoc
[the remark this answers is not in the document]
steffenlarsen: Definitely seems more formal. I have applied this, thanks!

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
+provided when allocating memory on the host.
+
+The host address is [code]#const T#, so the host accesses can be read-only.
+However, the [code]#typename T# is not const so the device accesses can be both
+read and write accesses.
+Since, the [code]#hostData# is const, this buffer is only initialized with this
+memory and there is no write back after its destruction, unless the
+[code]#buffer# has another valid non-null final data address specified via the
+member function [code]#set_final_data()# after construction of the
```
Pennycook: I don't think the bit about the host vs device access adds very much here.  We could keep it as non-normative, but I think it's likely to confuse people more than it helps -- we shouldn't have to explain what `const` means.

I think replacing this entire paragraph with the below would be a lot clearer:

```suggestion
_Remarks_: The buffer will not write back to any host memory
unless requested via the member function [code]#set_final_data()#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+elements starting at [code]#std::data(container)# and containing
+[code]#std::size(container)# number of elements.
+The buffer is initialized with the contents of [code]#container#, and the buffer
+assumes exclusive access to [code]#container# for the duration of its lifetime.
 
 Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+[code]#buffer# destruction if the return type of [code]#std::data(container)# is
+not [code]#const#.
```
Pennycook: I think we should turn this into a _Remark_, because it's not really an effect of the constructor.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
-      Zero or more properties can be provided to the constructed SYCL [code]#buffer# via an instance of [code]#property_list#.
+       const property_list& propList = {});
+----
+_Effects (1):_ Create a new allocated 1D buffer initialized from the given
+elements ranging from [code]#first# up to one before [code]#last#.
+The data is copied to an intermediate memory position by the runtime.
+Data is not written back to the same iterator set provided.
+However, if the [code]#buffer# has a valid non-const iterator specified via the
+member function [code]#set_final_data()#, data will be copied back to that
+iterator.
```
Pennycook, line 5383: I don't think the part about the intermediate memory position is helpful.  We don't have this wording for the `hostMem` overload, and the inconsistency is weird.

I think we should re-use the same wording as the `hostMem` constructor, which I proposed changes to above:

```suggestion
_Remarks_: The buffer will not write back to any host memory
unless requested via the member function [code]#set_final_data()#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
 ----
-range<Dimensions> get_range() const
+range<Dimensions> get_range() const;
 ----
-   a@ Return a range object representing the
-      size of the buffer in terms of number
-      of elements in each dimension as passed
-      to the constructor.
+_Returns:_ A range object representing the size of the buffer in terms of number
+of elements in each dimension as passed to the constructor.
```
Pennycook:
```suggestion
_Returns:_ A [code]#range# representing the number of elements in the buffer.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+.[apititle]#buffer::size#
+[source,role=synopsis,id=api:buffer-size]
 ----
-AllocatorT get_allocator() const
+size_t size() const noexcept;
 ----
-   a@ Returns the allocator provided to the buffer.
+mode and target in the command group buffer.
+The value of target can be [code]#target::device#,
+[code]#target::constant_buffer# or [code]#target::host_task#.
```
Pennycook: This looks broken, but I can't tell if it's just GitHub being weird.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+template <typename... Ts> auto get_host_access(Ts...);                         (4)
 ----
-template <typename... Ts> auto get_access(Ts... args)
-----
-   a@ Returns a valid [code]#accessor# as if constructed via passing the buffer
-      and all provided arguments to the [code]#accessor# constructor.
+_Returns (1):_ A valid [code]#accessor# to the buffer with the specified access
+mode and target in the command group buffer.
+The value of target can be [code]#target::device#,
+[code]#target::constant_buffer# or [code]#target::host_task#.
```
Pennycook: Should these be preconditions or constraints?

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+
+_Returns (2):_ A valid [code]#accessor# to the buffer with the specified access
+mode and target in the command group buffer.
+The accessor is a <<ranged-accessor>>, where the range starts at the given
+offset from the beginning of the buffer.
+The value of target can be [code]#target::device#,
+[code]#target::constant_buffer# or [code]#target::host_task#.
```
Pennycook: Same question as above.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+
+_Returns (3):_ A valid [code]#accessor# as if constructed via passing the buffer
+and all provided arguments to the [code]#accessor# constructor.
 
 Possible implementation:
 
 [code]#+return accessor{*this, args...};+#
```
Pennycook: We should remove this.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
-----
-   a@ Returns a valid [code]#host_accessor# as if constructed via passing the
-      buffer and all provided arguments to the [code]#host_accessor#
-      constructor.
+_Returns (4):_ A valid [code]#host_accessor# as if constructed via passing the
+buffer and all provided arguments to the [code]#host_accessor# constructor.
 
 Possible implementation:
 
 [code]#+return host_accessor{*this, args...};+#
```
Pennycook: You can remove this too.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
+Use [code]#get_host_access()# instead.
+
+_Returns (1):_ A valid host [code]#accessor# to the buffer with the specified
+access mode and target.
+
+_Returns (2):_ A valid host [code]#accessor# to the buffer with the specified
+access mode and target.
+The accessor is a <<ranged-accessor>>, where the range starts at the given
+offset from the beginning of the buffer.
+The value of target can only be [code]#target::host_buffer#.
```
Pennycook: Should this be a precondition or a constraint?

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,637 @@ If no template argument is provided, then the default allocator for the SYCL
...
-      throw an [code]#exception# with the [code]#errc::invalid# error code
-      if the total size in bytes represented by this SYCL [code]#buffer# (or
-      sub-buffer) is not evenly divisible by [code]#sizeof(ReinterpretT)#.
-      Reinterpreting a sub-buffer provides a reinterpreted view
-      of the sub-buffer only,
-      and does not change the offset or size of the sub-buffer view (in bytes)
-      relative to the parent [code]#buffer#.
+_Preconditions (2):_ Available when [code]#(ReinterpretDim == 1)# or when
+[code]#\((ReinterpretDim == Dimensions) && (sizeof(ReinterpretT) ==
+sizeof(T)))#.
```
Pennycook: Preconditions are for UB. Conditions that affect availability should be constraints.

Pennycook (commented):
This is looking really good, but I found a few new things in my latest readthrough.

adoc/headers/buffer.h
```diff
@@ -2,29 +2,6 @@
...
-};
-
-class context_bound {
- public:
-  context_bound(context boundContext);
-
-  context get_context() const;
-};
-} // namespace buffer
-} // namespace property
```
keryell, line 26: Is this really to be removed? :thinking:

steffenlarsen: This was moved per @gmlueck's [request](https://github.com/KhronosGroup/SYCL-Docs/pull/824#issuecomment-2901512286):

> Remove the properties from the top-level synopsis. You have the synopses for these properties under the "Properties" section, and there is no need to duplicate them.

Unless I misunderstood the request.

adoc/chapters/programming_interface.adoc
steffenlarsen: Seems like it shouldn't be there at all. It is not a restriction but rather part of the return type, so I don't see why we would explicitly mention it in the description.

adoc/chapters/programming_interface.adoc
steffenlarsen: Remarks seem to be a separate segment of the description, so to avoid having to associate it with the constructors by their callouts, I added a few more words to the beginning.

adoc/chapters/programming_interface.adoc
steffenlarsen: Ah, good catch. This looks like a copy-paste error. It should be better now.

adoc/chapters/programming_interface.adoc
steffenlarsen: Absolutely! I've made it a constraint, though I've tried to associate it with the two valid functions. Is the right callout reference "(1-2)", "(1&2)" or something else?

gmlueck: We use the style "(1) - (2)".

adoc/headers/buffer.h
gmlueck: I think we resolved this in the WG meeting after @keryell added this comment.  The synopses for the properties weren't removed.  They were just moved to a different place in the document.  I don't think anyone has concerns about that.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,624 @@ If no template argument is provided, then the default allocator for the SYCL
...
-----
-template <access_mode Mode>
-accessor<T, Dimensions, Mode, target::host_buffer>
-get_access(range<Dimensions> accessRange, id<Dimensions> accessOffset = {})
+template <typename... Ts> auto get_access(Ts...);                              (3)
+
+template <typename... Ts> auto get_host_access(Ts...);                         (4)
 ----
-   a@ Deprecated in SYCL 2020.  Use [code]#get_host_access()# instead.
+_Constraints (1-2):_ Available only when [code]#Targ# is [code]#target::device#,
```
steffenlarsen:
```suggestion
_Constraints (1)-(2):_ Available only when [code]#Targ# is [code]#target::device#,
```

adoc/chapters/programming_interface.adoc
steffenlarsen: Great, thanks! That has been addressed.

adoc/chapters/programming_interface.adoc
```diff
@@ -5164,19 +5164,6 @@ still undefined behavior.
...
-==== Buffer interface
-
-The constructors and member functions of the SYCL [code]#buffer# class template
-are listed in <<table.constructors.buffer>> and <<table.members.buffer>>,
-respectively.
-The additional common special member functions and common member functions are
-listed in <<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
-
-Each constructor takes as the last parameter an optional SYCL
```
tomdeakin, line 5177: This text has vanished entirely. There are words later which say much the same thing though.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
-The host address is [code]#const T#, so the host accesses
-can be read-only. However, the [code]#typename T# is not const so
-the device accesses can be both read and write accesses. Since,
-the [code]#hostData# is const, this buffer is only initialized
-with this memory and there is no write back after its
-destruction, unless the [code]#buffer# has another valid
-non-null final data address specified via the member function
-[code]#set_final_data()# after construction of the
-[code]#buffer#.
+_Effects (3):_ Construct a SYCL [code]#buffer# instance with the
```
tomdeakin, line 5246: When this is rendered, the new lines make it difficult to associate all the sentences with _Effects (3)_.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
 
-The constructed SYCL [code]#buffer# will use the
-[code]#allocator# parameter provided when allocating
-memory on the host.
+_Effects (2):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
 
-The host address is [code]#const T#, so the host accesses
-can be read-only. However, the [code]#typename T# is not const so
-the device accesses can be both read and write accesses. Since,
```
tomdeakin, line 5308: This text has gone missing in Effects (3).

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
-[source]
-----
-template <typename Container>
-buffer(Container& container,
-       const property_list& propList = {})
+_Effects (4):_ Equivalent to `buffer(hostData, bufferRange, AllocatorT{},
+propList)`.
+
+_Remarks:_ When [code]#hostData# is a pointer to a const-qualified type, the
+buffer will not write back to any host memory unless requested via the member
```
tomdeakin, line 5261: The original wording made it clear that `set_final_data` needed to be called with a non-const pointer different to `hostData`.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
+buffer(Container& container, AllocatorT allocator,
+       const property_list& propList = {});
 
-Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+template <typename Container>                                      (2)
+buffer(Container& container, const property_list& propList = {});
+----
+Preconditions: [code]#container# is a contiguous container.
```
tomdeakin: Is this a new precondition?

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
+elements starting at [code]#std::data(container)# and containing
+[code]#std::size(container)# number of elements.
+The buffer is initialized with the contents of [code]#container#, and the buffer
+assumes exclusive access to [code]#container# for the duration of its lifetime.
 
-Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+The constructed SYCL [code]#buffer# will use the [code]#allocator# parameter
+provided when allocating memory on the host.
```
tomdeakin, line 5293: Can you remove empty lines - the rendering makes them hard to associate with the correct _Effect_

adoc/headers/buffer.h
tomdeakin: That is my memory of the discussion too. I've just read through this section, and they are in the main text.

adoc/chapters/programming_interface.adoc
```diff
@@ -5164,19 +5164,6 @@ still undefined behavior.
 The SYCL [code]#buffer# class template provides the common reference semantics
 (see <<sec:reference-semantics>>).
 
-
-==== Buffer interface
```
tomdeakin, line 5168: This causes a change to the section numbering. E.g., 4.7.2.2 Buffer Properties would now be 3.7.2.3. The first-level is changed because Acknowledgements has been moved to an unnumbered section to the end, and I think that won't happen if we apply this to SYCL 2020, but it would still cause a new numbering in that final level.

tomdeakin (changes requested):

adoc/chapters/programming_interface.adoc
steffenlarsen: Under the "Constructors" section a little further down we have:

> All [code]#buffer# constructors take a parameter named [code]#propList# which
> allows the application to pass zero or more properties.
> These properties may specify additional effects of the constructor and resulting
> [code]#buffer# object.
> See <<sec:buffer-properties>> for the buffer properties that are defined by the
> <<core-spec>>.

That uses similar wording to `queue`. I figured that was appropriate, but I can add this back if you think it would be better.

adoc/chapters/programming_interface.adoc
steffenlarsen: This was reworded as per the discussion in https://github.com/KhronosGroup/SYCL-Docs/pull/824#discussion_r2166261451.

adoc/chapters/programming_interface.adoc
steffenlarsen: This was mostly intended as a more formal definition of what was meant. See discussion: https://github.com/KhronosGroup/SYCL-Docs/pull/824#discussion_r2112760213

adoc/chapters/programming_interface.adoc
steffenlarsen: Sure! I am just worried that one big block of text will be a lot to consume. What do other people think?

adoc/chapters/programming_interface.adoc
steffenlarsen: Ooof, good point. @gmlueck - Thoughts on this numbering change? Will it be an issue?

gmlueck: I think this is OK.  All of the other "restructuring" PRs we've done have also made similar changes to the section numbering.

adoc/chapters/programming_interface.adoc
gmlueck: I agree with @tomdeakin.  I think it reads better without the blank line.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,625 @@ If no template argument is provided, then the default allocator for the SYCL
...
-Data is written back to [code]#container# before the completion of
-[code]#buffer# destruction if the return type of [code]#std::data(container)#
-is not [code]#const#.
+template <typename Container>                                      (2)
+buffer(Container& container, const property_list& propList = {});
+----
+Preconditions: [code]#container# is a contiguous container.
 
-The constructed SYCL [code]#buffer# will use a default constructed [code]#AllocatorT# when allocating memory on the host.
+Constraints: Available only when:
```
```suggestion
_Constraints:_ Available only when:
```

adoc/chapters/programming_interface.adoc
```suggestion
_Preconditions:_ [code]#container# is a contiguous container.
```

adoc/chapters/programming_interface.adoc
steffenlarsen: That's good enough for me! I've removed the spacing in these cases.

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
-the [code]#hostData# is const, this buffer is only initialized with
-this memory and there is no write back after its destruction,
-unless the [code]#buffer# has another valid non-null final
-data address specified via the member function
-[code]#set_final_data()# after construction of the
-[code]#buffer#.
+.[apititle]#Construct with host memory#
+[source,role=synopsis,id=api:buffer-ctor-host-mem]
+----
+buffer(T* hostData, const range<Dimensions>& bufferRange,          (2)
```
```suggestion
buffer(T* hostData, const range<Dimensions>& bufferRange,          (1)
```

This is the first overload in this synopsis, so it should be `(1)`.  The one below is currently `(1)`, and that should be changed to `(2)`.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
-queue's device and the secondary queue's device, otherwise this exception is
-thrown.
+* An [code]#exception# with the [code]#errc::invalid# error code if
+[code]#subRange# corresponds to a a non-contiguous region of [code]#b#.
 
-Must throw an [code]#exception# with the [code]#errc::invalid# error code if
-[code]#b# is a sub-buffer.
+* An [code]#exception# with the [code]#errc::invalid# error code if the sum of
+[code]#baseIndex# and [code]#subRange# in any dimension exceeds the parent
+buffer [code]#b# size [code]#bufferRange# in that dimension.
```
```suggestion
buffer [code]#b# size ([code]#bufferRange#) in that dimension.
```

The original text had these parens, and I think they help make this sentence somewhat more understandable.  I think we could probably reword this sentence a bit, but let's save that for a future PR.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
-[code]#info::device::mem_base_addr_align#.  Violating this requirement causes
-the implementation to throw an [code]#exception# with the [code]#errc::invalid#
-error code from the [code]#accessor# constructor (if the accessor is not a
-placeholder) or from [code]#handler::require()# (if the accessor is a
-placeholder).  If the accessor is bound to a <<command-group>> with a secondary
-queue, the sub-buffer's alignment must be compatible with both the primary
-queue's device and the secondary queue's device, otherwise this exception is
-thrown.
+* An [code]#exception# with the [code]#errc::invalid# error code if
+[code]#subRange# corresponds to a a non-contiguous region of [code]#b#.
```
```suggestion
the subrange formed by [code]#baseIndex# and [code]#subRange# is not a contiguous region of [code]#b#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
-      and an [code]#exception# with the [code]#errc::invalid# error code must be thrown if violated.
+       const range<Dimensions>& subRange);
+----
+_Effects:_ Create a new sub-buffer without allocation to have separate accessors
+later.
+[code]#b# is the buffer with the real data.
+[code]#baseIndex# specifies the origin of the sub-buffer inside the buffer
+[code]#b#.
+[code]#subRange# specifies the size of the sub-buffer.
+The origin (based on [code]#baseIndex#) of the sub-buffer being constructed must
```
gmlueck, line 5409: It's OK to have multiple paragraphs in the _Effects_ section.  I think it would make sense to have this sentence start a new paragraph because the next several sentences are all about this restriction.

Note that the original text also had these sentences in their own paragraph.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
-   a@ The [code]#finalData# points to where the outcome of all
-      the buffer processing is going to be copied to at destruction
-      time, if the buffer was involved with a write accessor.
+Deprecated in SYCL 2020.
+Use [code]#get_host_access()# instead.
 
-Destination can be either an output iterator or a
-[code]#std::weak_ptr<T>#.
+_Returns (1):_ A valid host [code]#accessor# to the buffer with the specified
+access mode and target.
```
```suggestion
access mode.
```

I know the original text had "and target", but it makes no sense.  There is no `target` parameter because this is returning a host accessor.

adoc/chapters/programming_interface.adoc
```diff
@@ -5186,718 +5173,609 @@ If no template argument is provided, then the default allocator for the SYCL
...
 
-Note that a raw pointer is a special case of output iterator and
-thus defines the host memory to which the result is to be
-copied.
+_Returns (2):_ A valid host [code]#accessor# to the buffer with the specified
+access mode and target.
```
```suggestion
access mode.
```

Here too.

gmlueck (commented):
Very nice!  Sorry for the delayed review.  I just have a few small comments below.

adoc/chapters/programming_interface.adoc
steffenlarsen: Seems like a mistake too. The ordering in the description correspond to your suggested change.

gmlueck (approved):

adoc/chapters/programming_interface.adoc
tomdeakin: Looks fine, thanks for clarifying.

tomdeakin (approved):

PR #827, opened 2025-05-29
Clarify specializing sycl:: templates is valid

### LINKED ISSUE #341: User-defined function objects
In the description of SYCL function objects in [4.17.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:function-objects) there is no mention that user cannot define them for any types not covered by the implementation itself. IMHO, it should be clarified whether it is allowed or not.

Pennycook:
This is aligned with the behavior of `std::` in ISO C++, and is necessary for developers to be able to define specializations of things like `sycl::is_device_copyable` and `sycl::minimum`.

The wording allows the specification to add restrictions in specific cases, if defining a specialization for a user-defined class would break something.

I also added a note to the section about function objects like `std::minimum`, because this change was motivated by an issue raised specifically about that section.  Closes #341.


adoc/chapters/programming_interface.adoc
```diff
@@ -116,6 +116,9 @@ Unless otherwise specified, the behavior of a SYCL program is undefined if it
 adds any entity to namespace [code]#sycl# or to a namespace within namespace
 [code]#sycl#.
 
+Unless explicitly prohibited, a SYCL program may add a template specialization
+for class templates defined in namespace [code]#sycl# if the specialization
+depends on at least one program-defined type.
```
gmlueck, line 121: Do we list any prohibitions elsewhere in the spec?  If not, this paragraph allows the application to specialize *any* class template, so long as the specialization is for a type that the application defines.  Do we want that?  It seems like it would be safer to say that specialization is disallowed unless the spec specifically says it is OK.

For example, do we want applications to be able to specialize `sycl::is_group<T>` for their own types?  I think a strict reading of the spec (after this PR) would allow an application to specialize `sycl::is_group<T>` on their own type UDT and then call any of the group functions (e.g. `sycl::group_broadcast`) with UDT with the expectation that it would work.  However, that can't possibly work because the implementation of `sycl::broadcast` would need to know about UDT.

It seems like it would be better if the spec did not have a blank statement allowing template specialization.  Instead, it seems like the spec should specifically say which class templates can be specialized, and the spec should also clearly state what the effect of specializing the class template is.

Pennycook: I like following the C++ precedent here, because it seems less surprising to C++ developers.  C++ has some additional wording to the effect of "the specialization must be in line with everything else the specification says about the template" -- would you be more comfortable with this if we added something similar?

Whether we stick with my initial proposed wording or negate it, I agree we'll need to add new wording to specific classes.  I've taken a stab at allow and deny lists below -- did I miss anything?

## Allow
- `known_identity`
- `has_known_identity`
- `atomic_ref`
- `atomic`
- `plus`
- `multiplies`
- `bit_and`
- `bit_or`
- `bit_xor`
- `logical_and`
- `logical_or`
- `minimum`
- `maximum`

## Deny
- `is_property`
- `buffer`
- `sampled_image`
- `unsampled_image`
- `accessor`
- `host_accessor`
- `local_accessor`
- `sampled_image_accessor`
- `unsampled_image_accessor`
- `host_sampled_image_accessor`
- `host_unsampled_image_accessor`
- `multi_ptr`
- `usm_allocator`
- `specialization_id`
- `is_group`

## Not Sure
- `private_memory` - This seems fine.  I don't really care what we do, though, given that it's part of hierarchical parallelism.
- `vec` - If a user wants to define a specialization for a user-defined type that somehow sits on top of `vec`, that seems reasonable to me.
- `marray` - This doesn't seem dangerous.  Probably pointless, but not dangerous.

gmlueck: WG discussion today: Wording should say that only specific classes can be specialized.  For now, let's just allow `know_identity`, `has_known_identity`, and `is_device_copyable`.

Pennycook: I've made the changes in https://github.com/KhronosGroup/SYCL-Docs/pull/827/commits/ca0751a5bb3dd16af06954582c31f1771096e5ce.

I thought it would be confusing to have two commits back to back where one inverts wording introduced by the other, so I purged the old version from the history.

There was already wording saying that `is_device_copyable` could be specialized, and I left that as it was.

adoc/chapters/programming_interface.adoc
```diff
@@ -21660,6 +21666,12 @@ All function objects obey {cpp} conversion and promotion rules.
 Each function object is additionally specialized for [code]#void# as a
 _transparent_ function object that deduces its parameter types and return type.
 
+{note}Using these function objects with program-defined types requires those
+types to define the associated operator (e.g., [code]#plus# requires
+[code]#pass:[+]#).
```
etomzak: > e.g., `plus` requires `+`

Is `+` the same as `operator+()`?

gmlueck: Agreed.  Probably better to say `operator+` here.

Pennycook: Fixed in 8c5e0636041f0b4c2bd924c7e96ed038c1066c29.

adoc/chapters/programming_interface.adoc
```diff
@@ -21660,6 +21666,12 @@ All function objects obey {cpp} conversion and promotion rules.
 Each function object is additionally specialized for [code]#void# as a
 _transparent_ function object that deduces its parameter types and return type.
 
+{note}Using these function objects with program-defined types requires those
+types to define the associated operator (e.g., [code]#plus# requires
+[code]#operatorpass:[+]#).
```
```suggestion
[code]#pass:[operator+]#).
```

What you have seems to work, but it seems surprising, and I wonder if it is a bug in Asciidoctor.  I'd suggest this instsead.

Or, this also seems to work:

```suggestion
[code]#+operator++#).
```

gmlueck (approved):

keryell (approved):
That makes sense.

PR #830, opened 2025-05-30
Clarify value of info::device::vendor_id

### LINKED ISSUE #166: Clarify info::device::vendor_id uniqueness requirements
Is this supposed to be an id for the vendor, or and id for the device specified by the vendor/backend? I am looking for a unique identifier per device, which I can use to verify that my MPI rank -> GPU assignment is working as I expect on multi-gpu machines. The description of vendor_id in the docs, "Returns a unique vendor device identifier.", is ambiguous. I was hoping it would do what I want, and at least be unique per device per machine. After testing with oneAPI (seems to always return 0x8086) and looking at the hipSYCL source, e.g. (https://github.com/illuhad/hipSYCL/blob/develop/src/runtime/cuda/cuda_hardware_manager.cpp#L319), it looks like this is interpreted as a vendor id, not a device id.

When using CUDA and HIP directly, I generally use the PCI bus id encoded as a uint32_t for this purpose.

The behavior that would be useful to me:
```
Returns a unique device identifier provided by the vendor. Guaranteed to be unique across backends and physical devices and subdevices.
```
If uniqueness across backends is challenging, at least uniqueness across devices on a fixed backend would be helpful.

Pennycook:
This device descriptor corresponds to CL_DEVICE_VENDOR_ID. The description in OpenCL is clear that this must return a unique identifier for the device vendor, and also describes how these identifiers are synchronized across APIs.

Closes #166.

---

Although I usually argue in favor of removing functionality with ties to OpenCL, I think this is a special case. The vendor ID here is not unique to OpenCL, but is rather a Khronos-wide concept; since SYCL is a Khronos API, it would be weird for this vendor ID to mean anything other than what it means everywhere else.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1888,6 +1888,15 @@ _Remarks:_ Template parameter to [api]#device::get_info#.
 
 _Returns:_ A unique vendor device identifier.
 
+If the vendor has a PCI vendor ID, the low 16 bits must contain that PCI vendor
+ID, and the remaining bits must be set to zero.
+Otherwise, the value returned must be a valid Khronos vendor ID.
+Khronos vendor IDs are allocated starting at 0x10000, to distinguish them from
+the PCI vendor ID namespace.
+
+{note}Khronos vendor IDs are synchronized across APIs by utilizing Vulkan’s
+vk.xml as the central Khronos vendor ID registry.{endnote}
+
 '''
 
 .[apidef]#info::device::max_compute_units#


TApplencourt:
Agree, if it's a full Khronos concept, we should keep it!

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -1888,6 +1888,15 @@ _Remarks:_ Template parameter to [api]#device::get_info#.
...
 _Returns:_ A unique vendor device identifier.
 
+If the vendor has a PCI vendor ID, the low 16 bits must contain that PCI vendor
+ID, and the remaining bits must be set to zero.
+Otherwise, the value returned must be a valid Khronos vendor ID.
+Khronos vendor IDs are allocated starting at 0x10000, to distinguish them from
+the PCI vendor ID namespace.
+
+{note}Khronos vendor IDs are synchronized across APIs by utilizing Vulkan’s
+vk.xml as the central Khronos vendor ID registry.{endnote}
```
gmlueck, line 1898: If a vendor decides to support SYCL on a device that is not supported by Vulkan, would that vendor still update the "vk.xml" file?  I assume the answer would have to be "yes"?

Pennycook: Not necessarily.  Vendors have two options:
1. Get a PCI vendor ID (see https://pcisig.com/membership/member-companies); or
2. Get a Khronos vendor ID, via the vk.xml file.

Option 1. is much more likely -- there are only currently [eight vendors with Khronos vendor IDs](https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/refs/heads/main/xml/vk.xml).  I don't think supporting Vulkan is a requirement, rather that Khronos has decided to use Vulkan's XML file as the source of truth.  It doesn't even look like you have to have a "real" physical device: PoCL is an OpenCL implementation, and they have a Khronos vendor ID.

etomzak:
LGTM

tomdeakin (approved):

tomdeakin:
WG approved to merge for SYCL Next

PR #831, opened 2025-06-02
Clarify thread-safety guarantees

### LINKED ISSUE #415: Thread safety guarantees.
From https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#chapter:sycl-programming-interface

> SYCL guarantees that all the member functions and special member functions of the SYCL classes described are thread safe.

Some functionality is exposed as free functions or function templates i.e. not member functions or special member functions. Does SYCL guarantee thread safety? If so, should we make the wording more clear/precise?

Pennycook:
Previously, the thread-safety guarantee only applied to member functions of SYCL classes and did not apply to free functions, function templates, etc. This change clarifies that the thread-safety guarantee applies to all functions.

A new non-normative note further clarifies that this guarantee is not intended to eliminate data races from SYCL programs entirely.

Closes #415.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11,8 +11,11 @@ available for any <<backend, SYCL backends>>, with the exception of the
 interoperability interface, which is described in general terms in this
 document, not pertaining to any particular <<backend>>.
 
-SYCL guarantees that all the member functions and special member functions of
-the SYCL classes described are thread safe.
+All functions defined in this specification are thread-safe, unless otherwise
+specified.
+{note}If a function returns a pointer or reference, potentially concurrent
+conflicting actions via that pointer or reference still constitute a data
+race.{endnote}
 
 The underlying types for all enumerations defined in this specification are
 implementation-defined.


TApplencourt (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -11,8 +11,11 @@ available for any <<backend, SYCL backends>>, with the exception of the
 interoperability interface, which is described in general terms in this
 document, not pertaining to any particular <<backend>>.
 
-SYCL guarantees that all the member functions and special member functions of
-the SYCL classes described are thread safe.
+All functions defined in this specification are thread-safe, unless otherwise
+specified.
+{note}If a function returns a pointer or reference, potentially concurrent
+conflicting actions via that pointer or reference still constitute a data
+race.{endnote}
```
gmlueck: I'm not sure this non-normative note adds more clarity.  I wonder if some readers might think that APIs returning a SYCL object (e.g. `get_context`) are somehow not thread safe?  I know that's not a pointer or reference, but the distinction is a bit subtle.

Do we really think that when an API returns a `T*` or `T&` that people might think the returned pointer/reference is automatically atomic when dereferenced?

Pennycook: > I'm not sure this non-normative note adds more clarity. I wonder if some readers might think that APIs returning a SYCL object (e.g. `get_context`) are somehow not thread safe? I know that's not a pointer or reference, but the distinction is a bit subtle.

I see what you mean.  The fact that `context` has "reference semantics" probably doesn't help, either.

> Do we really think that when an API returns a T* or T& that people might think the returned pointer/reference is automatically atomic when dereferenced?

I'm worried about it, at least.   I'm a bit uncomfortable with there being such a blanket statement about thread-safety, but I went with it because that's what had been proposed in the linked issue. 

Here's an example of the sort of thing I'm worried about:

```c++
accessor a = ...;
a[0] += 1;
```

- `a[0]` is a function call (so it's "thread-safe", by default);
- `+=` is a function call (so it's "thread-safe", by default).

We say work-items are equivalent to threads of execution, so I think a reader could reach the conclusion that the above code is "thread-safe".  Perhaps part of the problem is that we're relying on "thread-safe", which doesn't appear to be defined anywhere in SYCL or ISO C++.

What are we really trying to say here?  Would it be clearer to say something verbose, like:

> Implementations must guarantee that it is safe for multiple host threads to call member functions and free functions that operate on the same instance of a `buffer`, `context`, `device`, `device_image`, `event`, `kernel`, `kernel_id`, `kernel_bundle`, `platform`, `queue`, `sampled_image`, or `unsampled_image`.
>
> [_Note_: This guarantee allows multiple host threads in a SYCL program to submit work to a single `device`, using a single `queue`, without burdening a SYCL developer with the need to guard against data races. -- _end note_]
>
> Implementations must guarantee that it is safe for multiple threads (or threads of execution) to read data from SYCL objects concurrently. Potentially concurrent conflicting actions from multiple threads (or threads of execution) acting on any SYCL object not listed above result in a data race.

Something like that would make a lot more sense to me, and it would be clear that developers need to pay very close attention to anything in device code where work-items attempt to share data.

gmlueck: > Perhaps part of the problem is that we're relying on "thread-safe", which doesn't appear to be defined anywhere in SYCL or ISO C++.

Yes, I also think this is not precise.  I also think we do not really intend for all of the functions in the spec to be thread-safe.  Even limiting this to host code is probably not right.  For example, do we expect `marray::operator++()` to be atomic when executed on the host?  Presumably not.

I think the right fix is to make this blanket statement more precise like:

> Unless otherwise specified, all functions in this specification are "thread safe".  Specifically, this means that multiple threads may execute SYCL host functions simultaneously, and multiple work-items may execute SYCL device functions simultaneously.  The actions performed by any one function call appear to occur atomically with respect to the actions performed by function calls in other threads or work-items.

And then I think we need to go through the spec and examine each function to decide if we really intend this guarantee on a case by case basis.

Pennycook: I agree we should examine each function to see if thread-safety applies on a case by case basis.

My gut says that thread safety is going to be the exception rather than the norm, though.  If the list of classes that guarantee thread-safety is smaller than the list that don't, do you agree the blanket statement should be reversed?

---

Aside: I like your wording about functions occurring atomically, but it made me realize that there might be some visible synchronization effects that we need to think through.  All threads have to agree on the order in which work was submitted to the queue, for example...

gmlueck: > My gut says that thread safety is going to be the exception rather than the norm, though.

My gut feeling is the opposite.  Most functions in SYCL only read information, and I expect those will all be thread safe.

gmlueck: > All threads have to agree on the order in which work was submitted to the queue, for example..

I think we could say this in a general way in the blanket statement.  Something like:

> The order in which these actions are performed is indeterminant, but all threads or work-items will observer the same order.

Pennycook: > My gut feeling is the opposite. Most functions in SYCL only read information, and I expect those will all be thread safe.

Even if a function only reads information, we still have to consider if it's possible for another thread to overwrite the information being read.   I don't think it will be sufficient to look at functions in isolation.

I'll start looking through the specification and draw up a shortlist to discuss here.

Pennycook: Okay, there are a _lot_ more functions than I realized.  There are quite a few that I think we'd both agree meet the intent of what we've been calling "thread-safe": everything in `platform`, `context`, `device`, `queue`, `event` and `buffer`; the USM allocation functions; everything in `handler`, `interop_handle`, `kernel_bundle`, `kernel_id`, `kernel`, `device_image` and `stream`...

But I still think there are going to be too many cases to label as "not thread-safe", and I'm not sure it would even make sense to a reader if we did it.

As a simple example of why this could be complicated (or of how I'm overthinking things), consider just these three functions in `marray`:
```c++
marray<bool, NumElements> operator==(const marray& lhs, const DataT& rhs); (1)
marray operator+(const marray& lhs, const marray& rhs);                    (2)
marray& operator++();                                                      (3)

```

**EDIT**: I mistakenly wrote `operator +=` for (2), before.

It's safe for two threads to both call (1), both call (2), or call (1) and (2) concurrently with the same `marray` objects.  It's unsafe for two threads to call (1) and (3), (2) and (3), or both call (3).  I don't think that this behavior is surprising, and it's what I'd expect given normal C++ rules -- concurrent reads are fine, but concurrent potentially conflicting actions are a data race.

Your proposal requires us to decide whether each of (1), (2) and (3) are "thread-safe".  How would you label them?

gmlueck: I don't think it will be as simple as adding a "not thread safe" label to the relevant functions.  I think we will need to describe the limitations in words.  For `marray`, we might add a new paragraph in the introductory section of the `marray` class.

For `marray`, why is it safe for two threads to call (2) simultaneously?  Wouldn't the `+` part of that function need to be atomic?  It also seems unsafe for threads to call (1) and (2) simultaneously because the read in `==` might conflict with a non-atomic update in `+`.

Pennycook: > I don't think it will be as simple as adding a "not thread safe" label to the relevant functions. I think we will need to describe the limitations in words. For `marray`, we might add a new paragraph in the introductory section of the `marray` class.

Okay.  ISO C++ doesn't have anything like this (as far as I'm aware) so we're going to have to invent it.

> For `marray`, why is it safe for two threads to call (2) simultaneously? Wouldn't the `+` part of that function need to be atomic? It also seems unsafe for threads to call (1) and (2) simultaneously because the read in `==` might conflict with a non-atomic update in `+`.

Because I made a copy-paste error!  (2) was supposed to be just `operator+` -- it would be safe for two threads to call it simultaneously because they're only reading the `marray`.  (3) would then be the only operator in the list that modifies an `marray` and, as you noted, would use some non-atomic updates.

etomzak: > Implementations must guarantee that it is safe for multiple host threads to call member functions and free functions that operate on the same instance of a `buffer`, `context`, ...

FWIW, I'm in favor of something like this.

I'd argue that the sentence that this PR is modifying is narrow in scope and it's OK for it to be narrow in scope. I.e., what this sentence is trying to tell us is that when SYCL "manages" [1] certain resources on behalf of the user, and SYCL provides "handles"[2] to those resources, it is well-defined for the user to interact with those "handles" from more than one host thread. I don't think this sentence needs to say anything more than that, so it's just a case of coming up with the wording.

What exactly thread-safety means for device code is an interesting question, but the spec already has sections on barriers and such, so I think device code is out of scope for this PR.

SYCL provides utilities, like `marray` to help the user manage the user's data, but this is again different from the SYCL runtime "managing" implementation-defined (backend) resources on behalf of the user, so I think it's also out of scope for this PR.

[1] [3.7.1.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_backend_resources_managed_by_the_sycl_application) has a list of resources that SYCL "manages," though it seems to be missing `buffer` and friends
[2] The common reference semantics types work like handles (except maybe accessors, which helps explain why we've also previously questioned if accessors are really common reference types)

TL;DR: This sentence should really just be about the host-side thread-safety of common reference classes that SYCL uses to manage implementation-defined resources (that are not visible to the user, such as backend resources)

adoc/chapters/programming_interface.adoc
```diff
@@ -11,8 +11,11 @@ available for any <<backend, SYCL backends>>, with the exception of the
 interoperability interface, which is described in general terms in this
 document, not pertaining to any particular <<backend>>.
 
-SYCL guarantees that all the member functions and special member functions of
-the SYCL classes described are thread safe.
+All functions defined in this specification are thread-safe, unless otherwise
+specified.
+{note}If a function returns a pointer or reference, potentially concurrent
```
gmlueck: Need a blank line before the note to put it in its own paragraph.

gmlueck (commented):
It seems like there are bigger issues being discussed in the comments, but I added a small note about the change you have now.

adoc/chapters/programming_interface.adoc
Pennycook: Thanks.  Fixed in 9f12acd9c9d42ed6b63e3ad7208c73a2d5843e40.

(Aside: It would be nice if `reflow.py` could detect and fix these things. Tagging @TApplencourt for awareness.)

tomdeakin:
@TApplencourt

TApplencourt:
Agree with greg. We can remove the note, and then if some people have the stamina go over all the API and put exception.

adoc/chapters/programming_interface.adoc
```diff
@@ -11,8 +11,12 @@ available for any <<backend, SYCL backends>>, with the exception of the
...
 document, not pertaining to any particular <<backend>>.
 
-SYCL guarantees that all the member functions and special member functions of
-the SYCL classes described are thread safe.
+All functions defined in this specification are thread-safe, unless otherwise
+specified.
+
+{note}If a function returns a pointer or reference, potentially concurrent
+conflicting actions via that pointer or reference still constitute a data
+race.{endnote}
```
tomdeakin:
```suggestion

```

adoc/chapters/programming_interface.adoc
```diff
@@ -11,8 +11,9 @@ available for any <<backend, SYCL backends>>, with the exception of the
 interoperability interface, which is described in general terms in this
 document, not pertaining to any particular <<backend>>.
 
-SYCL guarantees that all the member functions and special member functions of
-the SYCL classes described are thread safe.
+All functions defined in this specification are thread-safe, unless otherwise
+specified.
+
```
keryell, line 16:
```suggestion
```

gmlueck: Thanks.  Fixed in ffe9519aa68d2c7ac2243e58f56ce4bb4997914f.

PR #832, opened 2025-06-02
Clarify availability of sycl::byte/half on host

### LINKED ISSUE #436: clarify sycl::half support in non-device code
This code:
```
#include <iostream>

#include <sycl/sycl.hpp>

int main(int argc, char **argv) {
  sycl::half a=1, b=1, c=0;
  c = a + b;
  std::cout << "c = " << c << std::endl;
}
```
works on by Intel oneAPI 2023.1.0 and recent OpenSYCL:
```
icpx -fsycl -O2 -g -o half half.cxx
syclcc --opensycl-targets=omp -O2 -g -o half half.cxx
``` 
but I don't see anything in the spec about this, so seems like it's undefined behavior? I have not tried with CUDA and HIP opensycl targets, I imagine that may be more hairy.

I think it would be great if host use of sycl::half was part of the standard - being able to use a type on both host and device makes it much easier to work with. I think having it fallback to float32 conversion on hosts that don't have dedicated float16 instructions is reasonable, and better than having it be UB.

Pennycook:
Existing implementations already provide implementations of these classes, and developers probably rely on that. A std:: equivalent of byte is already available in C++17, and std::float16_t exists in C++23.

Closes #436.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17876,6 +17876,7 @@ Note these types are fundamental and therefore do not exist within the
 
 Additional scalar data types which are supported by SYCL within the [code]#sycl#
 namespace are described in <<table.types.additional>>.
+These data types are available on the host and in device code.
 
 
 [[table.types.additional]]


TApplencourt:
(but let reall implementer veto this one :))

adoc/chapters/programming_interface.adoc
```diff
@@ -17876,6 +17876,7 @@ Note these types are fundamental and therefore do not exist within the
 
 Additional scalar data types which are supported by SYCL within the [code]#sycl#
 namespace are described in <<table.types.additional>>.
+These data types are available on the host and in device code.
```
etomzak, line 17879: Is an implementation required to support `half` in host code if none of the supported devices support `aspect::fp16`?

Pennycook: My reading of the specification is "Yes".  I don't think there's anything in the specification that suggests `half` isn't available in host code, and there's precedent for the host being expected to support things that devices don't (e.g., `double`).

`half` is very poorly specified, so implementing it in host code is straightforward.  Just defining `half` as a wrapper around `float` would be sufficient, because the specification doesn't provide any guarantees about the precision or behavior of `half`.

etomzak: That's how I was reading it as well. I don't immediately see a problem with an implementation doing something like `using half = float` if it doesn't already have `half`. But I don't know enough about the topic to say that with confidence.

Might be good for a few more knowledgeable people to weigh in.

gmlueck: You cannot just do `using half = float` because the spec requires it to be a 16-bit data type with the format specified in IEEE 754.  See:

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_scalar_data_types

I think @Pennycook meant that you could define a C++ class that stores data in this format, but performs all of the operations by up-converting to `float` and then down-converting to the 16-bit type.

etomzak: :thinking: In that case, it does seem a little strange to force implementations to provide `half` on host even if they don't want to provide it on device. But I don't have a strong opinion either way.

AerialMantis (approved):

tomdeakin:

PR #833, opened 2025-06-02
Deprecate the `half_precision::` namespace

Pennycook:
The functions in the half_precision namespace correspond to math functions with the half_ prefix in OpenCL. In OpenCL (and SYCL 1.2.1), the functions have a guaranteed minimum precision, but in SYCL 2020 the precision is backend-specific and non-portable.

Deprecating these functions in favor of using real half precision (via `sycl::half` or `std::float16_t`) will make SYCL simpler.

---

I deliberately didn't update the "what changed" section or assign a milestone, because I wanted to collect some feedback first. 
There is a precedent for deprecating features in SYCL 2020 that should have been deprecated as part of the move away from OpenCL -- should we do that here, or can we only deprecate in SYCL-Next?

My personal opinion: we should deprecate in SYCL 2020, so we can remove in SYCL-Next.  As far as I can tell, these functions are basically pointless, and implementations are free to implement them by just using 32-bit precision.  If a developer violates the precondition (of being in the range of a `half`), many implementations will continue to work.  Getting people to use `float` or `half` depending on what the device actually supports would be a much better user experience.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -25784,7 +25784,7 @@ The return type is [code]#NonScalar# unless [code]#NonScalar# is the
 '''
 
 
-=== Half precision math functions
+=== Half precision math functions (deprecated)
 
 This section describes the half precision math functions that are available in
 the [code]#sycl::half_precision# namespace in both host and device code.
@@ -25806,9 +25806,9 @@ description.
 .[apidef]#half_precision::cos#
 [source,role=synopsis,id=api:half-cos]
 ----
-float cos(float x)                (1)
+float cos(float x)                (1) /* deprecated */
 
-template<typename NonScalar>      (2)
+template<typename NonScalar>      (2) /* deprecated */
 /*return-type*/ cos(NonScalar x)
 ----
 
@@ -25840,9 +25840,9 @@ The return type is [code]#NonScalar# unless [code]#NonScalar# is the
 .[apidef]#half_precision::divide#
 [source,role=synopsis,id=api:half-divide]
 ----
-float divide(float x, float y)                      (1)
+float divide(float x, float y)                      (1) /* deprecated */
 
-template<typename NonScalar1, typename NonScalar2>  (2)
+template<typename NonScalar1, typename NonScalar2>  (2) /* deprecated */
 /*return-type*/ divide(NonScalar1 x, NonScalar2 y)
 ----
 
@@ -25870,9 +25870,9 @@ The return type is [code]#NonScalar# unless [code]#NonScalar# is the
 .[apidef]#half_precision::exp#
 [source,role=synopsis,id=api:half-exp]
 ----
-float exp(float x)                (1)
+float exp(float x)                (1) /* deprecated */
 
-template<typename NonScalar>      (2)
+template<typename NonScalar>      (2) /* deprecated */
 /*return-type*/ exp(NonScalar x)
 ----
 
@@ -25900,9 +25900,9 @@ The return type is [code]#NonScalar# unless [code]#NonScalar# is the
 .[apidef]#half_precision::exp2#
 [source,role=synopsis,id=api:half-exp2]
 ----
-float exp2(float x)                (1)
+float exp2(float x)                (1) /* deprecated */
 
-template<typename NonScalar>       (2)
+template<typename NonScalar>       (2) /* deprecated */
 /*return-type*/ exp2(NonScalar x)
 ----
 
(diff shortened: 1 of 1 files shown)


illuhad:
> Deprecating these functions in favor of using real half precision (via sycl::half or std::float16_t) will make SYCL simpler.

Not necessarily opposed to this, but one thing we should discuss: `half` is an optional feature in SYCL. The `half_precision` namespace as far as I'm aware is not.
So currently, application developers can use these functions to hint to SYCL *in a portable way* that these operations can be executed in fp16, even if the backend/target does not support the half type (perhaps the backend has some special fp16 math builtins).

Perhaps the `native::` functions could be used as well for that purpose, but I can imagine that developers might feel less comfortable using math functions that might have even lower precision.

illuhad:
> Clearly the developer here intended for cos to be evaluated in half-precision, and they used both the half type and the half_precision namespace to convey that (which seems reasonable). But half_precision::cos only accepts float, and some implementations define an implicit conversion from half to float... So, this code compiles without error, but probably doesn't run in half precision.

Yeah, I can see that this can be a source of problems - but I would say that this is orthogonal to the question of whether there is value in `half_precisision::`. If we decided that we want to keep `half_precision`, we could solve this particular problem by just providing `half` overloads for `half_precision::` functions.


FWIW - I just checked what AdaptiveCpp does with `half_precision`. Looks like we just map it to the normal builtins, except for `recip` which uses a `native::` builtin internally. So it seems that currently our users don't really benefit a lot from `half_precision` anyway.

Pennycook:
> Yeah, I can see that this can be a source of problems - but I would say that this is orthogonal to the question of whether there is value in `half_precisision::`. If we decided that we want to keep `half_precision`, we could solve this particular problem by just providing `half` overloads for `half_precision::` functions.

Are you suggesting that we'd say that `half_precision::cos(half)` guarantees half-precision, but `half_precision::cos(float)` has implementation-defined precision?  That seems even more confusing, to me.

> FWIW - I just checked what AdaptiveCpp does with `half_precision`. Looks like we just map it to the normal builtins, except for `recip` which uses a `native::` builtin internally. So it seems that currently our users don't really benefit a lot from `half_precision` anyway.

Thanks for checking.  I think DPC++ is inconsistent: it looks like our old host device forwarded to the regular `std::` built-ins, and our device path currently generates SPIR-V intrinsics corresponding to the OpenCL `half_` functions.  The fact that implementations disagree on how to handle these makes me more convinced we should just get rid of them!

gmlueck:
FWIW, CUDA also has a variety of math functions with different accuracy guarantees and parameter types.  For example:

* `__half hcos(const __half a)`
* `__nv_bfloat16 hcos(const __nv_bfloat16 a)`
* `float cosf(float x)` -- 2 ULP accuracy
* `double cos(double x)` -- 2 ULP accuracy
* `float __cosf(float x)` -- For x in [-pi, pi], the maximum absolute error is 2^-21.19, and larger otherwise

(https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____HALF__FUNCTIONS.html)
(https://docs.nvidia.com/cuda/cuda-c-programming-guide/#mathematical-functions-appendix)

Note that they have both a function that takes `__half` (and `__nv_bfloat16`) and also a low-precision function that takes `float`.  I could not find any documentation on the accuracy guarantee of the functions that take `__half` or `__nv_bfloat16`.

I grant you, though, that SYCL (and OpenCL) are a bit weird because they have both "native" precision and "half" precision functions.

How much of this is just confusion because we chose bad names?  For example, if the SYCL type was called `fp16` would things be more clear?  Then, we would have a version of `cos` that takes an `fp16` parameter and a half-precision version of `cos` that takes a `float` parameter.

illuhad:
> Are you suggesting that we'd say that half_precision::cos(half) guarantees half-precision, but half_precision::cos(float) has implementation-defined precision? That seems even more confusing, to me.

No, I would have rather thought in the direction of something like: `half_precision::cos(float)` is equivalent to `half_precision::cos(static_cast<half>(x))`.

Pennycook:
> No, I would have rather thought in the direction of something like: `half_precision::cos(float)` is equivalent in terms of precision to `half_precision::cos(static_cast<half>(x))`.

Would that mean making the `half_precision::` functions optional, then?  Giving them an accuracy requirement would force implementers to emulate 16-bit versions of these functions even on devices that don't support `half`.

illuhad:
> Would that mean making the half_precision:: functions optional, then? Giving them an accuracy requirement would force implementers to emulate 16-bit versions of these functions even on devices that don't support half.

We could also say that the float version must be at least as precise as the half version, with a non-normative note that implementations are encouraged to exploit half precision there for performance, if possible.

I also wouldn't necessarily be opposed to making `half` (at least with a limited set of supported builtins) mandatory in principle. Every SYCL implementation already needs to support `half`<->`float` conversions to make the `half` class work on the host, independently from whether a device supports fp16. So, every SYCL implementation should already have mechanisms to emulate `half`. And even if a device does not have native support, emulated half could still be useful e.g. as a storage optimization. But I suppose that's a bigger discussion.

gmlueck (approved):

tomdeakin:

PR #835, opened 2025-06-02
Clarify dimension must be in range [0, Dimensions)

### LINKED ISSUE #551: Expected behaviour of nd_item get_global_id(x) where x>NDims
What is the expected behaviour of `nd_item::get_global_id(x)` where `x` is greater than the number of dimensions templating the `nd_item`.

e.g.
```cpp
constexpr int NDims = 1;
q.submit([&](sycl::handler &cgh) {
  cgh.parallel_for<KernelName>(sycl::nd_range<NDims>{globalSize, localSize},
                               [=](sycl::nd_item<NDims> it) {
                                 size_t dim0 = it.get_global_id(0);
                                 size_t dim1 = it.get_global_id(1);
                                 size_t dim2 = it.get_global_id(2);
                               });
});
```

What should `it.get_global_id(1)` and `it.get_global_id(2)` return in this case? The spec only states in the `nd_item` member description of  `size_t get_global_id(int dimension) const`: 

> Return the constituent element of the [global id](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#global-id) representing the work-item’s position in the [nd-range](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#nd-range) in the given Dimension.

I see a few potential options to remedy the spec with:

1. Return `0` 
2. Throw an asynchronous error
3. Undefined behaviour

Ideally, IMO, returning `0` would be preferable since it is precisely defined and does not affect the execution of the kernel.

Pennycook:
Accessing a dimension outside of the range supported by a class should result in undefined behavior. This clarification impacts all classes that are used to describe the kernel index space (e.g., id, item, group).

Closes #551.

---

Note that I changed some of the wording from "Returns" to "Equivalent to".  This was deliberate, because "Equivalent to" inherits preconditions, whereas "Returns" does not.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -12059,6 +12059,8 @@ size_t get(int dimension) const
 ----
    a@ Return the value of the specified dimension of the
       [code]#range#.
+      Results in undefined behavior if [code]#dimension# is not in the range
+      [code]#[0, Dimensions)#.
 
 a@
 [source]
@@ -12067,6 +12069,8 @@ size_t& operator[](int dimension)
 ----
    a@ Return the l-value of the specified dimension of the
       [code]#range#.
+      Results in undefined behavior if [code]#dimension# is not in the range
+      [code]#[0, Dimensions)#.
 
 a@
 [source]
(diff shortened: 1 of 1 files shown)


etomzak:
LGTM

tomdeakin:

PR #836, opened 2025-06-02
Add `noexcept` to index space classes

### LINKED ISSUE #626: Consider making methods of range and index space types `noexcept`
This report originated from a discussion about a change to Intel's SYCL implementation to handle exceptions which can potentially fire from within `noexcept` functions (as discovered by a static code analyzer) here: https://github.com/intel/llvm/pull/15173#discussion_r1767195124

Section [4.9.1. Ranges and index space identifiers](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#ranges-identifiers) defines several classes: `range`, `nd_range`, `id`, `item`, `h_item`, `nd_item` and `group` with many methods and I think that all those methods should be marked as `noexcept`.

`item`, `nd_item`, `h_item` and `group` (?) are not even user-constructible, meaning that they can only be used within kernels where exception-handling cannot be used. There is some unclarity about what that means discussed in #543, but my understanding is that the spirit of the SYCL specification is that you can't throw exceptions from device code. Therefore, I don't think that users should expect exceptions being thrown from any methods of those functions.

`range`, `nd_range` and `id` are user-constructible and they *are* used on host as well, so they are a bit trickier. I suppose that it is not a good idea to have different `noexcept` specification for host and device compilation (won't it be an ODR violation or something like that?), but perhaps all their methods should anyways also be marked as `noexcept` consistently.

A few more notes about those user-constructible classes:
- They have `get_*` APIs and subscript operators which accept an index and the spec is not clear what should the behavior be if the index is invalid, see #551. An exception is one of a possible solutions, but we can't use them on device, or can we?
- Some of those have `.size()` method which can potentially results in integer overflow (if `range` is too huge, for example). But considering that their components are `size_t` and therefore unsigned, I think that we can ignore that overflow, because it is well-defined for unsigned types.

I think that addition of `noexcept` sets better requirements for implementations and improves consistency of the specification with respect to things which are illegal in device code.

Pennycook:
Classes like `id`, `range`, `group` and `sub_group` are used predominantly inside of device code where exceptions are not supported. Over time we have slowly gotten better at marking functions like these `noexcept`, and the equivalents of these functions are already marked `noexcept` in various places (e.g., KHR extensions).

Closes #626.

### DIFF
--- a/adoc/headers/group.h
+++ b/adoc/headers/group.h
@@ -12,101 +12,101 @@ template <int Dimensions = 1> class group {
 
   /* -- common interface members -- */
 
-  id<Dimensions> get_group_id() const;
+  id<Dimensions> get_group_id() const noexcept;
 
-  size_t get_group_id(int dimension) const;
+  size_t get_group_id(int dimension) const noexcept;
 
-  id<Dimensions> get_local_id() const;
+  id<Dimensions> get_local_id() const noexcept;
 
-  size_t get_local_id(int dimension) const;
+  size_t get_local_id(int dimension) const noexcept;
 
-  range<Dimensions> get_local_range() const;
+  range<Dimensions> get_local_range() const noexcept;
 
-  size_t get_local_range(int dimension) const;
+  size_t get_local_range(int dimension) const noexcept;
 
-  range<Dimensions> get_group_range() const;
+  range<Dimensions> get_group_range() const noexcept;
 
-  size_t get_group_range(int dimension) const;
+  size_t get_group_range(int dimension) const noexcept;
 
-  range<Dimensions> get_max_local_range() const;
+  range<Dimensions> get_max_local_range() const noexcept;
 
-  size_t operator[](int dimension) const;
+  size_t operator[](int dimension) const noexcept;
 
-  size_t get_group_linear_id() const;
+  size_t get_group_linear_id() const noexcept;
 
-  size_t get_local_linear_id() const;
+  size_t get_local_linear_id() const noexcept;
 
-  size_t get_group_linear_range() const;
+  size_t get_group_linear_range() const noexcept;
 
-  size_t get_local_linear_range() const;
+  size_t get_local_linear_range() const noexcept;
 
-  bool leader() const;
+  bool leader() const noexcept;
 
   // Deprecated in SYCL 2020.
   template <typename WorkItemFunctionT>
-  void parallel_for_work_item(const WorkItemFunctionT& func) const;
+  void parallel_for_work_item(const noexcept WorkItemFunctionT& func) const noexcept;
 
   // Deprecated in SYCL 2020.
   template <typename WorkItemFunctionT>
   void parallel_for_work_item(range<Dimensions> logicalRange,
-                              const WorkItemFunctionT& func) const;
+                              const noexcept WorkItemFunctionT& func) const noexcept;
 
   // Deprecated in SYCL 2020. 
   template <typename DataT>
   device_event async_work_group_copy(local_ptr<DataT> dest,
                                      global_ptr<DataT> src,
-                                     size_t numElements) const;
+                                     size_t numElements) const noexcept;
 
   // Deprecated in SYCL 2020.
   template <typename DataT>
   device_event async_work_group_copy(global_ptr<DataT> dest,
                                      local_ptr<DataT> src,
-                                     size_t numElements) const;
+                                     size_t numElements) const noexcept;
 
   // Deprecated in SYCL 2020.
   template <typename DataT>
   device_event async_work_group_copy(local_ptr<DataT> dest,
                                      global_ptr<DataT> src,
                                      size_t numElements,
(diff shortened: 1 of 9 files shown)


TApplencourt:
I suppose it's not an `abi break`? 
Also that mean we forbid CPU implementation to throw, but i guess it's worse it? (but now that I think about it, not to be sure what the advantage is to call them no-except . To avoid `catch / throw` on CPU code?)

Pennycook:
> I suppose it's not an `abi break`?

My understanding is that it's not an ABI break, because `noexcept` doesn't affect the mangled name of a function.

> Also that mean we forbid CPU implementation to throw, but i guess it's worse it? (but now that I think about it, not to be sure what the advantage is to call them no-except . To avoid `catch / throw` on CPU code?)

It's a compiler optimization.  If the compiler knows that a function can't possibly throw an exception, it can optimize more aggressively.  I've only read about this a little bit, but if the compiler thinks a function _may_ throw an exception, it needs to generate special code to unwind the call stack and call the destructor of any objects within the `try` block.

This is probably most important for SYCL implementations targeting devices that support exceptions, but such implementations do exist.

> In principle this change could also be observable by user code if users rely on `noexcept(some expression involving those SYCL classes)`. Not sure if anyone does that, but in theory it could break user code.

I've not used `noexcept` this way before.  Would it really be a break?  If somebody is branching on the result of a `noexcept`, then shouldn't their code already be handling both branches?  Making something `noexcept` when it wasn't `noexcept` before seems okay to me, since if anything codes should get pushed down a more optimized path.

But if people are really worried about this, I'd be okay with retargeting this for SYCL-Next.  SYCL 2020 code wouldn't break, but developers would have to adjust to the `noexcept` changes if they target SYCL-Next instead.  This is probably going to be true anyway, since as I noted above many of the KHRs are much better about declaring everything `noexcept`.

illuhad:
>  Would it really be a break? If somebody is branching on the result of a noexcept, then shouldn't their code already be handling both branches? Making something noexcept when it wasn't noexcept before seems okay to me, since if anything codes should get pushed down a more optimized path.

I think you could construct a case where this causes a problem:

```c++
// Be noexcept if `operator+(id,id)` is noexcept
void some_user_host_function() noexcept(noexcept(id{}+id{})) {
  if(some_error)
    throw some_execption{};
}
```
Currently, since `id` functionality is not noexcept, this function would not currently be `noexcept` and thus throwing the exception would be fine. If we change SYCL classes to `noexcept`, error handling via exceptions would break in this code.

This code is pretty artificial of course, and I don't know if anybody does that or whether there are "reasonable" use cases for something like this.
You could argue, as you say, that the code should be prepared for the noexcept to return `true` if the code checks behavior for noexcept. But perhaps the situation is not as obvious, and the exception is thrown deep in the call graph of `some_user_host_function` by some other function.

I'm not saying that it makes a lot of sense to write code like this, I'm just pointing out that this change *could* cause code that until now has worked fine (even if perhaps in a somewhat sketchy design) to no longer behave correctly.

keryell (approved):
That makes sense.
Please resolve the conflicts.

adoc/headers/group.h
```diff
@@ -12,101 +12,101 @@ template <int Dimensions = 1> class group {
...
-  std::size_t get_local_linear_range() const;
+  std::size_t get_local_linear_range() const noexcept;
 
-  bool leader() const;
+  bool leader() const noexcept;
 
   // Deprecated in SYCL 2020.
   template <typename WorkItemFunctionT>
-  void parallel_for_work_item(const WorkItemFunctionT& func) const;
+  void parallel_for_work_item(const noexcept WorkItemFunctionT& func) const noexcept;
```
gmlueck: Is `noexcept` allowed to decorate a parameter like this?  It seems like this is diagnosed as an error: https://godbolt.org/z/4a9bGqfsM

Even if it were allowed, I assume this would mean that the user needs to define their callback function `noexcept`.  Adding that requirement would be an API break since existing applications may not do this.

adoc/headers/group.h
```diff
@@ -12,101 +12,101 @@ template <int Dimensions = 1> class group {
...
 
   // Deprecated in SYCL 2020.
   template <typename WorkItemFunctionT>
   void parallel_for_work_item(range<Dimensions> logicalRange,
-                              const WorkItemFunctionT& func) const;
+                              const noexcept WorkItemFunctionT& func) const noexcept;
```
gmlueck: Same here.

adoc/headers/group.h
Pennycook: No, this was just a find-and-replace bug.  Good catch.  Fixed in c6f7b78da0e26c5ce674b1566013d7332650119b.

adoc/headers/group.h
Pennycook: Should also be fixed in c6f7b78da0e26c5ce674b1566013d7332650119b.

PR #837, opened 2025-06-02
Explicitly delete handler move/copy constructors

### LINKED ISSUE #807: What prevents `handler` from being moved or copied?
### Specification Version

SYCL 2020 (Revision 9)

### Section Number(s)

- Section 4.9.4. Command group `handler` class, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:handlerClass

### Issue Description

The specification says:
> It is disallowed for an instance of the SYCL `handler` class to be moved or copied.

It is unclear whether "disallowed" means that the implementation must prevent this somehow, or if it's simply undefined behavior. The definition of the `handler` class does not show that the copy and move constructors are deleted, nor that they exist but are private.

Note that the issue and potential fixes are very similar to those being discussed for the `group` class in #627.

### Code Example (Optional)

_No response_

Pennycook:
Previously we said in words that a handler could not be moved or copied, but it was unclear whether this meant that it should be forbidden by the class implementation or simply resulted in undefined behavior.

This change explicitly deletes the handler move and copy constructors, making it clear that it is the implementation's responsibility to prevent a handler from being moved or copied.

Closes #807.

### DIFF
--- a/adoc/headers/commandGroupHandler.h
+++ b/adoc/headers/commandGroupHandler.h
@@ -9,6 +9,12 @@ class handler {
   handler(___unspecified___);
 
  public:
+  // A handler cannot be moved or copied.
+  handler(const handler&) = delete;
+  handler(handler&&) = delete;
+  handler& operator=(const handler&) = delete;
+  handler& operator=(handler&&) = delete;
+
   template <typename DataT, int Dimensions, access_mode AccessMode,
             target AccessTarget, access::placeholder IsPlaceholder>
   void require(


adoc/headers/commandGroupHandler.h
```diff
@@ -9,6 +9,12 @@ class handler {
   handler(___unspecified___);
 
  public:
+  // A handler cannot be moved or copied.
+  handler(const handler&) = delete;
+  handler(handler&&) = delete;
+  handler& operator=(const handler&) = delete;
+  handler& operator=(handler&&) = delete;
+
```
etomzak, line 15: I think this PR should align with however the non-user-constructible discussions is resolved (e.g., https://github.com/KhronosGroup/SYCL-Docs/pull/627).

Is there a reason that the implementation-defined constructor is `private` and these functions are `= delete`d? Ideally we'd decide on one way to enforce non-user-constructibility and apply that to all ways of creating an object.

gmlueck: I think it would make sense to specify the default `handler` constructor as deleted, which would align with #627.

I think we should just remove the `private` constructor from the synopsis.  We generally don't show private implementation details in the spec.  Also, since the parameters to the private constructor are `__unspecified__`, it's not clear to me what value there is in listing it in the spec.

Pennycook: I agree, and I'm happy to make this change.

But I want to point out that if we remove the private constructor now, it will renumber all the tables in the specification because it will remove Table 108.  If that's not a concern, I can make the change here.  If we don't want the table numbers to change until SYCL-Next, I guess we should split it out into a separate PR.

gmlueck: I think it is ok to renumber the tables.  We have already renumbered the tables in our previous updates because the "new format" replaces tables that describe the member functions with paragraphs (that are not in tables).  As a result, there are fewer tables, so all the subsequent tables got renumbered.

Pennycook: Removed in ec62b7f4524881fa7241487f88af669542608e32 🥳 .

adoc/headers/commandGroupHandler.h
```diff
@@ -4,11 +4,13 @@
 namespace sycl {
 
 class handler {
- private:
-  // implementation defined constructor
-  handler(___unspecified___);
-
  public:
+  // A handler cannot be moved or copied.
```
gmlueck, line 10: Should we also show the default constructor as deleted?

Pennycook: Oops.  Yes, sorry, I meant to make that change too.  Fixed in 488e03838e968b6d9d3c7f9ae5a52cf7cfb8d22f.

adoc/headers/commandGroupHandler.h
etomzak: As a matter of protocol, I don't think we should be closing conversations before the person initiating the conversation has had a chance to review.

I don't think #627 *has* converged on a solution, so I don't think it's correct to say that ec62b7f aligns with #627.

Pennycook: Apologies, I should have waited.

My recollection was that the last time we discussed #627 in a SYCL WG meeting, there was consensus that we would delete the constructor.  There were still some open questions about whether we should introduce additional wording about `group` being an aggregate, but I don't think that applies to `handler`.

nliber (approved):

adoc/headers/commandGroupHandler.h
etomzak: I think we still need a definition for *non-user-constructible*, but that's definitely out of scope for this PR.

Some arguments in favor of deleting these functions that came up in the call today are:

* Many ways exist for implementations to construct `handler` objects, so deleting these functions isn't a substantial hindrance (my main concern)
* Consistency with `std`
* Clearer compiler errors for users

LGTM

keryell (approved):
Thanks.

PR #839, opened 2025-06-05
[Editorial] Add "khr" to paragraph titles

gmlueck:
These functions are all in the `khr` namespace, so "khr::" should be listed in the paragraph titles.


gmlueck:
Merging as editorial.

PR #844, opened 2025-06-19
Change event API description to new style

etomzak:


adoc/chapters/programming_interface.adoc
```diff
@@ -4704,295 +4704,358 @@ execution.
...
+_Effects:_ Constructs an [code]#event# that is immediately ready.
+The [code]#event# has no dependencies and no associated commands.
+Waiting on this [code]#event# will return immediately and querying its status
+will return [api]#info::event_command_status::complete#.
 
-|====
+_Remarks:_ The event is constructed as though it were created from a
+default-constructed [code]#queue#.
+Therefore, its backend is the same as the backend of the device selected by
+[code]#default_selector_v#.
```
etomzak, line 4752: I've added the reference to `default_selector_v`, because SYCL doesn't actually have a default device.

adoc/chapters/programming_interface.adoc
```diff
@@ -4704,295 +4704,374 @@ execution.
...
-[code]#info::event_profiling#.
-The possible values for each info parameter and any restrictions are defined in
-the specification of the <<backend>> associated with the <<event>>.
-All info parameters in [code]#info::event_profiling# are specified in
-<<table.event.profilinginfo>> and the synopsis for [code]#info::event_profiling#
-is described in <<appendix.event.descriptors>>.
+'''
+
+[[sec:event-profiling-info-descriptors]]
+==== Profiling information descriptors
```
etomzak, line 4946: I've used the same documentation format for `get_profiling_info()` as `get_info()`. I.e., the general behavior of the function template is described in the member functions section, and the specific behavior of each specialization is described in the descriptors section. This makes sense for `get_info()`, because different specializations could do very different things WRT `_Effects_`, `_Throws_`, etc. It makes less sense for `get_profiling_info()`, because all the specializations have (almost) the same behavior. Would it be better to somehow treat `namespace sycl::info::event_profiling` in a section like "other enumerations" instead?

For example, all specializations of `get_profiling_info()` return `uint64_t` and throw the same exception under the same conditions. ATM this information appears twice, because I've just lifted the text over from the previous format. But it seems redundant.

However, `get_profiling_info` has different blocking behavior depending on the specialization, so each specialization does need its own text somewhere.

TL;DR: Among the specializations of `get_profiling_info()`, there is significant shared behavior and significant diverging behavior. It's un-ergonomic to keep jumping around the spec to try to understand what each specialization does. How to best structure this?

adoc/chapters/programming_interface.adoc
```diff
@@ -4704,295 +4704,374 @@ execution.
...
-|====
-@ Status @ Description
-a@
-[source]
+_Throws:_ An [code]#exception# with the [code]#errc::invalid# error code if the
+SYCL queue that submitted the <<command-group>> that this event is associated
+with was not constructed with the [api]#property::queue::enable_profiling#
+property.
+
+Each profiling information descriptor specifies the return value.
```
```suggestion
Each profiling information descriptor specifies the return value and may also specify preconditions, additional exceptions that are thrown, etc.  See <link> for the profiling information descriptors that are defined by the core SYCL specification.
```

We should use text like this, similar to what we do for `get_info` and `get_backend_info`.  In particular, it's important to have a link to the profiling descriptors, like we do for the information descriptors.

adoc/chapters/programming_interface.adoc
```diff
@@ -4704,295 +4704,374 @@ execution.
...
+};
+} // namespace sycl::info::event_profiling
 ----
 
-    @ [.code]#uint64_t#
-   a@ Returns a timestamp telling when the associated <<command-group>> was
-      submitted to the [code]#queue#.  This is always some time after the
-      <<command-group-function-object>> returns and before the associated call
-      to [code]#queue::submit# returns.
+_Returns:_ The timestamp for when the associated <<command-group>> was submitted
```
```suggestion
_Returns:_ The timestamp when the associated <<command-group>> was submitted
```

The double preposition "for when" is awkward.  I think we just want "when" here.  Same for the other profiling descriptors below.

adoc/chapters/programming_interface.adoc
```diff
@@ -4704,295 +4704,374 @@ execution.
...
+to the [code]#queue#.
 
-a@
-[source]
+_Throws:_ An [code]#exception# with the [code]#errc::invalid# error code if the
+SYCL queue that submitted the <<command-group>> that this event is associated
+with was not constructed with the [api]#property::queue::enable_profiling#
+property.
```
gmlueck: Do not duplicate the _Throws_ paragraph for each of these profiling descriptors.  We already document this in the description of `event::get_profiling_info`.

This is similar to the way we will handle the backend info descriptors.  The description of `get_backend_info` specifies that `errc::backend_mismatch` will be thrown under certain circumstances, so there is no need to specify this also for each backend info descriptor.

adoc/chapters/programming_interface.adoc
gmlueck: > Would it be better to somehow treat `namespace sycl::info::event_profiling` in a section like "other enumerations" instead?

No.  I think what you have is the better direction.  I've added some comments below that will reduce the duplicate text.

adoc/chapters/programming_interface.adoc
etomzak: I don't think this suggestion is correct. A timestamp isn't a *when*; it's a *what* that represents a *when*, if that makes sense.

I think what we need here is something like, "The timestamp that corresponds to the time when the associated command-group ...," but more concise. Some options might be:

* "The timestamp corresponding to when the associated ..."
* "The time when the associated ..."

adoc/chapters/programming_interface.adoc
etomzak: Changed in 4c6bfc0

adoc/chapters/programming_interface.adoc
etomzak: Changed in 4c6bfc0

adoc/chapters/programming_interface.adoc
gmlueck: I think we could use the current wording:

> The timestamp telling when the associated ...

If you don't like that, I think your first suggestion is OK too:

> The timestamp corresponding to when the associated ...

etomzak: Done in e9bb73d

keryell (approved):
Thanks.

tomdeakin:
WG approved to merge as clarification to 2020.

Issue #845, opened 2025-06-23
get_native implies device exceptions


etomzak:
### Specification Version

SYCL 2020 (Source)

### Section Number(s)

[§ 4.5.1.2. Template function get_native](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_template_function_get_native)

### Bug Description

`get_native()` may be used in device code, because it "supports kernel function interoperability." `get_native()` is required to throw when there's a mismatch between its object parameter's backend and its template parameter. Throwing isn't supported in device code, so this part of the specification isn't implementable.

Pragmatically, I expect that this is a low-priority issue. Few users are likely to run into it. Short of defining device-side error handling, I'm not sure how to fix it. The best option might, unfortunately, be to make a device-side `get_native()` backend mismatch into UB.

Pennycook:
> Agree, I don't see the real purpose of `get_native` in device code neither. like you want the `visa / assembly` of some math operation? Seem unlikely to me!

I think the intended use-case is to enable developers to call functions written in other languages from SYCL.

#### OpenCL C Function
```c
void my_opencl_c_function(__global int* ptr, image1d_t im) {
  /* Do something in OpenCL */
}
```

#### SYCL Kernel
```c++
struct Kernel {

  void operator()(sycl::nd_item<1> it) const {
    my_opencl_c_function(get_native<backend::opencl>(my_accessor), get_native<backend::opencl>(my_image));
  }

  accessor<int, 1, access::mode::read_write, target::device> my_accessor;
  unsampled_image_accessor<float, 1, access::mode::read, image_target::device> my_image;

};
```

I can see the value in this, but I agree that the existing `get_native()` interface is problematic.

I agree with @etomzak that UB is probably the best we can do here, and I think it makes sense.  It seems possible that `get_native<backend::X>` might return a type that isn't even guaranteed to be defined (e.g., if the device compiler knows it's targeting a different backend) and so things might break at compile-time.

If we remove the `make_` functions (see https://github.com/KhronosGroup/SYCL-Docs/issues/777), maybe it makes sense to deprecate the `get_native` functions too?  Similar to my argument in that issue: it seems like better design to let each backend define its own interoperability interface, rather than try and force everything to adhere to a single interface.

etomzak:
I've taken a crack at fixing the wording in #847

PR #847, opened 2025-06-25
Add UB to get_native()

etomzak:
Clarify that `get_native()` only needs to throw in application scope.

Addresses #845

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -260,9 +260,11 @@ for kernel function interoperability.
 The availability and behavior of these template functions are defined by the
 <<backend>> specification document.
 
-The [code]#get_native# function must throw an [code]#exception# with the
-[code]#errc::backend_mismatch# error code if the backend of the SYCL object
-doesn't match the target backend.
+In <<application-scope>>, the [code]#get_native# function must throw a
+synchronous [code]#exception# with the [code]#errc::backend_mismatch# error code
+if the backend of the SYCL object doesn't match the target backend.
+In <<command-group-scope>> and <<kernel-scope>>, the behavior is undefined if
+the backend of the SYCL object doesn't match the target backend.
 
 [[sec:backend-interoperability-make]]
 ==== Template functions [code]#make_*#


adoc/chapters/programming_interface.adoc
```diff
@@ -260,9 +260,11 @@ for kernel function interoperability.
...
 <<backend>> specification document.
 
-The [code]#get_native# function must throw an [code]#exception# with the
-[code]#errc::backend_mismatch# error code if the backend of the SYCL object
-doesn't match the target backend.
+In <<application-scope>>, the [code]#get_native# function must throw a
+synchronous [code]#exception# with the [code]#errc::backend_mismatch# error code
+if the backend of the SYCL object doesn't match the target backend.
+In <<command-group-scope>> and <<kernel-scope>>, the behavior is undefined if
+the backend of the SYCL object doesn't match the target backend.
```
Pennycook: I think this should be:

```suggestion
In <<application-scope>> or <<command-group-scope>>, the [code]#get_native# function must throw a
synchronous [code]#exception# with the [code]#errc::backend_mismatch# error code
if the backend of the SYCL object doesn't match the target backend.
In <<kernel-scope>>, the behavior is undefined if
the backend of the SYCL object doesn't match the target backend.
```

or

```suggestion
In host code, the [code]#get_native# function must throw a
synchronous [code]#exception# with the [code]#errc::backend_mismatch# error code
if the backend of the SYCL object doesn't match the target backend.
In device code, the behavior is undefined if
the backend of the SYCL object doesn't match the target backend.
```

All code in command-group scope runs on the host, so we should still be able to throw the exception there.

etomzak: Fixed in a615f7d

I got turned around when trying to think about this in terms of "application interoperability" vs "kernel function interoperability."

I'm now wondering: would the spec benefit from a non-normative note pointing out that `get_native<some_backend>(my_accessor)` (et al) can throw in command group scope but can be UB in kernel scope? It's potentially surprising that the same code misbehaves in different ways depending on which `{}` it's in.

Pennycook: I agree that a non-normative note would be helpful here.

etomzak: I've added a note in 936b181.

I've phrased it in quite general terms. Reading the spec again, each specialization of `get_native()` is permitted either on host or on device, never both, so it's a little bit less foot-gun-y than I initially thought.

adoc/chapters/programming_interface.adoc
```diff
@@ -260,10 +260,18 @@ for kernel function interoperability.
...
+[code]#exception# with the [code]#errc::backend_mismatch# error code if the
+backend of the SYCL object doesn't match the target backend.
+In device code, the behavior is undefined if the backend of the SYCL object
 doesn't match the target backend.
 
+{note}When calling [code]#get_native#, care must be taken to ensure that the
+backend of the SYCL object matches the target backend.
+While a mismatched call to [code]#get_native# can be relied on to throw in host
+code, it will invoke undefined behavior in device code.
+{endnote}
```
gmlueck: Do we need this non-normative note?  It seems like it just duplicates the information from the previous paragraph, adding only that "care must be taken".  I don't think we need a non-normative note just to tell people to take care to avoid UB.

etomzak: See the discussion here: https://github.com/KhronosGroup/SYCL-Docs/pull/847#discussion_r2166963517

I don't have a strong opinion either way.

gmlueck: Thanks for pointing to the discussion.  I still feel that the non-normative note doesn't add anything beyond the preceding paragraph, so I still think we should remove it.

TApplencourt: Agree with greg. And nice clarification :)

etomzak: I've removed the note

gmlueck (approved):

TApplencourt (approved):

gmlueck:

PR #849, opened 2025-06-25
Clarify work-item scope atomics (and memory model in general)

### LINKED ISSUE #665: What does it mean for an atomic_ref to have work_item scope?
What is the expected behaviour of an `atomic_ref` with `work_item` scope? Does it make sense to apply a per-thread atomic?

The memory ordering enabled by the `atomic_ref` may still provide some benefit to individual threads, however, couldn't the same be achieved with `atomic_fence`s instead of using an `atomic_ref`?

Are there any cases where `work_item` scoped atomics are meaningful, other than the aforementioned case?

How should an implementation handle `work_item` scoped atomics?

Pennycook:
This started as an attempt just to clarify work-item scope atomics, but I think the fix has ended up addressing some broader long-standing issues with the memory model.

I think figuring out the steps that a SYCL application needs to take to guarantee "sequential consistency" is the only remaining open in the memory model, and I'm afraid to touch it. 😆 

Closes #665.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1096,14 +1096,23 @@ programming and to provide a meaningful way to describe the behavior of
 <<group,groups>> containing a single work-item.
 {endnote}
 
-Potentially concurrent conflicting actions with different memory scopes are
-considered a data race, resulting in undefined behavior.
+Potentially concurrent conflicting actions with different memory scopes may lead
+to a data race, resulting in undefined behavior.
+An atomic operation _A_ with scope _S~1~_ operating on the same memory location
+as atomic operation _B_ with scope _S~2~_ is a data race if:
+
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~1~_; or
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~2~_.
 
 An atomic operation _A_ with scope _S~1~_ can only synchronize with another
 atomic operation _B_ with scope _S~2~_ if:
 
-* The work-item which executed _A_ is in both _S~1~_ and _S~2~_; and
-* The work-item which executed _B_ is in both _S~1~_ and _S~2~_.
+* The work-items which executed _A_ and _B_ are both in the same group of
+  work-items associated with scope _S~1~_; and
+* The work-items which executed _A_ and _B_ are both in the same group of
+  work-items associated with scope _S~2~_.
 
 The memory scopes are listed above from narrowest
 ([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1099,6 +1099,12 @@ programming and to provide a meaningful way to describe the behavior of
 Potentially concurrent conflicting actions with different memory scopes are
 considered a data race, resulting in undefined behavior.
 
+An atomic operation _A_ with scope _S~1~_ can only synchronize with another
+atomic operation _B_ with scope _S~2~_ if:
+
+* The work-item which executed _A_ is in both _S~1~_ and _S~2~_; and
+* The work-item which executed _B_ is in both _S~1~_ and _S~2~_.
+
 The memory scopes are listed above from narrowest
 ([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
 
@@ -1119,10 +1125,9 @@ supplied.
 ====
 The addition of memory scopes to the {cpp} memory model modifies the definition
 of some concepts from the {cpp} core language.
-For example: the synchronizes-with relationship and sequential consistency must
-be defined in a way that accounts for atomic operations with differing (but
-compatible) scopes, in a manner similar to the <<opencl20, OpenCL 2.0
-specification>>.
+For example: sequential consistency must be defined in a way that accounts for
+atomic operations with differing (but compatible) scopes, in a manner similar to
+the <<opencl20, OpenCL 2.0 specification>>.
 Efforts to formalize the memory model of SYCL are ongoing, and a formal memory
 model will be included in a future version of the SYCL specification.
 ====
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1096,6 +1096,9 @@ programming and to provide a meaningful way to describe the behavior of
 <<group,groups>> containing a single work-item.
 {endnote}
 
+Potentially concurrent conflicting actions with different memory scopes are
+considered a data race, resulting in undefined behavior.
+
 The memory scopes are listed above from narrowest
 ([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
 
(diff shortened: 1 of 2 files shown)


Pennycook:
> We use the term "set" , should we replace `group` for `set` here?
> ...
> I tried to rewrite this little section, but really not sure if it's better...

I think I'll need to read (and re-read!) things a few times before I can determine which wording I prefer.

But to shed some light on why I chose the wording I did... One thing that is pretty subtle here is that it's not enough to just compare the `memory_scope` values themselves. As a concrete example of what I mean, this is a data race:

```c++
float* ptr = 0x42;

// Work-item 0, in Work-group 0
atomic_ref<float, memory_order::seq_cst, memory_scope::work_group>(*ptr) += 1;

// Work-item 0, in Work-group 1
atomic_ref<float, memory_order::seq_cst, memory_scope::work_group>(*ptr) += 1;
```

Even though these atomics use the same `memory_scope` value, the work-items themselves are in different work-groups.  This is what I was trying to convey when talking about the "group of work-items associated with the scope".

The OpenCL version of this wording defines the concept of "inclusive scope" to try and explain this (see [here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_overview_of_atomic_and_fence_operations)) but I  personally think their wording is quite unclear.

keryell (approved):
Thanks!

adoc/chapters/programming_interface.adoc
```diff
@@ -20340,6 +20340,10 @@ scope for the atomic operations.
 Most member functions also provide an optional parameter that allows the
 application to override this default.
 
+All accesses to an object referenced by an [code]#sycl::atomic_ref# must
+exclusively occur through instances of an [code]#sycl::atomic_ref# with the same
+[code]#DefaultScope#.
```
gmlueck: What is the reason for adding this since you also defined the behavior when two atomic operations have different scopes?  In any case, even if the `DefaultScope` is the same, you can always override the default by providing an explicit value for `scope` to the member function which performs the atomic operation.

Pennycook: Good question.

The intent of `sycl::atomic_ref` has always been to align with `std::atomic_ref`, which allows for non-lock-free implementations of atomics on arbitrary trivially-copyable types. We limited `sycl::atomic_ref` in SYCL 2020 to fundamental types that could always be lock-free, but wanted to leave the door open to extensions and future specification updates that removed this restriction.

The idea here was to allow optimizations in the lock case: for example, if an implementation knows that the `atomic_ref` will be used only to synchronize at work-group scope, it could use a lock implemented in work-group-local memory.

...but having written this out, I realize that such an optimization would only be valid if explicitly-overridden scope values had to be narrower than `DefaultScope`.  I don't think we say that anywhere, so maybe it's better to take this out.  We could always reintroduce something like this later, and tie it to the behavior of non-lock-free types.

TApplencourt (approved):

tomdeakin:
@gmlueck Please can you add to your cherry-pick list. Thanks.

adoc/chapters/architecture.adoc
```diff
@@ -1089,6 +1089,31 @@ values:
...
+
+Potentially concurrent conflicting actions with different memory scopes may lead
+to a data race, resulting in undefined behavior.
+An atomic operation _A_ with scope _S~1~_ operating on the same memory location
+as atomic operation _B_ with scope _S~2~_ is a data race if:
+
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~1~_; or
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~2~_.
```
gmlueck, line 1107: John and I were having a side discussion about this part of the PR before he left.  The question is whether operations _A_ and _B_ need to have the **same** scope, or whether it is sufficient for the scopes to include both work-items.  To illustrate, consider the following example:

```
// Work-item A
sycl::atomic_ref<int, memory_order::release, memory_scope::work_group> a(mem);
a.store(1);

// Work-item B (in the same work-group a A)
sycl::atomic_ref<int, memory_order::acquire, memory_scope::device> a(mem);
int x = a.load();
```

Note that the two operations have different scopes, but each scope includes both _A_ and _B_.

The question is whether SYCL should guarantee that these operations are atomic even though the scopes are different.  My first question to John was about Intel hardware.  We think that Intel hardware *is* guaranteed to be atomic in this scenario, so we have no concerns from the standpoint of our own ability to implement the proposed SYCL wording.

However, then we realized that the OpenCL specification seems to **not** guarantee atomicity in this case.  Instead, the OpenCL wording seems to require both operations to have the **same** scope in order to guarantee atomicity.  There is some debate, though, about whether the OpenCL wording should be changed.  There is an open internal issue against the OpenCL specification on this point:

https://gitlab.khronos.org/opencl/OpenCL-Docs/-/issues/367

The SYCL WG should consider whether we want to adopt the wording that John proposes in this PR even though it guarantees atomicity in a case that OpenCL does not guarantee.  Or, whether we should adopt the same language about atomicity that is currently in the OpenCL spec,

gmlueck:
I realize that the WG approved this already, but I was wondering if we could reconsider.  There are two point I would like to raise:

* The review comment [here](https://github.com/KhronosGroup/SYCL-Docs/pull/849#discussion_r2246145903) did not get resolved.  In fact, John's last response ends with "... so maybe it's better to take this out".  I think this means that he intended to remove that paragraph before merging this PR.

* John and I were having an internal side discussion about this PR, which never really got resolved before he left.  That conversation was initially about Intel hardware (which is why it was internal), but then it branched out into OpenCL semantics.  In retrospect, we should have made the conversation public at that point.  I tried to capture our discussion in [this](https://github.com/KhronosGroup/SYCL-Docs/pull/849/files#r2320255099) comment.

adoc/chapters/programming_interface.adoc
tomdeakin: Let's remove this.

gmlueck: Done in 04a8e80cf06e21ab654b2f6d9d594aa1351a50ab

adoc/chapters/architecture.adoc
TApplencourt: PR OpenCL up (https://gitlab.khronos.org/opencl/OpenCL-Docs/-/merge_requests/226)

TApplencourt:
>Will adopt OpenCL into SYCL when OpenCL change have been merged.

Thomas: Look like OpenCL have been merged. So time to review and resurrect this PR.

adoc/chapters/architecture.adoc
```diff
@@ -1089,6 +1089,31 @@ values:
     the memory allocation containing the referenced object, as defined by the
     capabilities of <<buffer,buffers>> and <<usm>>.
 
+{note}An atomic operation with work-item scope is effectively the same as a
```
VerenaBeckham, line 1092: work-item scope is now removed, so we can remove this whole paragraph.

adoc/chapters/architecture.adoc
```diff
@@ -1089,6 +1089,31 @@ values:
...
+non-atomic operation.
+[code]#sycl::memory_scope::work_item# is primarily intended to simplify generic
+programming and to provide a meaningful way to describe the behavior of
+<<group,groups>> containing a single work-item.
+{endnote}
+
+Potentially concurrent conflicting actions with different memory scopes may lead
```
VerenaBeckham, line 1099: Define "conflicting"?

adoc/chapters/architecture.adoc
VerenaBeckham: This looks like it's now supported in OpenCL 3.1, but not in < 3.1. I guess the WG needs to decide whether we want to mandate 3.1 for a SYCL backend.

TApplencourt:
Verana will take the lead on this one. Thanks!

PR #850, opened 2025-06-26
Clarify accumulator type for reductions and scans

### LINKED ISSUE #342: Reductions and scans: types of intermediate results
In the spec [4.17.4.5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_reduce) and [4.17.4.6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_exclusive_scan_and_inclusive_scan) some of the functions allow 2 or even 3 different types for the values scanned/reduced, initial value and values finally stored. The widest example is
``` cxx
template <typename Group, typename InPtr, typename OutPtr, typename T,
          typename BinaryOperation>
OutPtr joint_exclusive_scan(Group g, InPtr first, InPtr last, OutPtr result,
                            T init, BinaryOperation binary_op);
```
with 3 different types allowed. It is said in the spec that `binary_op` should return specific type for the specified input types, but it is not stated that intermediate result should be stored as this return type. From the tests of the functions on hipSYCL I've got an impression that at least some functions use for intermediates type of `std::remove_pointer_t<OutPtr>` and not `T` returned by `binary_op(init, *first)`.
If storage is intended to be in return type of operation indeed, I suggest to clarify the spec.

Pennycook:
Previously we said that the binary operator must return a value of a specific type, but were not explicit that this type should also be used to accumulate partial results.

Closes #342.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -22371,6 +22371,10 @@ _Returns:_ The result of combining the values resulting from dereferencing all
 pointers in the range [code]#[first, last)# using the operator
 [code]#binary_op#, where the values are combined according to the generalized
 sum defined in standard {cpp}.
+
+_Remarks:_ Intermediate results are stored as objects of type
+[code]#std::iterator_traits<Ptr>::value_type#.
+
 --
 
   . _Constraints:_ Available only if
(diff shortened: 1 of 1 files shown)


gmlueck:

PR #851, opened 2025-06-26
Add `host_task` to handler synopsis

### LINKED ISSUE #843: Why is `host_task` excluded from the `handler` synopsis?
### Specification Version

SYCL 2020 (Revision 9)

### Section Number(s)

- Section 4.9.4, `handler` class, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:handlerClass
- Section 4.10, Host tasks, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:interfaces.hosttasks
- Section 4.10.3, Additions to the handler class, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:interfaces.hosttasks

### Issue Description

While working on #644, I noticed that we didn't define an equivalent function for `host_task`.  The original DPC++ extension that inspired #644 also didn't include a `host_task` equivalent.  I was trying to work out how this might have happened, and I noticed something weird about the specification.

The `handler` synopsis in Section 4.9.4 does not include `host_task`.  (I suspect this is how we missed it.)

Section 4.10 includes a small snippet that defines `host_task` inside of an incomplete `handler` class, and there is a dedicated section (Section 4.10.3) describing these "additions to the `handler` class.

Why is `host_task` described this way?  Excluding it from the `handler` synopsis makes it seem as though it might not be there in all implementations.  Is there some restriction like that which I'm missing?

Additionally: It's weird that `host_task` is introduced alongside `interop_handle` and a bunch of backend-specific stuff.  There are use-cases for host tasks that aren't related to interop, and I don't think many people would expect to find `host_task` buried in this part of the specification.

### Code Example (Optional)

_No response_

Pennycook:
`host_task` is unconditionally available via `handler` and should be listed in the handler synopsis.

Rewriting the specification such that `host_task` is described alongside other handler functions would be a large change that would renumber sections, and should be postponed until SYCL-Next.

Closes #843.

### DIFF
--- a/adoc/headers/commandGroupHandler.h
+++ b/adoc/headers/commandGroupHandler.h
@@ -130,5 +130,9 @@ class handler {
   template <auto& SpecName>
   typename std::remove_reference_t<decltype(SpecName)>::value_type
   get_specialization_constant();
+
+  template <typename T>
+  void host_task(T&& hostTaskCallable);
+
 };
 } // namespace sycl


gmlueck:
Merging as editorial

PR #852, opened 2025-06-26
Clarify that the CL/sycl.hpp header is deprecated

### LINKED ISSUE #137: Can SYCL implementation define symbols within ::cl::sycl namespace in <sycl/sycl.hpp> header?
The question originally raised by @psalz in https://github.com/KhronosGroup/SYCL-CTS/pull/108.

[Section 4.3](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:headers-and-namespaces) of the spec says following:

> SYCL provides one standard header file: <sycl/sycl.hpp>, which needs to be included in every translation unit that uses the SYCL programming API.
> 
> All SYCL classes, constants, types and functions defined by this specification should exist within the `::sycl` namespace.
> 
> For compatibility with SYCL 1.2.1, SYCL provides another standard header file: <CL/sycl.hpp>, which can be included in place of <sycl/sycl.hpp>. In that case, all SYCL classes, constants, types and functions defined by this specification should exist within the `::cl::sycl` C++ namespace.
> 
> For consistency, the programming API will only refer to the <sycl/sycl.hpp> header and the ::sycl namespace, but this should be considered synonymous with the SYCL 1.2.1 header and namespace.

@psalz noticed inconsistency between implementation in defining `::cl::sycl` namespace symbols in `<sycl/sycl.hpp>` header.
DPC++ defines `::cl::sycl` namespace, whereas hipSYCL doesn't.

Is SYCL implementation allowed to defined `::cl::sycl` namespace from `<sycl/sycl.hpp>` or it can be done only from `<CL/sycl.hpp>`?

Pennycook:
This header was retained only for backwards compatibility.

Other backwards-compatibility-only interfaces in the specification are deprecated, and this one was missed.

Closes #137.

---

@gmlueck - I've marked this editorial because it obviously won't require CTS changes. I defer to you regarding whether it requires discussion in a WG meeting, or if the discussion in #137 is sufficient.

FWIW, DPC++ already says this header is deprecated. AdaptiveCpp does not (as far as I can tell).

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -91,6 +91,7 @@ For compatibility with SYCL 1.2.1, SYCL provides another standard header file:
 [code]#<sycl/sycl.hpp>#.
 In that case, all SYCL classes, constants, types and functions defined by this
 specification should exist within the [code]#::cl::sycl# {cpp} namespace.
+The [code]#<CL/sycl.hpp># header is deprecated.
 
 For consistency, the programming API will only refer to the
 [code]#<sycl/sycl.hpp># header and the [code]#::sycl# namespace, but this should


adoc/chapters/programming_interface.adoc
```diff
@@ -91,6 +91,7 @@ For compatibility with SYCL 1.2.1, SYCL provides another standard header file:
 [code]#<sycl/sycl.hpp>#.
 In that case, all SYCL classes, constants, types and functions defined by this
 specification should exist within the [code]#::cl::sycl# {cpp} namespace.
+The [code]#<CL/sycl.hpp># header is deprecated.
```
gmlueck: The header is tightly connected to the `cl::sycl::` namespace.  Should we say that the namespace is deprecated too?

Pennycook: Good point.  Fixed in 167e295e6dcb6b7fd66cb2916d9d3b5982b6020f.

PR #853, opened 2025-06-26
Clarify that fp16 and fp64 aspects affect pointers

### LINKED ISSUE #526: Do `fp16` and `fp64` aspects cover `half *` and `double *` data types?
I've been looking at [`atomic_ref_add_sub_op_all_types_test_pointers.cpp`](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/atomic_ref/atomic_ref_add_sub_op_all_types_test_pointers.cpp) CTS test failure on a device which does not support `fp64` and realized that the spec is likely not clear enough about what exactly is guarded by `fp16` and `fp64` aspects.

`atomic_ref` tests for pointers unconditionally cover `double *` data type: [`atomic_ref_common.h:166`](https://github.com/KhronosGroup/SYCL-CTS/blob/9691bc093d4d5b8965be68bfb0716e20fda6205b/tests/atomic_ref/atomic_ref_common.h#L166). When implementation targets SPIR-V as intermediate representation for SYCL kernels, use of `double *` leads to emission of the following instructions sequence in SPIR-V module:
```
OpCapability Float64
%Double = OpTypeFloat 64
%DoblePtr = OpTypePointer %Double
```
Which makes it legal for a backend to discard such SPIR-V module if `double` is not supported by a target device. However, at SYCL level the app only uses `double *` and never dereferences that pointer, so there are no direct use of that data type.

The question is what was the intent of the spec with `fp16` and `fp64` aspects? Are they expected to also guard pointers to those optional data types?

I guess that from user experience point of view, pointers to optional data types should be allowed as long as they are not dereferenced on a device which doesn't support those types. Should this be clarified somehow in the SYCL 2020 spec?

Note: SPIR-V spec has special capability `Float16Buffer` capability, which allows pointers to `half` in a module, but doesn't allow their dereferencing. Unfortunately, there is no similar `Float64Buffer` capability.

Pennycook:
The SYCL-CTS was previously modified to run tests requiring `double*` (and `atomic_ref<double*>`) only on devices with the `fp64` aspect. This change updates the specification to reflect the WG's decision.

Closes #526.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3388,7 +3388,7 @@ The precise definition of this aspect is left open to the SYCL implementation.
 [role=synopsis,id=api:aspect-fp16]
 --
 Indicates that kernels submitted to the device may use the [code]#sycl::half#
-data type.
+data type and pointers to [code]#sycl::half#.
 --
 
 '''
@@ -3397,7 +3397,7 @@ data type.
 [role=synopsis,id=api:aspect-fp64]
 --
 Indicates that kernels submitted to the device may use the [code]#double# data
-type.
+type and pointers to [code]#double#.
 --
 
 '''


etomzak:
This is nit-picky, so apologies in advance.

As worded, this PR answers the question, "Am I allowed to use `double*` on a device that supports `fp64`?" From my reading, though, #526 is asking, "Am I disallowed from using `double*` on a device that does not support `fp64`?" I expect the answer is *yes*, but I don't think that's clear. I.e., the question in #526 is essentially, "Does `double*` constitute a 'use' of `double` as far as `aspect::fp64` is concerned?"

I *think* what's missing is some further clarification in [§5.7. Optional kernel features](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:optional-kernel-features) about what's disallowed when an optional feature is not supported. ATM, it's quite difficult to work out what constitutes "compile-able but not used," particularly in light of this sentence in the spec:

> In order to guarantee source code portability of SYCL applications that use optional kernel features, all SYCL implementations must be able to compile device code that uses these optional features regardless of whether the implementation supports the features on any of its devices.

\* All of the above applies to `fp16` as well

Pennycook:
> As worded, this PR answers the question, "Am I allowed to use `double*` on a device that supports `fp64`?" From my reading, though, #526 is asking, "Am I disallowed from using `double*` on a device that does not support `fp64`?" I expect the answer is _yes_, but I don't think that's clear. I.e., the question in #526 is essentially, "Does `double*` constitute a 'use' of `double` as far as `aspect::fp64` is concerned?"

I see what you mean, but this was already a problem: the existing entry in this table says that you can use `double` on a device that supports `fp64`, but isn't necessarily clear that you can't use `double` on a device that doesn't support `fp64`.

> I _think_ what's missing is some further clarification in [§5.7. Optional kernel features](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:optional-kernel-features) about what's disallowed when an optional feature is not supported. ATM, it's quite difficult to work out what constitutes "compile-able but not used," particularly in light of this sentence in the spec:

I'm not sure if this is covered somewhere else already.  I'd like @gmlueck to weigh in on how and where to address this, because if I recall correctly he wrote a lot of the original text for aspects.

gmlueck:
These sentences were intended to make it clear that users may not submit a kernel to a device if a kernel uses an optional feature that the device does not support:

> Of course, applications that make use of optional kernel features should ensure that a kernel using such a feature is submitted only to a device that supports the feature. [...]  If an application fails to do this, the implementation must throw a synchronous `exception` with the `errc::kernel_not_supported` error code from the kernel invocation command (e.g. `parallel_for()`).

gmlueck (approved):

adoc/chapters/programming_interface.adoc
```diff
@@ -3388,7 +3388,7 @@ The precise definition of this aspect is left open to the SYCL implementation.
 [role=synopsis,id=api:aspect-fp16]
 --
 Indicates that kernels submitted to the device may use the [code]#sycl::half#
-data type.
+data type and pointers to [code]#sycl::half#.
```
etomzak, line 3391: How about adding something like this?

```suggestion
data type and pointers to [code]#sycl::half#.
Note that including a pointer to [code]#sycl::half# in a kernel is considered to be use, even if the pointer is never dereferenced.
```

The problem is that without a statement like this, it's unclear how the existing text should be interpreted:

> Of course, applications that make use of optional kernel features should ensure that a kernel using such a feature is submitted only to a device that supports the feature.

It should be made clear that the existence of `half*` is considered "use."

Pennycook: I'm not sure how far we want to go here, so let's discuss this with the wider WG.

One interpretation of your note is that this code should throw an exception:
```c++
q.single_task([=]() {
  double* unused_pointer_variable;
  *a = 1; // could be literally anything
});
```

...but if a compiler optimizes this variable away, the "use" might not exist.  You could get in a similar situation with an unused `double` variable, and we don't currently say anything about those.  Whatever we decide here shouldn't require implementers to optimize `double` and `double*` differently to match "optional feature" semantics.

My understanding of the requested clarification is that _any_ use of a pointer to `double` is considered to be use, even if that use doesn't involve a dereference. So, we're not worried about cases where an unused variable exists, but rather cases where somebody writes something like:

```c++
double** array_of_double_pointers = ...;
q.single_task([=]() {
  array_of_double_pointers[0]++; // NB: This is a use of double*
});
```

But "using a pointer is considered to be use" is a bit too tautological for my liking, and I think adding a note to that effect might lead to even more confusion.

gmlueck: I think all of these optional features have a similar issue.  Does a kernel really "use" a feature if the optimizer happens to remove the code?

TApplencourt: Adding one point, the "if I don't use, it's ok", is a nice feature to have. For example when mapping a struct with lot of field (lot of them potentially invalid -- function pointer, half, hash table, what not) and accessing only the correct one.

keryell: And what about references to `double` and... ?

keryell (changes requested):
I do not want to enter this rabbit hole and prefer the current wording.
I think that "using" is OK for me just with the meaning of "referenced outside of a manifestly constant evaluated expression".

TApplencourt (commented):
After re-consideration, we should keep the vague "use" to allow implementation freedom.

PR #860, opened 2025-06-27
Add new requirements for sub-group sizes

### LINKED ISSUE #193: [sub-groups] Clarify requirements around sub_groups
I came across some questions regarding sub_group semantics and requirements lately (from an implementation point of view).

1. As mentioned in #190, in SYCL there's no notion of control flow divergence yet, although most sub_group algorithm implementations will require that all work-items of the sub_group are active,
2. From [3.9.4](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_work_group_data_parallel_kernels): "The size of any sub-group in the dispatch is between 1 and this maximum sub-group size..". Does this imply an implementation may use sub_groups of e.g. size 4 for work-items 0-123, when the wg size is 127, and 3 sub_groups with 1 element at the end (analogous to having a vectorized loop with a scalar epilogue)? Or is it required for a single kernel dispatch to yield consistent sub_group sizes (i.e. either mask away the last elements or fall back to sg size 1)?
3. what happens if `reqd_sub_group_size(4)` and wg size is 127? Could an implementation define the execution with non-uniform wg/sg-sizes as a "kernel feature" and throw an exception with `errc::kernel_not_supported` in this case?

Thanks for any feedback in advance!

Pennycook:
Since the introduction of SYCL 2020 there have been multiple requests from developers and implementers to provide more detail about how sub-groups are expected to work.

This change is a first step towards more portable sub-groups, by ensuring that sub-group behavior is well-defined in cases where the highest-numbered dimension of a work-group is divisible by the sub-group size. Dividing the highest-numbered dimension into sub-groups is the obvious thing to do in this case, and matches the behavior of existing implementations.

The way in which other work-groups are divided into sub-groups remains implementation-defined, but this change introduces a non-normative note to hint at possible implementation choices. Eventually we may want to consider exposing these choices via device queries or similar, and implementations are exploring this via extensions.

Closes #193.

---

Although I don't think this would break any implementations, I've positioned it as a SYCL-Next feature rather than a SYCL 2020 clarification to be on the safe side.  I think it's important that we do this, but I'm willing to wait for SYCL-Next.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1248,10 +1248,9 @@ single sub-group of size 6.{endnote}
 
 The numbering of work-items in a sub-group reflects the linear numbering of the
 work-items in the work-group.
-Specifically, if a work-item has a linear ID latexmath:[i_s] in the sub-group
-and a linear ID latexmath:[i_w] in the work-group, the work-item with linear ID
-latexmath:[i_s+1] in the sub-group has linear ID latexmath:[i_w+1] in the
-work-group.
+Specifically, if a work-item has a linear ID i~s~ in the sub-group and a linear
+ID i~w~ in the work-group, the work-item with linear ID i~s~+1 in the sub-group
+has linear ID i~w~+1 in the work-group.
 
 Similarly to work-groups, the [code]#group_barrier# function can be used to
 block a work-item until all work-items in the same sub-group arrive at the


adoc/chapters/architecture.adoc
```diff
@@ -1230,23 +1230,37 @@ When the global size is zero, the kernel function is not executed, the local
...
+The size of each sub-group is between 1 and this maximum sub-group size, and is
+invariant for the duration of a kernel's execution.
+
+If the number of work-items in the highest-numbered dimension of a work-group is
+evenly divisible by the sub-group size, all sub-groups in the work-group contain
+the same number of work-items.
+Otherwise, the division of a work-group into sub-groups is
+implementation-defined.
+
+{note}For example, a two-dimensional work-group of size {3, 3} with a maximum
```
keryell: Do not reuse the same number, since it is confusing and might hide bugs.
Pick for example different prime numbers.

adoc/chapters/architecture.adoc
```diff
@@ -1230,23 +1230,37 @@ When the global size is zero, the kernel function is not executed, the local
...
+sub-group size of 16 may be split into two sub-groups of size 3, or treated as a
+single sub-group of size 9.{endnote}
+
+The numbering of work-items in a sub-group reflects the linear numbering of the
+work-items in the work-group.
+Specifically, if a work-item has a linear ID latexmath:[i_s] in the sub-group
+and a linear ID latexmath:[i_w] in the work-group, the work-item with linear ID
+latexmath:[i_s+1] in the sub-group has linear ID latexmath:[i_w+1] in the
```
gmlueck:
```suggestion
i~s~+1 in the sub-group has linear ID i~w~+1 in the
```

I have a slight preference to use asciidoc syntax for the subscripts instead of latexmath.  It renders just as well and avoids bringing in another dependency.  Long term, I think it would be nice to avoid the build dependency on latexmath.

(Same for the other 2 uses of latex math earlier in this paragraph.)

adoc/chapters/architecture.adoc
Pennycook: Fixed in d9a852cb68f755854f6b848a93afab75e38fe72a.  It _was_ hiding a bug, so this is great advice.

adoc/chapters/architecture.adoc
Pennycook: I've made this change in d852560d3a775e949ac7bb1f9a4fd3bebbb4d345.

I'm not opposed to avoiding the build dependency on latexmath, but it might still be nice to invest in a convention that renders a bit differently.  I have a weak preference for latexmath today mainly because it switches font, and I find it slightly easier to read "i+1" as one thing:

image

vs

image

...but the specification isn't consistent about using latexmath for numbers or mathematical expressions, and consistency is more important.

gmlueck: The nice thing about using Asciidoc syntax is that it can be combined with any of the Asciidoc formatting also.  So, we could put this in `[code]#` formatting if we want.

FWIW, we're not consistent about how we format expressions like "i+1", or even a phrase like "Returns 1" (describing a function that returns a value).  Sometimes the expression is in code format like "`i+1`" or "Returns `1`", but more often we don't have any special formatting for the expressions in these cases.

gmlueck: Oh, and one more thing about latexmath ... we used to have cases where the latexmath formatting rendered correctly in HTML but rendered wrong in PDF (and rendered wrong in a way that changed the meaning of the specification).  This is one reason I started replacing latexmath with Asciidoc formatting.

gmlueck (approved):

adoc/chapters/architecture.adoc
Pennycook: Ah, I didn't realize there were cases like that.  I'm happy not to use latexmath, then.

TApplencourt (approved):

PR #861, opened 2025-06-27
Clarify `device_type` and `get_devices()` functions

### LINKED ISSUE #771: Some device types are only defined via OpenCL
### Specification Version

SYCL 2020 (Revision 9)

### Section Number(s)

- [Section 4.6.2 Platform class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:platform-class)
- [Section 4.6.3 Context class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interface.context.class)
- [Section 4.6.4. Device class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-class)
- [Section A.3. Device information descriptors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#appendix.device.descriptors)

### Issue Description

The `device_type` class is defined as:
```c++
enum class device_type : /* unspecified */ {
  cpu,         // Maps to OpenCL CL_DEVICE_TYPE_CPU
  gpu,         // Maps to OpenCL CL_DEVICE_TYPE_GPU
  accelerator, // Maps to OpenCL CL_DEVICE_TYPE_ACCELERATOR
  custom,      // Maps to OpenCL CL_DEVICE_TYPE_CUSTOM
  automatic,   // Maps to OpenCL CL_DEVICE_TYPE_DEFAULT
  host,
  all // Maps to OpenCL CL_DEVICE_TYPE_ALL
};
```

The mapping between these SYCL types and OpenCL is relevant, but should be moved to the OpenCL backend specification. However, removing these comments effectively removes the definition of what the device types "automatic" and "all" do.  "automatic" does not appear anywhere else in the specification.

I think the correct fix here is to:

- Move these comments to the OpenCL backend specification;
- Clarify that a device cannot have `device_type::all` _or_ `device_type::automatic`;
- Clarify that `get_devices(device_type::all)` returns all devices; and
- Clarify that `get_devices(device_type::automatic`) returns a list containing a single device, which must be the same device as the one returned by `platform(default_selector_v).get_devices()`.

### Code Example (Optional)

This is valid SYCL 2020 code:
```c++
sycl::platform p{};
auto devices = p.get_devices(sycl::info::device_type::automatic);
```

...but the specification doesn't currently say what `devices` should contain.

Pennycook:
This pull request makes a couple of changes, all related to clarifying `device_type` and the `device::get_devices()` and `platform::get_devices()` functions.  Specifically:

- Move the mapping between `device_type` and `CL_DEVICE_TYPE` into the OpenCL backend specification.
- Deprecate `device_type::host`, because it has no meaning in SYCL 2020.
- Clarify that a device cannot identify itself as an "automatic" device, just like how it cannot be an "all" device.
- Clarify that `get_devices(automatic)` is intended to return some sort of "default" device.

Closes #771. 

---

It wasn't clear what the difference between `platform::get_devices(device_type::automatic)` and `devices::get_devices(device_type::automatic)` should be from the original wording, but I think what I've proposed here makes sense.

In OpenCL, `clGetDeviceIDs(platform, CL_DEVICE_TYPE_DEFAULT, ...)` returns a per-platform default device, and so it seems pretty likely that the original intent was for `platform::get_devices(device_type::automatic)` to have equivalent behavior.

Given that per-platform behavior, it seems logical that `devices::get_devices(device_type::automatic)` should return the list of all per-platform default devices, instead of just a single device.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3608,7 +3608,7 @@ enum class device_type : /* unspecified */ {
   accelerator,
   custom,
   automatic,
-  host,
+  host, // Deprecated
   all
 };
 } // namespace sycl::info
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1130,7 +1130,7 @@ devices>> associated with this platform which have the device type specified by
 [code]#type#.
 
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
-[code]#std::vector# contains all devices of any type.
+[code]#std::vector# contains all root devices in this platform.
 If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
 contains a single device corresponding to an implementation-defined default
 device for this platform.
@@ -1851,7 +1851,7 @@ devices>> from all <<backend, SYCL backends>> available in the system which have
 the device type [code]#type#.
 
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
-[code]#std::vector# contains all devices of any type.
+[code]#std::vector# contains all root devices in the system.
 If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
 contains one device from each platform, corresponding to the device returned by
 [code]#platform::get_devices(info::device_type::automatic)#.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1129,6 +1129,10 @@ _Returns:_ A [code]#std::vector# containing all of the <<root-device,root
 devices>> associated with this platform which have the device type specified by
 [code]#type#.
 
+{note}Since the concept of a "host device" does not exist in SYCL 2020, if
+[code]#type# is [code]#info::device_type::host# this function will always return
+an empty vector.{endnote}
+
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
 [code]#std::vector# contains all root devices in this platform.
 If [code]#type# is [code]#info::device_type::automatic# and the platform is not
@@ -1850,6 +1854,10 @@ get_devices(info::device_type type = info::device_type::all)
 _Returns:_ A [code]#std::vector# containing all the <<root-device, root
 devices>> from all <<backend, SYCL backends>> available in the system which have
 the device type [code]#type#.
+
+{note}Since the concept of a "host device" does not exist in SYCL 2020, if
+[code]#type# is [code]#info::device_type::host# this function will always return
+an empty vector.{endnote}
 
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
 [code]#std::vector# contains all root devices in the system.
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1870,7 +1870,8 @@ struct device_type {
 _Remarks:_ Template parameter to [api]#device::get_info#.
 
 _Returns:_ The device type associated with the device.
-May not return [api]#info::device_type::all#.
+May not return [code]#info::device_type::all# or
+[code]#info::device_type::automatic#.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3630,7 +3630,7 @@ enum class device_type : /* unspecified */ {
   accelerator,
   custom,
   automatic,
-  host, // Deprecated
+  host, // Deprecated by SYCL 2020
   all
 };
 } // namespace sycl::info
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1129,6 +1129,12 @@ _Returns:_ A [code]#std::vector# containing all of the <<root-device,root
 devices>> associated with this platform which have the device type specified by
 [code]#type#.
 
+_Remarks_: If [code]#type# is [code]#info::device_type::all#, the
+[code]#std::vector# contains all devices of any type.
+If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
+contains a single device corresponding to an implementation-defined default
+device for this platform.
+
 '''
 
 [[sec:platform-static-member-funcs]]
@@ -1843,6 +1849,12 @@ get_devices(info::device_type type = info::device_type::all)
 _Returns:_ A [code]#std::vector# containing all the <<root-device, root
 devices>> from all <<backend, SYCL backends>> available in the system which have
 the device type [code]#type#.
+
+_Remarks_: If [code]#type# is [code]#info::device_type::all#, the
+[code]#std::vector# contains all devices of any type.
+If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
+contains one device from each platform, corresponding to the device returned by
+[code]#platform::get_devices(info::device_type::automatic)#.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -1131,9 +1131,10 @@ devices>> associated with this platform which have the device type specified by
 
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
 [code]#std::vector# contains all root devices in this platform.
-If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
-contains a single device corresponding to an implementation-defined default
-device for this platform.
+If [code]#type# is [code]#info::device_type::automatic# and the platform is not
+empty, the [code]#std::vector# contains a single root device corresponding to an
+implementation-defined default device for this platform.
+If the platform is empty, any call to this function returns an empty vector.
 
 '''
 
@@ -1853,8 +1854,8 @@ the device type [code]#type#.
 _Remarks_: If [code]#type# is [code]#info::device_type::all#, the
 [code]#std::vector# contains all root devices in the system.
 If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
-contains one device from each platform, corresponding to the device returned by
-[code]#platform::get_devices(info::device_type::automatic)#.
+contains one root device from each non-empty platform, corresponding to the
+device returned by [code]#platform::get_devices(info::device_type::automatic)#.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3603,13 +3603,13 @@ application will not be able to submit kernels to that device _D_.
 ----
 namespace sycl::info {
 enum class device_type : /* unspecified */ {
-  cpu,         // Maps to OpenCL CL_DEVICE_TYPE_CPU
-  gpu,         // Maps to OpenCL CL_DEVICE_TYPE_GPU
-  accelerator, // Maps to OpenCL CL_DEVICE_TYPE_ACCELERATOR
-  custom,      // Maps to OpenCL CL_DEVICE_TYPE_CUSTOM
-  automatic,   // Maps to OpenCL CL_DEVICE_TYPE_DEFAULT
+  cpu,
+  gpu,
+  accelerator,
+  custom,
+  automatic,
   host,
-  all // Maps to OpenCL CL_DEVICE_TYPE_ALL
+  all
 };
 } // namespace sycl::info
 ----
--- a/adoc/chapters/opencl_backend.adoc
+++ b/adoc/chapters/opencl_backend.adoc
@@ -158,6 +158,22 @@ properties as described in the table below:
 | [code]#info::device::version#        | [code]#CL_DEVICE_VERSION#
 |====
 
+Additionally, several values of the [code]#device_type# enumeration map directly
+to OpenCL:
+
+[[table.opencl.device_type]]
+.Mapping of SYCL [code]#device_type# enumeration to OpenCL
+[width="40%",options="header",cols="50%,50%"]
+|====
+| SYCL | OpenCL
+| [code]#device_type::cpu#  | [code]#CL_DEVICE_TYPE_CPU#
+| [code]#device_type::gpu#  | [code]#CL_DEVICE_TYPE_GPU#
+| [code]#device_type::accelerator#  | [code]#CL_DEVICE_TYPE_ACCELERATOR#
+| [code]#device_type::custom#  | [code]#CL_DEVICE_TYPE_CUSTOM#
+| [code]#device_type::automatic#  | [code]#CL_DEVICE_TYPE_DEFAULT#
+| [code]#device_type::all#  | [code]#CL_DEVICE_TYPE_ALL#
+|====
+
 === OpenCL memory model
 
 The memory model for SYCL devices running on OpenCL platforms follows the memory
--- a/adoc/headers/deviceInfo.h
+++ b/adoc/headers/deviceInfo.h
@@ -94,7 +94,7 @@ enum class device_type : /* unspecified */ {
   accelerator,
   custom,
   automatic,
-  host,
+  host, // Deprecated
   all
 };
 
--- a/adoc/headers/deviceInfo.h
+++ b/adoc/headers/deviceInfo.h
@@ -89,13 +89,13 @@ struct partition_type_affinity_domain;
 } // namespace device
 
 enum class device_type : /* unspecified */ {
-  cpu,         // Maps to OpenCL CL_DEVICE_TYPE_CPU
-  gpu,         // Maps to OpenCL CL_DEVICE_TYPE_GPU
-  accelerator, // Maps to OpenCL CL_DEVICE_TYPE_ACCELERATOR
-  custom,      // Maps to OpenCL CL_DEVICE_TYPE_CUSTOM
-  automatic,   // Maps to OpenCL CL_DEVICE_TYPE_DEFAULT
+  cpu,
+  gpu,
+  accelerator,
+  custom,
+  automatic,
   host,
-  all // Maps to OpenCL CL_DEVICE_TYPE_ALL
+  all
 };
 
 enum class partition_property : /* unspecified */ {


adoc/chapters/programming_interface.adoc
```diff
@@ -3603,13 +3616,13 @@ application will not be able to submit kernels to that device _D_.
...
-  custom,      // Maps to OpenCL CL_DEVICE_TYPE_CUSTOM
-  automatic,   // Maps to OpenCL CL_DEVICE_TYPE_DEFAULT
-  host,
-  all // Maps to OpenCL CL_DEVICE_TYPE_ALL
+  cpu,
+  gpu,
+  accelerator,
+  custom,
+  automatic,
+  host, // Deprecated
```
etomzak: Is this deprecated, because implementations should just be reporting `cpu`? I.e., now that there is no requirement for a host device, there's never a reason to report `host`?

I'm trying to think if there could be a reason for an implementation to make a distinction between a `cpu` device and a `host` device, but I haven't come up with anything.

Pennycook: That was my logic, exactly.

There are 0 usages of `device_type::host` in the specification, and no definition of what it means. SYCL 1.2.1 provided a `host_selector`, but it wasn't defined in terms of `device_type::host` (which may have just been an oversight).

`cpu_selector` is defined as selecting a device with `device_type::cpu`, so any "host device" will be ignored by most code.  The only way to get a "host device" would be if it was returned by `default_selector_v` (unlikely) or if the developer manually selected it.

I suspect that we should actually have removed this device type when we removed the whole concept of a "host device".  I'd be open to doing that if that's what the SYCL WG decides, but I thought it was safer to start with deprecation.

etomzak: > `cpu_selector` is defined as selecting a device with `device_type::cpu`, so any "host device" will be ignored by most code.

That's a strong argument for making this change, and I'm supportive.

But it *could* change the behavior of existing code, so I think it'll need a WG discussion.

adoc/chapters/programming_interface.adoc
```diff
@@ -1129,6 +1129,12 @@ _Returns:_ A [code]#std::vector# containing all of the <<root-device,root
 devices>> associated with this platform which have the device type specified by
 [code]#type#.
 
+_Remarks_: If [code]#type# is [code]#info::device_type::all#, the
+[code]#std::vector# contains all devices of any type.
```
TApplencourt: Nitpicking, but I personally find the `of any type` more confusing than helping. `contains all devices` seem enough for me.

adoc/chapters/programming_interface.adoc
```diff
@@ -1129,6 +1129,12 @@ _Returns:_ A [code]#std::vector# containing all of the <<root-device,root
...
+If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
+contains a single device corresponding to an implementation-defined default
+device for this platform.
```
TApplencourt: I look like you and Erik got already this discussion, but why not said it's equivalent to `default_seletor_v`? 

I'm not here to discuss if that make sense of not for `default_seletor_v` to return a random new device each time, just that the subtile diference between `automatic` (should always return the same device) and `default_seletor_v` seem too subtle for my liking

adoc/chapters/programming_interface.adoc
Pennycook: That's fair.  I've tweaked this in 7e169a12f7ff5091a6d504fc27adafbc9113b2bd -- is that better?

adoc/chapters/programming_interface.adoc
Pennycook: The device returned by `default_selector_v` is a single device.  When you use `default_selector_v`, the implementation is responsible for choosing which platform and backend you use.  There is a `platform` constructor that accepts a `default_selector_v`, but it doesn't give you a default platform -- it gives you the platform associated with the default device.

So the key difference here is that `device_type::automatic` can be used to programmatically identify the default device for a specific platform.  So if a developer enumerated some platforms (e.g., OpenCL, CUDA), they could ask the implementation to give them the default OpenCL device and the default CUDA device, rather than just *the* default device.

The conversation we were having about whether or not `default_selector_v` always returns the same device is orthogonal to that.

adoc/chapters/programming_interface.adoc
TApplencourt: Yes, Thanks a lot!

```
[code]#std::vector# contains all root devices in the system.
```
What is a system? :p Is aurora with is 60k gpu a system? 

More seriously we use the terme system in other place
> In some cases, an implementation might also choose to expose empty sycl::platform objects, for example if a vendor’s runtime is available, but no devices supported by that runtime are available in the system.

So I guess it's good. system == all-platform. Put I will prefer to said all platform if possible. 
As usual try to minimize the amount of concept introduce : )


Also if we use `root devices` here we should use it everywhere. I personally think that it "obviously-implicit" that it's `root devices`. Because I can create an implementation where my device and infinitly partionnable (make no sense to do so, but I don't see why I cannot do that : ) if I want )

adoc/chapters/programming_interface.adoc
TApplencourt: > The conversation we were having about whether or not default_selector_v always returns the same device is orthogonal to that.

Indeed! Sorry about the confusion

> So the key difference here is that device_type::automatic can be used to programmatically identify the default device for a specific platform. So if a developer enumerated some platforms (e.g., OpenCL, CUDA), they could ask the implementation to give them the default OpenCL device and the default CUDA device, rather than just the default device.

Oh I see
> Given that per-platform behavior, it seems logical that devices::get_devices(device_type::automatic) should return the list of all per-platform default devices, instead of just a single device.

So 
```
for auto p: sycl::platform.get_platforms()
   ds = p.get_devices(`default_selector_v`)
   if !ds.empty
      v.push_back(ds[0])
```
==
```
sycl::device::get_devices(device_type::automatic)
```

Is this correct? 

It really non obvious to me that `automatic` mean `default_selector_per_platorm` but if that make sense for OpenCL people why not.

Pennycook: Yes, I think that matches what I've tried to describe here.  The list contains all of the per-platform default devices.

I agree that it's not obvious, and I made a bit of a judgement call here.  These things individually seemed obvious-ish:

- `platform::get_devices(device_type::automatic)` should return something per platform. It would be weird if it returned nothing, which is what you'd get for some platforms if you tied the behavior to `default_selector_v`. (And there's the precedent that this is a per-platform query in OpenCL.)

- `device::get_devices(device_type::automatic)` was _probably_ intended to do something different to `sycl::device()` and `sycl::device(default_selector_v)`, because otherwise we introduced a really complicated API to do something simple.

- It would be nice if the link between `platform::get_devices(device_type::automatic)` and `device::get_devices(device_type::automatic)` had symmetry with the other device types.  `platform::get_devices(device_type::cpu)` returns all the CPUs in a platform, and `device::get_devices(device_type::cpu)` returns all the CPUs in all the platforms, so it seems logical(-ish) that `device::get_devices(device_type::automatic)` should return _all_ the default devices in all the platforms.

I think part of the problem here is that "automatic" is a bad name.  I don't know why it was renamed, given that it's called "default" in OpenCL.

TApplencourt: >platform::get_devices(device_type::automatic) should return something per platform. It would be weird if it returned nothing, which is what you'd get for some platforms if you tied the behavior to default_selector_v. (And there's the precedent that this is a per-platform query in OpenCL.)

Ah, I see the problem. But I think that it may not be the case. `default_selector_v` said:

>Select a SYCL device from any supported [SYCL backend](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#backend) based on an implementation-defined heuristic. Since all implementations must support at least one device, this selector must always return a device.

The " implementation-defined heuristic" may be "If I'm run by a platform object I will return a device from this platform; If I'm run as a free function from devices, I will return return a random device". It seem to be a valid heuristic to me.

> because otherwise we introduced a really complicated API to do something simple.

Yeah... I guess my problem is that I'm still stuck to this idea. I don't see any advantage of `automatic`. Just adding new complexity of the spec, where a `for loop` may do for some use case.  Currently, I see it like a super diabetic syntaxic-sugar :p But I'm probably missing something, as you said if we have it, I suppose it's because of some reason :) 

>I think part of the problem here is that "automatic" is a bad name. I don't know why it was renamed, given that it's called "default" in OpenCL.

I cannot agree more! And having a `default / automatic` device_type AND a `default_selector` seem redundant. We need one or the other IMO,

TApplencourt:
>I'm not opposed to deprecating it, and I agree it's weird. But even if we deprecate it, we have to figure out what it's supposed to do so people can implement it. 

I know, I know! @nliber give me all the trick, we can just said... UB! :D More seriously agree : ( If we choose  to remove it in SYCL next, then maybe whatever is the simplest / minimal change here / whatever implementer are already doing / CTS asume

adoc/chapters/programming_interface.adoc
Pennycook: > The " implementation-defined heuristic" may be "If I'm run by a platform object I will return a device from this platform". It seem to be a valid heuristic to me.

I don't think this is possible today.  Every API I've found that accepts a `DeviceSelector` expects the `DeviceSelector` to select a device, and things like the platform are derived from the device that was selected.

Remember also that `DeviceSelector` is a general concept, and all it does is _score_ a device.  `default_selector_v` is just a callable that returns a score, and you can see this by running something like:

```c++
#include <sycl/sycl.hpp>
#include <iostream>

int main()
{
    auto selector = sycl::default_selector_v;
    for (auto d : sycl::device::get_devices()) {
        std::cout << selector(d) << std::endl;
    }
}
```

On my system using DPC++, this returns 300 for an Intel CPU and 500 for an Intel GPU.  The fact that I get my GPU as the default device is a direct result of these scores.

TApplencourt: Oh indeed, maybe not implementable to have a `context-aware selector` for all general case indeed. 
But I suppose one can implement:

```
sycl::platform::get_devices(auto c) {
    if c == sycl::default_selector_v
        return self.devices().max_by([](auto d):{ return c(d) } )[0] 
}
```

As you said it's just a scoring. We never said we should return the overall `max`. I think we can return the `max` of any subset. At least it was my understanding.
 `sycl::platform::get_devices(sycl::default_selector_v)` take all the device of this platform, and then rank them with the `default_selector_v` and return the max one (honestly I think `get_device` should return the max one, and `get_devices` the sorted list but it's too late to change that) . 

---- 

>On my system using DPC++, this returns 300 for an Intel CPU and 500 for an Intel GPU. The fact that I get my GPU as the default device is a direct result of these scores.

(Note that in theory i can be a random number :p, so each call you will have a different default device)

Pennycook: Just to make sure I understand:  `get_devices(auto c)` there is a proposed alternative for `automatic`, right?  You can't call `get_devices()` with a `DeviceSelector` today, and we'd need to add that.

I agree that in principle we could define an interface with the behavior you show here, but I think we'd need to call out in the definition that the behavior was special-cased for `default_selector_v`.  The interface would have to work for other `DeviceSelector` types, and so we'd need to define the general behavior too, as something like "The vector contains all the devices for which the selector returns a value >= 0"... But then the special-case behavior is potentially _really_ weird, because there's no guarantee that `default_selector_v` returns a negative value for every device it doesn't select.

---

...This is a much bigger can of worms than I expected.

TApplencourt: Sorry, I was really not clear. And got confused between `device_type` and `selector`. I finaly understand the problem. Sorry all the span! 

I think my argument, is that for both `platform::get_devices` and `devices::get_devices` with `automatic` device type, we should return the sorted devices by `default_selector_v` that they would have returned with `info::device_type:all`

```
device::get_devices(info::device_type type = info::device_type::all) const
       if type ==  info::device_type::automatic {
            auto selector = sycl::default_selector_v;
           return self.get_devices(info::device_type::all).sort_by([](auto d):{ return selector(d) } )
       } 
```
That it. It's more or less useless I agrees, but at least it's simple and no special case.

We have a clear mapping between automatic and `device_selector_v` scoring system. 

And IMO it's nice that `::device.get_devices( info::device_type::automatic)[0] == sycl::device();` (assuming non random `device_selector_v` :p )

Pennycook: Although I agree that what you've proposed would result in a clear mapping, I'm not sure if it's the one we want. I think we might need to have a discussion about this at the SYCL WG meeting and get a few more opinions about what we want this function to do.

If `platform::get_devices(automatic)` is supposed to return a single device, then sorting according to the `default_selector_v` scores isn't enough -- we'd need to sort by the scores and then return the device with the maximum score.  If it returns multiple devices, I think that's would be a break with SYCL 1.2.1 (where `automatic` mapped to `CL_DEVICE_TYPE_DEFAULT` in OpenCL).

I think we'd also need to filter out the devices with negative scores, because the specification says that an implementation will never select those.  But that means an implementation could define a `default_selector_v` that returns -1 for everything except one device, and that would mean that `platform::get_devices(automatic)` could return no devices at all (if the default device was in a different platform).

Saying this function returns one device and leaving it at that might be the simplest thing to do for SYCL 2020.  If we want to redefine the behavior in terms of `default_selector_v`, we might need to make changes to `default_selector_v`, and that might have to wait until SYCL-Next.  (Or we could just deprecate it.)

TApplencourt: > If platform::get_devices(automatic) is supposed to return a single device, then sorting according to the default_selector_v scores isn't enough -- we'd need to sort by the scores and then return the device with the maximum score. If it returns multiple devices, I think that's would be a break with SYCL 1.2.1 (where automatic mapped to CL_DEVICE_TYPE_DEFAULT in OpenCL).

You convinced me! We cannot return more than one device, it will break code for not really good reason.
Logicaly, this lead hat `devices::get_devices(automatic)` should indeed return all the `CL_DEVICE_TYPE_DEFAULT` (so one per platform).

----

Also interestingly, we need to change the spec in another point... We cannot have `empty platform` anymore. Because `CL_DEVICE_TYPE_DEFAULT` should always return true for one device of the platform...  Or can this thing throw? Will be weird... 

> One device in the platform must be returned as the CL_DEVICE_TYPE_DEFAULT device when passed as the device_type to clGetDeviceIDs. (https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clGetDeviceIDs.html) 

>An implementation may also expose empty [platforms](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#platform) that do not contain any [devices](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#device).

So we should change that back... =( 

----

Just super ugly that we know have 2 "default" concept  (`default_selector_v`, and `automatic`) but 🤷🏽 . At least know the spec is clearer thanks to your change!  (and will be still in favor of deprecating `automatic` honestly. But discussion for a futur PR :) ...)

Pennycook: Eugh, I forgot about that.  Do you think we could just say that it must return one device _unless the platform is empty_? If the platform is empty, every call to `platform::get_devices()` should return an empty list.

That seems an easier fix than undoing the previous change.

TApplencourt: Agreed! Let's look at it this way: we're not changing the OpenCL mapping of CL_DEVICE_TYPE_DEFAULT; we're just fixing an old commit where we allowed an empty platform. :)

But one more reason to deprecate it! :p

Pennycook: I've acknowledged the existence of empty platforms in 8b9ea8983b3917915fc6429ae14857396b328160.  Good catch.

TApplencourt (approved):

adoc/chapters/programming_interface.adoc
etomzak: What about something like

```
[code]#std::vector# contains all root devices available to the <<sycl-runtime>>.
```

? It works around @TApplencourt's concern that *system* can mean many things. It also avoids the problem where you can have a single box with 1 CPU + 8 GPUs, where (e.g.,) an environment variable is used to expose only a subset of the GPUs to a single SYCL instance.

Pennycook: I don't think we should try to avoid "system".  As @TApplencourt said, we already use "system" elsewhere (e.g., system allocator, system memory scope, system USM).

`platform::get_platforms()` is defined as returning all platforms in the system, and I think it would be weird for `device::get_devices()` to say something different.  I'm not opposed to revisiting/clarifying what "system" means, but I think it is beyond the scope of this PR.

adoc/chapters/programming_interface.adoc
tomdeakin: The host device was removed in 2020. Removing this enum will be a break, but IMO was probably an oversight.

adoc/chapters/programming_interface.adoc
```diff
@@ -3603,13 +3625,13 @@ application will not be able to submit kernels to that device _D_.
...
-  custom,      // Maps to OpenCL CL_DEVICE_TYPE_CUSTOM
-  automatic,   // Maps to OpenCL CL_DEVICE_TYPE_DEFAULT
-  host,
-  all // Maps to OpenCL CL_DEVICE_TYPE_ALL
+  cpu,
+  gpu,
+  accelerator,
+  custom,
+  automatic,
+  host, // Deprecated
```
```suggestion
  host, // Deprecated by SYCL 2020
```

Unless, of course, we decide to remove it.

adoc/chapters/programming_interface.adoc
```diff
@@ -1844,6 +1855,16 @@ _Returns:_ A [code]#std::vector# containing all the <<root-device, root
...
 
+{note}Since the concept of a "host device" does not exist in SYCL 2020, if
+[code]#type# is [code]#info::device_type::host# this function will always return
+an empty vector.{endnote}
+
+_Remarks_: If [code]#type# is [code]#info::device_type::all#, the
+[code]#std::vector# contains all root devices in the system.
+If [code]#type# is [code]#info::device_type::automatic#, the [code]#std::vector#
+contains one root device from each non-empty platform, corresponding to the
+device returned by [code]#platform::get_devices(info::device_type::automatic)#.
```
gmlueck, line 1866: I think we should also deprecate `info::device_type::automatic`.  The intended way to get the default device is to use the default device selector, not `info::device_type::automatic`.  I'm not sure why we would want two different ways to get a default device.  Furthermore, this behavior where `get_devices(info::device::automatic)` returns a *list* of default devices seems non-intuitive.

My suggestion for SYCL 2020 is:

* Deprecate `info::device_type::automatic`,
* Specify that `info::device_type::automatic` returns an implementation-defined set of devices.

Pennycook: > I'm not sure why we would want two different ways to get a default device. 

`platform::get_devices(info::device_type::automatic)` is the only way to get a default device for a specific platform.

There was some discussion last week about how we could replace these functions with something like `platform::get_devices(DeviceSelector selector)`, rather than using `device_type`.  But then, the behavior of `platform::get_devices(default_selector_v)` could potentially be very surprising, because `default_selector_v` is not currently required to return values compatible with this approach.

To be clear: I think it would be possible to define functions like this and make them well-defined, but we'd have to change what `default_selector_v` means to do it.

gmlueck: > `platform::get_devices(info::device_type::automatic)` is the only way to get a default device for a specific platform.

Is anyone asking for this feature, though?  It seems like we are trying to come up with some rationale for keeping `info::device_type::automatic` in the spec just because it happens to be there now.  IMO, if an application wants to find a default device for a particular platform, they can just write their own selector which gives preference to that platform.  There is no need for us to add any new APIs to the spec.

FWIW, DPC++ currently implements `info::device_type::automatic` the same as `info::device_type::all`, and no one has complained.  I doubt anyone is using it, and the spec has never specified what it's behavior is.

Pennycook: > Is anyone asking for this feature, though? It seems like we are trying to come up with some rationale for keeping `info::device_type::automatic` in the spec just because it happens to be there now.

The feature exists in OpenCL, so somebody wanted it at some point.  You're right that nobody has asked for in SYCL, but it seems like a reasonable feature to me.  SYCL claims that it wants to allow people to use all the devices in a system simultaneously, and that's why we expose multiple platforms.

> IMO, if an application wants to find a default device for a particular platform, they can just write their own selector which gives preference to that platform. There is no need for us to add any new APIs to the spec.

What are you imagining that code would look like?

Let's say I have a platform _A_ and a platform _B_, each containing two devices: _A<sub>1</sub>_, _A<sub>2</sub>_, _B<sub>1</sub>_, and _B<sub>2</sub>_.  Assume the implementation has a preference for _A<sub>2</sub>_ and  _B<sub>2</sub>_ (it doesn't matter why, because it's an implementation-defined preference).  Only one of those two devices can be "the default device", so let's say that `default_selector_v` selects _A<sub>2</sub>_.

How would you write a device selector that returns _B<sub>2</sub>_? I can see how you would prefer platform _B_, but I can't see any way to query the implementation's preference within _B_.

> FWIW, DPC++ currently implements `info::device_type::automatic` the same as `info::device_type::all`, and no one has complained. I doubt anyone is using it, and the spec has never specified what it's behavior is.

I started looking at this because somebody asked why our implementation had this weird behavior.

SYCL 1.2.1 defined that `device_type::automatic` corresponds to `CL_DEVICE_TYPE_DEFAULT`, which I think is enough to infer the expected behavior of `platform::get_devices(info::device_type::automatic)`.  I would argue that we didn't implement it correctly, but nobody realized because there was no CTS test.

gmlueck: > Let's say I have a platform _A_ and a platform _B_, each containing two devices: _A1_, _A2_, _B1_, and _B2_. Assume the implementation has a preference for _A2_ and _B2_ (it doesn't matter why, because it's an implementation-defined preference).

I don't think implementations will have any preference, though.  If the spec forces DPC++ to choose a per-platform "default" device, we will probably just return the first one as an arbitrary decision.

I always thought the default selector was just a convenience for people learning SYCL.  Any serious application that wants to choose a device in a multi-device system should define their own selection algorithm, either by writing their own selector or using an environment variable like `ONEAPI_DEVICE_SELECTOR`.  I can't imagine that you would want to write a production application that relies on the system to determine the best device.  The choice of "best" is likely to depend on the characteristics of your application.  Or, more commonly, the application will consist of several processes, each of which runs on a different device.

Given this model, it's not clear to me why we would want a per-platform default device.  People learning SYCL just want an easy way to run on some device, and they don't care which one.  Production applications will devise their own mechanism for selecting a device.  Having a per-platform default device seems like a middle ground, and I'm not sure which customers would want this.

Pennycook: > I don't think implementations will have any preference, though. If the spec forces DPC++ to choose a per-platform "default" device, we will probably just return the first one as an arbitrary decision.

That would be fine.  But I can think of cases where an implementation may have a preference.

> I always thought the default selector was just a convenience for people learning SYCL. Any serious application that wants to choose a device in a multi-device system should define their own selection algorithm, either by writing their own selector or using an environment variable like `ONEAPI_DEVICE_SELECTOR`. I can't imagine that you would want to write a production application that relies on the system to determine the best device. The choice of "best" is likely to depend on the characteristics of your application. Or, more commonly, the application will consist of several processes, each of which runs on a different device.

I would agree with you if the user had to opt-in to using `default_selector_v`.  But something as simple as `queue q;` creates a queue associated with the default device, and SYCL training materials typically show this pattern. I don't usually see any usage of device selectors in real applications, because our training material tells people to use `ONEAPI_DEVICE_SELECTOR` (which exists outside of the standard).

If we think that production applications write their own device selectors, I think we need to do a better job of highlighting that.  I think rather than just deprecating `automatic` we should deprecate these functions entirely: instead of letting people filter by types, they should be forced to get a list of all the devices and score them.

> Given this model, it's not clear to me why we would want a per-platform default device. People learning SYCL just want an easy way to run on some device, and they don't care which one. Production applications will devise their own mechanism for selecting a device. Having a per-platform default device seems like a middle ground, and I'm not sure which customers would want this.

I can imagine a hypothetical implementation where the `default_device` is dependent on where the user's thread/process was started, for example... So if you have a multi-socket CPU system with multiple GPUs, the default selector could conceivably return the CPU platform and GPU associated with the "closest" NUMA domain.  You can do this with programmatically with something like `hwloc`, and I think there are applications that use this sort of scheme today to find the closest CUDA device in a large multi-GPU system.

But if I'm alone in thinking this is useful, I'm okay with just deprecating it so we can move on.

gmlueck: > I think rather than just deprecating `automatic` we should deprecate these functions entirely: instead of letting people filter by types, they should be forced to get a list of all the devices and score them.

If we were doing this from scratch, I'd definitely be on board with a simpler `get_devices` that does not take a `info::device_type` parameter.  Given that we have it now, though, I'm not sure I would be in a hurry to deprecate it.  Most of the `device_type` values are well defined in the spec, and I assume there is code out there using them.

> I can imagine a hypothetical implementation where the `default_device` is dependent on where the user's thread/process was started, for example... So if you have a multi-socket CPU system with multiple GPUs, the default selector could conceivably return the CPU platform and GPU associated with the "closest" NUMA domain.

I could imagine a device selector that has this functionality.  For example, the selector could take a Posix `cpu_set_t`, and the selector would return the GPU that is closest to one of the CPUs in that set.  This would allow an application to create a `cpu_set_t`, and then use it to set the CPU affinity of a host thread via `sched_setaffinity` and also to get an appropriate GPU device.

This seems unrelated to `device_type::automatic`, though.  We could deprecate that now, and then still add the device selector I suggest above at some later time.

adoc/chapters/programming_interface.adoc
keryell: `host` device was the device really running on the host where any code like system calls were allowed in kernels too.

tomdeakin:
@gmlueck please can you add this to your cherry-pick list. Thanks.

PR #862, opened 2025-06-27
Clarify effect of using accessors outside commands

### LINKED ISSUE #580: Underspecified behavior of accessor::get_pointer and accessor::get_multi_ptr
The specification says:

> This function may only be called from within a command.

...but it doesn't say what should happen if either function is called from outside of a command.  An implementation is free to return any pointer (including a NULL pointer) which may result in bugs that are difficult to track down or understand.  I've seen code like the below hang or segfault across different devices:

```c++
q.submit([&](sycl::handler &h) {

  // This is an (undiagnosed) error, because get_multi_ptr is called on the host
  ptr = buf.get_access<sycl::access::mode::read_write, sycl::access::target::device>(h)
           .get_multi_ptr<sycl::access::decorated::no>();

  h.parallel_for(..., [=](sycl::nd_item<1> it) {
    foo(ptr); // The pointer passed to the device code is undefined
  });
});
```

Is there a reason that we cannot define `get_multi_ptr()` to throw an error when called on the host?

If this is intended to be undefined behavior, I think the specification should state that explicitly.  But an error would be more useful to developers.

Pennycook:
The previous wording said that certain accessor member functions may only be called from within a command, but did not say what should happen if this condition was violated.

The new wording clarifies that this results in undefined behavior, giving different implementations the freedom to decide how to diagnose this error.

Closes #580.

---

Note that #580 was originally focused on just `accessor::get_pointer()` and `accessor::get_multi_ptr()`. While working on the new wording, I realized that the same problem existed for multiple `accessor` member functions, and applied the same fix everywhere.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -7890,7 +7890,7 @@ even if this is a <<ranged-accessor>> whose range does not start at the
 beginning of the buffer.  The return value is unspecified if the accessor is
 empty.
 
-This function may only be called from within a <<command>>.
+_Preconditions_: Must be called within a <<command>>.
 
 Deprecated in SYCL 2020.  Use [code]#get_multi_ptr# instead.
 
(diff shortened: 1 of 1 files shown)


tomdeakin:

PR #863, opened 2025-06-27
New format for reference and value semantics

etomzak:
This converts the reference and value semantics API descriptions from the table-based format to the new section-based format. I've removed the cross-references to these tables, added cross-references to the reference/value sections where they didn't already exist, and aligned the wording of those cross-references.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
 Hidden friend functions have the benefits of avoiding accidental implicit
 conversions and faster compilation.
 
-These common special member functions and hidden friend functions are described
-in <<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>> respectively.
-
 [source,,linenums]
 ----
 include::{header_dir}/common-reference.h[lines=4..-1]
 ----
 
-[[table.specialmembers.common.reference]]
-.Common special member functions for reference semantics
-[width="100%",options="header",separator="@",cols="1,2"]
-|====
-@ Special member function @ Description
-a@
-[source]
+[[sec:reference-semantics-special-members]]
+==== Special member functions
+
+This section describes the special member functions that must be provided by all
+classes providing the common reference semantics (see also the
+<<rule-of-five>>).
+Each of these classes may provide further member functions, static member
+functions, etc., as documented in their respective sections.
+
+'''
+
+.[apititle]#Copy constructor#
+[source,role=synopsis,id=api:reference-semantics-ctor-copy]
 ----
 T(const T& rhs)
 ----
-   a@ Constructs a [code]#T# instance as a copy of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Constructs a [code]#T# instance as a copy of [code]#rhs# in
+accordance with the requirements set out above.
+
+'''
+
+.[apititle]#Move constructor#
+[source,role=synopsis,id=api:reference-semantics-ctor-move]
 ----
 T(T&& rhs)
 ----
-   a@ Constructs a SYCL [code]#T# instance as a move of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Constructs a [code]#T# instance as a move of [code]#rhs# in
+accordance with the requirements set out above.
+
+'''
+
+.[apititle]#Copy assignment operator#
+[source,role=synopsis,id=api:reference-semantics-assignment-copy]
 ----
 T& operator=(const T& rhs)
 ----
-   a@ Assigns this SYCL [code]#T# instance with a copy of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Assigns a copy of [code]#rhs# to this [code]#T# instance in
+accordance with the requirements set out above.
+
+'''
+
+.[apititle]#Move assignment operator#
+[source,role=synopsis,id=api:reference-semantics-assignment-move]
 ----
 T& operator=(T&& rhs)
 ----
-   a@ Assigns this SYCL [code]#T# instance with a move of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Moves from [code]#rhs# and assigns to this [code]#T# instance in
+accordance with the requirements set out above.
+
+'''
+
+.[apititle]#Destructor#
+[source,role=synopsis,id=api:reference-semantics-destructor]
 ----
 ~T()
 ----
-   a@ Destroys this SYCL [code]#T# instance in accordance with the
-      requirements set out in <<sec:reference-semantics>>. On destruction
-      of the last copy, may perform additional lifetime related operations
-      required for the underlying <<native-backend-object>> specified in
-      the <<backend>> specification document, if this SYCL [code]#T# instance
-      was originally constructed using one of the backend interoperability
-      [code]#make_*# functions specified in
-      <<sec:backend-interoperability-make>>.
-      See the relevant backend specification for details.
 
-|====
+_Effects:_ Destroys this [code]#T# instance in accordance with the requirements
+set out above.
 
-[[table.hiddenfriends.common.reference]]
-.Common hidden friend functions for reference semantics
-[width="100%",options="header",separator="@"]
-|====
-@ Hidden friend function @ Description
-a@
-[source]
+_Remarks:_ On destruction of the last copy, may perform additional
+lifetime-related operations required for the underlying
+<<native-backend-object>> specified in the <<backend>> specification document,
+if this [code]#T# instance was originally constructed using one of the backend
+interoperability [code]#make_*# functions specified in
+<<sec:backend-interoperability-make>>.
+See the relevant backend specification for details.
+
+'''
+
+[[sec:reference-semantics-hidden-friends]]
+==== Hidden friend functions
+
+This section describes the hidden friend functions that must be provided by all
+classes providing the common reference semantics.
+Each of these classes may provide further member functions, static member
+functions, etc., as documented in their respective sections.
+
+'''
+
+.[apititle]#Equality operator#
+[source,role=synopsis,id=api:reference-semantics-equality]
 ----
 bool operator==(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is equal to the RHS SYCL
-      [code]#T# in accordance with the requirements set out above,
-      otherwise returns false.
 
-a@
-[source]
+_Effects:_ Returns true if [code]#lhs# is equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
+
+'''
+
+.[apititle]#Inequality operator#
+[source,role=synopsis,id=api:reference-semantics-inequality]
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is not equal to the RHS
-      SYCL [code]#T# in accordance with the requirements set out above,
-      otherwise returns false.
+_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
 
-|====
-
+'''
 
 [[sec:byval-semantics]]
 === Common by-value semantics
@@ -482,77 +503,104 @@ special member functions and member functions in order to fulfill the copy,
 move, destruction and equality requirements, following the <<rule-of-five>> and
 the <<rule-of-zero>>.
 
-These common special member functions and hidden friend functions are described
-in <<table.specialmembers.common.byval>> and
-<<table.hiddenfriends.common.byval>> respectively.
-
 [source,,linenums]
 ----
 include::{header_dir}/common-byval.h[lines=4..-1]
 ----
 
-[[table.specialmembers.common.byval]]
-.Common special member functions for by-value semantics
-[width="100%",options="header",separator="@",cols="2,1"]
-|====
-@ Special member function _(see <<rule-of-five>> and <<rule-of-zero>>)_ @ Description
-a@
-[source]
-----
-T(const T& rhs);
-----
-   a@ Copy constructor.
+[[sec:value-semantics-special-members]]
+==== Special member functions
 
-a@
-[source]
-----
-T(T&& rhs);
-----
-   a@ Move constructor.
+This section describes the special member functions that may be provided by
+classes providing the common by-value semantics.
+If a class provides any one of these special member functions, all of the
+special member functions must be provided (see also the <<rule-of-five>> and the
+<<rule-of-zero>>).
+Each of these classes may provide further member functions, static member
+functions, etc., as documented in their respective sections.
 
-a@
-[source]
-----
-T& operator=(const T& rhs);
-----
-   a@ Copy assignment operator.
+'''
 
-a@
-[source]
+.[apititle]#Copy constructor#
+[source,role=synopsis,id=api:value-semantics-ctor-copy]
 ----
-T& operator=(T&& rhs);
+T(const T& rhs)
 ----
-   a@ Move assignment operator.
 
-a@
-[source]
-----
-~T();
-----
-   a@ Destructor.
+_Effects:_ Constructs a [code]#T# instance as a copy of [code]#rhs#.
 
-|====
+'''
 
-[[table.hiddenfriends.common.byval]]
-.Common hidden friend functions for by-value semantics
-[width="100%",options="header",separator="@",cols="1,1"]
-|====
-@ Hidden friend function @ Description
-a@
-[source]
+.[apititle]#Move constructor#
+[source,role=synopsis,id=api:value-semantics-ctor-move]
+----
+T(T&& rhs)
+----
+
+_Effects:_ Constructs a [code]#T# instance as a move of [code]#rhs#.
+
+'''
+
+.[apititle]#Copy assignment operator#
+[source,role=synopsis,id=api:value-semantics-assignment-copy]
+----
+T& operator=(const T& rhs)
+----
+
+_Effects:_ Assigns a copy of [code]#rhs# to this [code]#T# instance.
+
+'''
+
+.[apititle]#Move assignment operator#
+[source,role=synopsis,id=api:value-semantics-assignment-move]
+----
+T& operator=(T&& rhs)
+----
+
+_Effects:_ Moves from [code]#rhs# and assigns to this [code]#T# instance.
+
+'''
+
+.[apititle]#Destructor#
+[source,role=synopsis,id=api:value-semantics-destructor]
+----
+~T()
+----
+
+_Effects:_ Destroys this [code]#T# instance.
+
+'''
+
+[[sec:value-semantics-hidden-friends]]
+==== Hidden friend functions
+
+This section describes the hidden friend functions that must be provided by all
+classes providing the common by-value semantics.
+Each of these classes may provide further member functions, static member
+functions, etc., as documented in their respective sections.
+
+'''
+
+.[apititle]#Equality operator#
+[source,role=synopsis,id=api:value-semantics-equality]
 ----
 bool operator==(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is equal to the RHS SYCL [code]#T# in accordance with the requirements set out above, otherwise returns false.
 
-a@
-[source]
+_Effects:_ Returns true if [code]#lhs# is equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
+
+'''
+
+.[apititle]#Inequality operator#
+[source,role=synopsis,id=api:value-semantics-inequality]
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is not equal to the RHS SYCL [code]#T# in accordance with the requirements set out above, otherwise returns false.
+_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
 
-|====
+'''
 
 === Properties
 
@@ -4706,9 +4754,6 @@ The SYCL [code]#event# class provides the common reference semantics (see
 
 The constructors and member functions of the SYCL [code]#event# class are listed
 in <<table.constructors.event>> and <<table.members.event>>, respectively.
-The additional common special member functions and common member functions are
-listed in <<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
 
 // Interface for class: event.h
 [source,,linenums]
@@ -5171,9 +5216,6 @@ The SYCL [code]#buffer# class template provides the common reference semantics
 The constructors and member functions of the SYCL [code]#buffer# class template
 are listed in <<table.constructors.buffer>> and <<table.members.buffer>>,
 respectively.
-The additional common special member functions and common member functions are
-listed in <<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
 
 Each constructor takes as the last parameter an optional SYCL
 [code]#property_list# to provide properties to the SYCL [code]#buffer#.
@@ -6040,9 +6082,6 @@ The constructors and member functions of the SYCL [code]#unsampled_image# and
 <<table.constructors.unsampledImage>>, <<table.members.unsampledImage>>,
 <<table.constructors.sampledImage>> and <<table.members.sampledImage>>,
 respectively.
-The additional common special member functions and common member functions are
-listed in <<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
 
 Where relevant, it is the responsibility of the user to ensure that the format
 of the data matches the format described by [code]#image_format#.
@@ -7537,10 +7576,8 @@ The constructors are listed in <<table.accessors.command.buffer.constructors>>,
 and the member functions are listed in <<table.accessors.common.members>> and
 <<table.accessors.command.buffer.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+The [code]#accessor# class provides the common reference semantics as defined in
+<<sec:reference-semantics>>.
 For valid implicit conversions between accessor types refer to
 <<sec:accessor.command.buffer.conversions>>.
 Additionally, accessors of the same type must be equality comparable both in the
@@ -8112,10 +8149,8 @@ The constructors are listed in
 are listed in <<table.accessors.deprecated.common.members>> and
 <<table.accessors.deprecated.constant.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+As with other [code]#accessor# specializations, the class provides the common
+reference semantics as defined in <<sec:reference-semantics>>.
 Additionally, accessors of the same type must be equality comparable.
 
 [source,,linenums]
@@ -8328,10 +8363,8 @@ and the member functions are listed in
 <<table.accessors.deprecated.common.members>> and
 <<table.accessors.deprecated.host.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+As with other [code]#accessor# specializations, the class provides the common
+reference semantics as defined in <<sec:reference-semantics>>.
 Additionally, accessors of the same type must be equality comparable.
 
 [source,,linenums]
@@ -8466,10 +8499,8 @@ The constructors are listed in
 listed in <<table.accessors.deprecated.common.members>> and
 <<table.accessors.deprecated.local.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+As with other [code]#accessor# specializations, the class provides the common
+reference semantics as defined in <<sec:reference-semantics>>.
 Additionally, accessors of the same type must be equality comparable.
 
 [source,,linenums]
@@ -8844,10 +8875,8 @@ The constructors are listed in <<table.accessors.host.buffer.constructors>>, and
 the member functions are listed in <<table.accessors.common.members>> and
 <<table.accessors.host.buffer.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+The [code]#host_accessor# class provides the common reference semantics as
+defined in <<sec:reference-semantics>>.
 For valid implicit conversions between accessor types refer to
 <<sec:accessor.host.buffer.conversions>>.
 Additionally, accessors of the same type must be equality comparable.
@@ -9145,10 +9174,8 @@ The constructors are listed in <<table.accessors.local.constructors>>, and the
 member functions are listed in <<table.accessors.common.members>> and
 <<table.accessors.local.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+The [code]#host_accessor# class provides the common reference semantics as
+defined in <<sec:reference-semantics>>.
 For valid implicit conversions between accessor types refer to
 <<sec:accessor.local.conversions>>.
 Additionally, accessors of the same type must be equality comparable.
@@ -9698,10 +9725,8 @@ The constructors for the two classes are described in
 Both classes also have member functions with the same name, which are described
 in <<table.accessors.unsampled.image.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+The two unsampled image accessor classes provides the common reference semantics
+as defined in <<sec:reference-semantics>>.
 For valid implicit conversions between unsampled accessor types refer to
 <<sec:accessor.unsampled.image.conversions>>.
 
@@ -9890,10 +9915,8 @@ The constructors for the two classes are described in
 Both classes also have member functions with the same name, which are described
 in <<table.accessors.sampled.image.members>>.
 
-The additional common special member functions and common member functions are
-listed in <<sec:reference-semantics>> in
-<<table.specialmembers.common.reference>> and
-<<table.hiddenfriends.common.reference>>, respectively.
+The two sampled image accessor classes provides the common reference semantics
+as defined in <<sec:reference-semantics>>.
 For valid implicit conversions between sampled accessor types refer to
 <<sec:accessor.sampled.image.conversions>>.
 
@@ -11988,16 +12011,13 @@ iteration domain of either a single work-group in a parallel dispatch, or the
 overall Dimensions of the dispatch.
 It can be constructed from integers.
 
-The SYCL [code]#range# class template provides the common by-value semantics
-(see <<sec:byval-semantics>>).
+The SYCL [code]#range# class template provides the common by-value semantics as
+defined in <<sec:byval-semantics>>.
 
 A synopsis of the SYCL [code]#range# class is provided below.
 The constructors, member functions and non-member functions of the SYCL
 [code]#range# class are listed in <<table.constructors.range>>,
 <<table.members.range>> and <<table.functions.range>> respectively.
-The additional common special member functions and common member functions are
-listed in <<sec:byval-semantics>> in <<table.specialmembers.common.byval>> and
-<<table.hiddenfriends.common.byval>> respectively.
 
 [source,,linenums]
 ----
@@ -12236,16 +12256,12 @@ To define this the [code]#nd_range# comprises two ranges: the whole range over
 which the kernel is to be executed, and the range of each work group.
 
 The SYCL [code]#nd_range# class template provides the common by-value semantics
-(see <<sec:byval-semantics>>).
+as defined in <<sec:byval-semantics>>.
 
 A synopsis of the SYCL [code]#nd_range# class is provided below.
 The constructors and member functions of the SYCL [code]#nd_range# class are
 listed in <<table.constructors.ndrange>> and <<table.members.ndrange>>
 respectively.
-The additional common special member functions and common member functions are
-listed in <<sec:byval-semantics>> in <<table.specialmembers.common.byval>> and
-<<table.hiddenfriends.common.byval>> respectively.
-
 
 [[table.constructors.ndrange]]
 .Constructors of the [code]#nd_range# class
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
-[width="100%",options="header",separator="@",cols="1,2"]
-|====
-@ Special member function @ Description
-a@
-[source]
+[[sec:reference-semantics-special-members]]
+==== Special member functions
+
+This section describes the special member functions that must be provided by all
+classes providing the common reference semantics (see also the
```
Pennycook:
```suggestion
This section describes the special member functions that must be provided by all
classes providing common reference semantics (see also the
```

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
 ----
 T(const T& rhs)
 ----
-   a@ Constructs a [code]#T# instance as a copy of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Constructs a [code]#T# instance as a copy of [code]#rhs# in
+accordance with the requirements set out above.
```
Pennycook: I know you didn't write this "in accordance" sentence, but it's not very clear which requirements it's talking about.

Since you're touching this text now, could we do something like:

```suggestion
_Effects:_ Constructs a [code]#T# instance as a copy of [code]#rhs#.

_Postconditions:_ The copy must behave as-if it were the original instance and
as-if any action performed on it is performed on the original instance.
The copy must represent the same underlying native backend object instance,
where applicable.
```

Then we can get rid of the big bullet list of conditions.

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
 ----
 T(T&& rhs)
 ----
-   a@ Constructs a SYCL [code]#T# instance as a move of the RHS SYCL
-      [code]#T# in accordance with the requirements set out above.
 
-a@
-[source]
+_Effects:_ Constructs a [code]#T# instance as a move of [code]#rhs# in
+accordance with the requirements set out above.
```
Pennycook:
```suggestion
_Effects:_ Constructs a [code]#T# instance via a move from [code]#rhs#.
```

I'm not sure if we need to duplicate the postconditions of the copy constructor here or not.  The way I think about move semantics, a move from `rhs` should naturally end up using the same resources.  But I'd to hear from @nliber.

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
-      required for the underlying <<native-backend-object>> specified in
-      the <<backend>> specification document, if this SYCL [code]#T# instance
-      was originally constructed using one of the backend interoperability
-      [code]#make_*# functions specified in
-      <<sec:backend-interoperability-make>>.
-      See the relevant backend specification for details.
 
-|====
+_Effects:_ Destroys this [code]#T# instance in accordance with the requirements
+set out above.
```
Pennycook: Similar to my suggestion above...

```suggestion
_Effects:_ Destroys this [code]#T# instance.
```

The "requirements set out above" in this sentence seem to refer to a paragraph containing "any behavior specific to `T` that is specified as performed on destruction is only performed if this instance is the last remaining host copy".  That seems to be partially duplicated here in the _Remarks_.

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
-@ Hidden friend function @ Description
-a@
-[source]
+_Remarks:_ On destruction of the last copy, may perform additional
+lifetime-related operations required for the underlying
+<<native-backend-object>> specified in the <<backend>> specification document,
+if this [code]#T# instance was originally constructed using one of the backend
+interoperability [code]#make_*# functions specified in
+<<sec:backend-interoperability-make>>.
+See the relevant backend specification for details.
```
Pennycook:
```suggestion
_Remarks:_ Any behavior of [code]#T# that is specified as performed on
destruction is only performed if this instance is the last remaining host
copy.

Destruction of the last copy may perform additional lifetime-related
operations required for the underlying <<native-backend-object>> specified in
the <<backend>> specification document, if this [code]#T# instance was
originally constructed using one of the backend interoperability
[code]#make_*# functions specified in <<sec:backend-interoperability-make>>.
See the relevant backend specification for details.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
 bool operator==(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is equal to the RHS SYCL
-      [code]#T# in accordance with the requirements set out above,
-      otherwise returns false.
 
-a@
-[source]
+_Effects:_ Returns true if [code]#lhs# is equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
```
Pennycook: Again, I think we could pull the equality definition down to here.

adoc/chapters/programming_interface.adoc
```diff
@@ -355,94 +355,115 @@ are hidden from qualified and unqualified lookup.
...
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is not equal to the RHS
-      SYCL [code]#T# in accordance with the requirements set out above,
-      otherwise returns false.
-
-|====
+_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
```
Pennycook:
```suggestion
_Effects:_ Equivalent to [code]#return not (lhs == rhs);#.
```

This would prevent us from having to duplicate the wording about equality, if we make that change.

adoc/chapters/programming_interface.adoc
```diff
@@ -482,77 +503,104 @@ special member functions and member functions in order to fulfill the copy,
...
+[source,role=synopsis,id=api:value-semantics-ctor-move]
 ----
-T(T&& rhs);
+T(T&& rhs)
 ----
-   a@ Move constructor.
 
-a@
-[source]
+_Effects:_ Constructs a [code]#T# instance as a move of [code]#rhs#.
```
Pennycook:
```suggestion
_Effects:_ Constructs a [code]#T# instance via a move from [code]#rhs#.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -482,77 +503,104 @@ special member functions and member functions in order to fulfill the copy,
...
+[source,role=synopsis,id=api:value-semantics-equality]
 ----
 bool operator==(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is equal to the RHS SYCL [code]#T# in accordance with the requirements set out above, otherwise returns false.
 
-a@
-[source]
+_Effects:_ Returns true if [code]#lhs# is equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
```
Pennycook: Same comment as for reference semantics: I think we should declare the requirements here.

adoc/chapters/programming_interface.adoc
```diff
@@ -482,77 +503,104 @@ special member functions and member functions in order to fulfill the copy,
...
+'''
+
+.[apititle]#Inequality operator#
+[source,role=synopsis,id=api:value-semantics-inequality]
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-   a@ Returns true if this LHS SYCL [code]#T# is not equal to the RHS SYCL [code]#T# in accordance with the requirements set out above, otherwise returns false.
+_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
+with the requirements set out above, otherwise returns false.
```
Pennycook:
```suggestion
_Effects:_ Equivalent to [code]#return not (lhs == rhs);#.
```

adoc/chapters/programming_interface.adoc
etomzak: > I know you didn't write this "in accordance" sentence, but it's not very clear which requirements it's talking about.

I had the same thought when doing this refactoring.

I think the big bullet list is valuable, though. Common reference semantics is an essential part of SYCL runtime behavior, and if we do away with the bullet list and tuck away the description of the behavior in the fine print of individual functions, it'll be difficult for a reader to reconstruct what common reference semantics are (and even to appreciate that it's a thing they need to be cognizant of).

My preference would be to put just the bullet list into its own subsection (e.g., a "Description" or "Requirements"). The post-conditions would then be:

```
_Postconditions:_ The copy must behave as-if it were the original instance and
as-if any action performed on it is performed on the original instance.
See Section x.y.z for details.
```

(It might also make sense to talk about common reference/by-value semantics in the architecture chapter, but that's probably out of scope for this PR.)

Pennycook: I think this probably comes down to personal preference, and I appreciate that my view here is extreme.  I would personally prefer that each of the common-by-reference classes declared these functions explicitly in their synopsis, instead of us pointing to an earlier section.

I don't believe that people read specifications from start to finish, and I suspect people tend to look at one piece of the specification at a time.  If somebody is trying to work out how accessor works, for example, I would expect them to Ctrl+F for "accessor" and look at its synopsis -- expecting them to read the whole description, figure out they need to read another section, and read through the bullet pointed list just to learn "it acts like a view" is pretty painful.

Something like you have proposed here could be a good compromise.  Then we could declare constructors with those _Postconditions_ for each relevant class, so that there's an obvious reference back from each constructor to the section that defines the behavior.

But this would be a big change, and I might be alone in wanting it. 😆

adoc/chapters/programming_interface.adoc
etomzak: Currently, every occurrence of `common reference semantics` in the spec takes a `the` in front, so if we change it, we need to change it everywhere.

... unless there's some grammatical subtlety here that I'm not quite grasping, whereby we deliberately want to refer to reference semantics as a general concept and not the SYCL incarnation of that concept.

Pennycook: I don't think "the" implies "SYCL", here, and I think the other "the"s are wrong too.  I haven't done a pass over the specification looking specifically for grammatical issues, but if I see them while reviewing a change I point them out.

For a sentence like:
> The context class provides the common reference semantics as defined in [Section 4.5.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics).

I would argue that the following is more correct:

> The context class provides the common reference semantics defined in Section 4.5.2.

...because then the "the" is referring to the semantics defined in a specific section.  I find the sentence confusing with both a "the" and an "as".

If we instead want to talk about the SYCL incarnation of the concept specifically, with an optional reference to a section, we should say:

> The context class provides SYCL common reference semantics, as defined in [Section 4.5.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics).

This sentence would still make sense even if we removed the "as defined in Section 4.5.2" clause.

---

As for this specific case... Because this "the" is in the section that actually defines common reference semantics, I don't think the "the" makes sense grammatically.  Either removing it or replacing it with "SYCL" would be an improvement, in my opinion.

adoc/chapters/programming_interface.adoc
etomzak: I've taken a stab at a bigger rephrasing of the reference semantics section in d5ed573 to remove some of the clunky phrasing. I'm happy to iterate on this. If we converge on something, I can apply the corresponding changes to the by-value section.

> I would personally prefer that each of the common-by-reference classes declared these functions explicitly in their synopsis, instead of us pointing to an earlier section.

I think there are three options: 1.) Synopsis of the special functions in the "Common interface" section only, 2.) synopsis of the special functions in each class's section only, and 3.) synopsis of the special functions in both the "Common interface" section and each class's section. For me, this falls into the "don't care" category; I can only see weak arguments for each. However, I don't think we should alter the status quo for the "Common interface" section in SYCL 2020, and this PR is intended for SYCL 2020 first. (If we decide to add copy constructors, etc. to reference semantics classes' synopses in SYCL 2020, I'd be OK with that, but that's out of scope for this PR.)

> expecting them to read the whole description, figure out they need to read another section, and read through the bullet pointed list just to learn "it acts like a view" is pretty painful.

I completely agree with this. But I think "tagging" certain classes with `reference semantics` is the best way to address this, which is exactly what the SYCL spec already does. A reader familiar with SYCL sees, "The `platform` class provides the common reference semantics as defined in [link]," and knows exactly what that means. A reader unfamiliar with SYCL can click through and find out. That's why the list of bullets is important. I've given them headings to make it more digestible.

If we were to remove the list of bullets, we'd end up with the exact problem you're trying to avoid. A reader would need to read through each individual class description: "Oh, this one behaves like a `std::shared_ptr`. This other one also behaves like a `std::shared_ptr`. All these SYCL `std::shared_ptr`-like things behave the same way!" We shouldn't leave a reader to discover this common behavior; the common behavior is there intentionally, and we should tell them.

For me, the open question is whether there is value in the API description in the "Common reference semantics" section. E.g., do we actually need to provide signatures for `T::operator==()`? I'm in favor of keeping them, primarily because of the SYCL 2020 argument.

adoc/chapters/programming_interface.adoc
etomzak: Yeah, there are definitely a lot of missing commas in the spec. d5ed573 has improved that for the reference semantics section.

The phrasing here is difficult for two reasons:

1. *Reference semantics* means lots of different things in the field of programming languages. `sycl::id&` has reference semantics, but not SYCL common reference semantics. ATM, the only way the spec has to indicate that it's referring to a specific, definite instance of reference semantics rather than reference semantics in general is with the definite article, which is grammatically correct, but unhelpfully subtle. Barring something better, I do think the *the* is always necessary.
2. The *semantics* in *reference semantics* is acting as an uncountable (indefinite) plural, while the *semantics* in *SYCL common reference semantics* is acting as a singular.

I think *SYCL common reference semantics* is the best way to be intelligible, unambiguous, and grammatically correct. I've added a glossary definition and links in the reference semantics section in 2d36459. I'm happy to iterate on this.

(Whether it's better to say *SYCL common reference semantics* or *the SYCL common reference semantics* is a question I'm hoping to avoid.)

adoc/chapters/glossary.adoc
```diff
@@ -520,6 +520,13 @@ For the full description please refer to <<subsec:buffers>>.
 [[backend-api]]SYCL backend API::
     The exposed API for writing SYCL code against a given <<backend>>.
 
+[[reference-semantics]]SYCL common reference semantics::
+    Several <<sycl-runtime>> classes have reference semantics that are
+    comparable to the semantics of `std::shared_ptr`.
+    These semantics are collectively known as the _SYCL common reference
+    semantics_.
+    See <<sec:reference-semantics>> for details.
```
Pennycook: This feels a little circular to me. Somebody might read this first occurrence of "reference semantics" as "common reference semantics" (because of the ambiguity issue that you pointed out).  I think it would also be nice to hint at _how_ things are similar to `shared_ptr`, rather than make a reader jump from the glossary back to the section right away.

So, how about:
```suggestion
    Several <<sycl-runtime>> classes have semantics 
    comparable to those of `std::shared_ptr`, where multiple instances of
    a class share ownership of a <<native-backend-object>>.
    These semantics are collectively known as _SYCL common reference
    semantics_.
    See <<sec:reference-semantics>> for details.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
-Hidden friend functions have the benefits of avoiding accidental implicit
-conversions and faster compilation.
+The following <<sycl-runtime>> classes provide <<reference-semantics>>:
+[code]#accessor#, [code]#buffer#, [code]#context#, [code]#device#,
+[code]#device_image#, [code]#event#, [code]#host_accessor#,
+[code]#host_sampled_image_accessor#, [code]#host_unsampled_image_accessor#,
+[code]#kernel#, [code]#kernel_id#, [code]#kernel_bundle#,
+[code]#local_accessor#, [code]#platform#, [code]#queue#, [code]#sampled_image#,
+[code]#sampled_image_accessor#, [code]#stream#, [code]#unsampled_image#, and
+[code]#unsampled_image_accessor#.
```
Pennycook: Could we make this into a bulleted list, like we have for the types in [Section 4.17.7](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:integer-functions)?

This would have been bad before because we'd have had two bulleted lists right up against one another, but your new "Requirements" subsection fixes this.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+
+[[sec:reference-semantics-reqs]]
+==== Requirements
```
Pennycook: Definitely nit-picking, but we might want to pick a different word here to avoid confusion with command-group "requirements" and the soon-to-be `khr::requirements` class.

How about "Interface"?

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+
+Each <<sycl-runtime>> class that provides <<reference-semantics>> must conform
+to the following requirements, where [code]#T# is the runtime class type:
+
+* _Copyability:_ [code]#T# must be copy constructible and copy assignable in the
+  host application and within SYCL kernel functions in the case that [code]#T#
+  is a valid kernel argument.
```
Pennycook: This is a little more verbose, but I think it's clearer:
```suggestion
* _Copyability:_ [code]#T# must be copy constructible and copy assignable in the
  host application.
  If [code]#T# is a valid kernel argument, [code]#T# must also be copy
  constructible and copy assignable within SYCL kernel functions.
```

I'll only make this comment once, but there are a few other places where you could make the same change if you agree.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+  Any instance of [code]#T# that is constructed as a copy of another instance,
+  via either the copy constructor or copy assignment operator, must behave as-if
```
Pennycook: I'm not sure if this adds anything.  If we just say "Any copy" we shouldn't need to list out the possibilities.  As an added bonus, if we add new ways to make copies we won't have to remember to rewrite this sentence.
```suggestion
  must behave as-if
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+  it were the original instance and as-if any action performed on it were also
+  performed on the original instance and must represent the same underlying
+  <<native-backend-object>> as the original instance where applicable.
```
Pennycook: Just a slight tweak for readability here:
```suggestion
  it were the original instance and must represent the same underlying
  <<native-backend-object>>, where applicable.
  Any action performed on such a copy must behave as-if it were also
  performed on the original instance.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+* _Destructibility:_ [code]#T# must be destructible in the host application and
+  within SYCL kernel functions in the case that [code]#T# is a valid kernel
+  argument.
+  When any instance of [code]#T# is destroyed, including as a result of the copy
+  assignment operator, any behavior specific to [code]#T# that is specified as
```
Pennycook: Same logic as above.  This is covered by "any".
```suggestion
  When any instance of [code]#T# is destroyed, any behavior specific to [code]#T# that is specified as
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+  performed on destruction is only performed if this instance is the last
+  remaining host copy, in accordance with the above definition of a copy.
+* _Movability:_ [code]#T# must be move constructible and move assignable in the
+  host application and within SYCL kernel functions in the case that [code]#T#
+  is a valid kernel argument.
+  Any instance of [code]#T# that is constructed as a move of another instance,
+  via either the move constructor or move assignment operator, must replace the
```
Pennycook:
```suggestion
  Any instance of [code]#T# that is constructed as a move of another instance
  must replace the
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+  original instance rendering said instance invalid and must represent the same
+  underlying <<native-backend-object>> as the original instance where
+  applicable.
```
Pennycook: A similar readability tweak:
```suggestion
  original instance and must represent the same underlying
  <<native-backend-object>>, where applicable.
  The original instance is rendered invalid.
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+* _Equality:_ [code]#T# must be equality comparable in the host application.
+  Equality between two instances of [code]#T# (i.e. [code]#a == b#) must be true
+  if one instance is a copy of the other and non-equality between two instances
+  of [code]#T# (i.e. [code]#a != b#) must be true if neither instance is a copy
+  of the other, in accordance with the above definition of a copy, unless either
+  instance has become invalidated by a move operation.
+  By extension of the requirements above, equality on [code]#T# must guarantee
+  to be reflexive (i.e. [code]#a == a#), symmetric (i.e. [code]#a == b# implies
+  [code]#b == a# and [code]#a != b# implies [code]#b != a#) and transitive (i.e.
+  [code]#a == b && b == c# implies [code]#c == a#).
```
Pennycook:
```suggestion
* _Equality:_ [code]#T# must be equality comparable in the host application.
  Equality between two instances of [code]#T# (i.e., [code]#a == b#) must be true
  if one instance is a copy of the other and non-equality between two instances
  of [code]#T# (i.e., [code]#a != b#) must be true if neither instance is a copy
  of the other, in accordance with the above definition of a copy, unless either
  instance has become invalidated by a move operation.
  By extension of the requirements above, equality on [code]#T# must guarantee
  to be reflexive (i.e., [code]#a == a#), symmetric (i.e., [code]#a == b# implies
  [code]#b == a# and [code]#a != b# implies [code]#b != a#) and transitive (i.e.,
  [code]#a == b && b == c# implies [code]#c == a#).
```

Recent changes to the specification prefer "i.e.," over "i.e." so we should be consistent.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+* _[code]#std::hash# specialization:_ A specialization of [code]#std::hash# for
```
Pennycook: I'm not sure about this one, but it seems weird that all the entries in this list are "-ability" words except this one.

```suggestion
* _Hashability:_ A specialization of [code]#std::hash# for
```

adoc/chapters/programming_interface.adoc
```diff
@@ -288,72 +288,71 @@ context doesn't match the target backend.
...
+  [code]#T# must exist in the host application that returns a unique value such
+  that if two instances of [code]#T# are equal, in accordance with the above
+  definition, then their resulting hash values are also equal and subsequently
+  if two hash values are not equal, then their corresponding instances are also
+  not equal, in accordance with the above definition.
```
Pennycook: I think in cases like this code can be much clearer:
```suggestion
  for two instances [code]#a# and [code]#b#, 
  [code]#((a == b) && std::hash(a) == std::hash(b)) || ((a != b) && std::hash(a) != std::hash(b))#.
```

Using code saves us from having to say "in accordance with the above definition", because we already defined how `==` and `!=` work for these types.

adoc/chapters/programming_interface.adoc
```diff
@@ -460,8 +472,11 @@ with the requirements set out above, otherwise returns false.
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
-with the requirements set out above, otherwise returns false.
+
+_Effects:_ Equivalent to [code]#!(lhs == rhs)#.
+
+_Remarks:_ Conforms to the equality comparability requirements
```
Pennycook:
```suggestion
_Remarks:_ Conforms to the equality comparability requirements of
```

adoc/chapters/programming_interface.adoc
```diff
@@ -460,8 +472,11 @@ with the requirements set out above, otherwise returns false.
 ----
 bool operator!=(const T& lhs, const T& rhs)
 ----
-_Effects:_ Returns true if [code]#lhs# is not equal to [code]#rhs# in accordance
-with the requirements set out above, otherwise returns false.
+
+_Effects:_ Equivalent to [code]#!(lhs == rhs)#.
```
Pennycook:
```suggestion
_Effects:_ Equivalent to [code]#return !(lhs == rhs)#.
```

I know it looks weird, but ISO C++ convention is to include the `return`.

Pennycook (commented):
I like the new restructuring.  Just a few small wording suggestions.

adoc/chapters/programming_interface.adoc
etomzak: Fixed in 8692759

adoc/chapters/programming_interface.adoc
etomzak: Fixed in 8692759

adoc/chapters/programming_interface.adoc
etomzak: I've sort-of done what you suggested, but I broke up the inline code a bit more.

I'm wondering, though, if we should be referring to the [`Hash` named requirement](https://en.cppreference.com/w/cpp/named_req/Hash.html) instead. It would make things easier for us. More importantly, the `Hash` named requirement allows for a low rate of collisions, whereas our wording implies a collision-free hash function.

Maybe @nliber has thoughts?

adoc/chapters/programming_interface.adoc
etomzak: I've changed it to `Hashability`, but I'm also unsure. I couldn't think of anything better. If we had C++20, we could express all of these as Concepts, I think.

adoc/chapters/programming_interface.adoc
etomzak: Fixed in 8692759

adoc/chapters/programming_interface.adoc
etomzak: I've rephrased this to try to also capture that objects that have been move-assigned to haven't been constructed by that assignment.

adoc/chapters/programming_interface.adoc
etomzak: Done in 8692759

adoc/chapters/programming_interface.adoc
etomzak: Done in 8692759

adoc/chapters/programming_interface.adoc
etomzak: Done in 8692759

adoc/chapters/programming_interface.adoc
etomzak: Agreed. The equivalent change for move was a little bit trickier, but I think it still works.

Done in 8692759

adoc/chapters/programming_interface.adoc
etomzak: Agreed. It was possible to misinterpret the original text.

Done in 8692759

adoc/chapters/programming_interface.adoc
etomzak: I've gone with *Interface requirements*. We tend to use *interface* when we have an API synopsis, which isn't quite what this section is. Other words that came to mind, like *concepts* and *traits*, also clash with other uses of those words.

adoc/chapters/programming_interface.adoc
etomzak: Done in 8692759

I've gone ahead and used the `[api]` tag to turn these into links. I expect some of the anchor names will be changed as those sections get the new format, but it's a starting point.

adoc/chapters/glossary.adoc
etomzak: I've tried to reword this so as not to imply a 1:1 mapping between SYCL common reference instances and backend objects.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,161 +288,213 @@ context doesn't match the target backend.
...
+  underlying <<native-backend-object>>, where applicable.
+  Move construction and move assignment render the moved-from instance of
+  [code]#T# invalid.
+* _Equality:_ [code]#T# must be equality comparable in the host application.
+  Equality between two instances of [code]#T# (i.e., [code]#a == b#) must be
+  true if one instance is a copy of the other, and non-equality between two
+  instances of [code]#T# (i.e., [code]#a != b#) must be true if neither instance
+  is a copy of the other, in accordance with the above definition of a copy.
+  An instance of [code]#T# that has been invalidated by a move operation is not
+  considered to be a copy of any other [code]#T# instance.
```
etomzak, line 367: The text originally said,

> unless either instance has become invalidated by a move operation

which I couldn't parse in context. I *think* the sentence I've added here captures the intent.

adoc/chapters/programming_interface.adoc
Pennycook: I'm not 100% sure because I've not seen this named requirement before, but it says it applies to function objects.  I think `std::hash<T>` is what would satisfy `Hash` here, rather than `T`.

The wording in [unord.hash](https://eel.is/c++draft/unord.hash#:hash,instantiation_restrictions) might be easier to align with.  It looks like we could say something like:
> An enabled specialization `std::hash<T>` must exist in the host application.

Because an "enabled specialization" is defined in [paragraph 5](https://eel.is/c++draft/unord.hash#5):

> An enabled specialization `hash<Key`> will:
> - [(5.1)](https://eel.is/c++draft/unord.hash#5.1) meet the [Cpp17Hash](https://eel.is/c++draft/hash.requirements#:Cpp17Hash) requirements (Table [37](https://eel.is/c++draft/hash.requirements#tab:cpp17.hash)), with `Key` as the function call argument type, the [Cpp17DefaultConstructible](https://eel.is/c++draft/utility.arg.requirements#:Cpp17DefaultConstructible) requirements (Table [30](https://eel.is/c++draft/utility.arg.requirements#tab:cpp17.defaultconstructible)), the [Cpp17CopyAssignable](https://eel.is/c++draft/utility.arg.requirements#:Cpp17CopyAssignable) requirements (Table [34](https://eel.is/c++draft/utility.arg.requirements#tab:cpp17.copyassignable)), the [Cpp17Swappable](https://eel.is/c++draft/swappable.requirements#:Cpp17Swappable) requirements ([[swappable.requirements]](https://eel.is/c++draft/swappable.requirements)),
> - [(5.2)](https://eel.is/c++draft/unord.hash#5.2) meet the requirement that if `k1 == k2` is `true`, `h(k1) == h(k2)` is also `true`, where `h` is an object of type `hash<Key>` and `k1` and `k2` are objects of type `Key`;
> - [(5.3)](https://eel.is/c++draft/unord.hash#5.3) meet the requirement that the expression `h(k)`, where `h` is an object of type `hash<Key>` and `k` is an object of type `Key`, shall not throw an exception unless `hash<Key>` is a program-defined specialization[.](https://eel.is/c++draft/unord.hash#5.sentence-1)

...but I defer to @nliber!

adoc/chapters/programming_interface.adoc
```diff
@@ -340,10 +340,12 @@ to the following requirements, where [code]#T# is the runtime class type:
...
-  When any instance of [code]#T# is destroyed, any behavior specific to
-  [code]#T# that is specified as performed on destruction is only performed if
-  this instance is the last remaining host copy, in accordance with the above
-  definition of a copy.
+  When an instance of [code]#T# is destroyed, any copies of the [code]#T#
+  instance (as defined above) are unaffected.
+  When the last remaining host copy of an instance of [code]#T# is destroyed,
+  lifetime-related operations may be performed on the underlying
+  implementation-defined resource, including, where applicable, the destruction
+  of a <<native-backend-object>>.
```
etomzak, line 349: The way this was worded seemed to conflate the destruction of `T` and the destruction of the underlying managed resource. I've tried to clarify.

adoc/chapters/programming_interface.adoc
```diff
@@ -451,10 +453,13 @@ _Remarks:_
...
-  the <<backend>> specification document, if this [code]#T# instance was
-  originally constructed using one of the backend interoperability
-  [code]#make_*# functions specified in <<sec:backend-interoperability-make>>.
+  operations required by any underlying <<native-backend-object>> or any other
+  implementation-defined resource managed by [code]#T#.
+
+* If this [code]#T# instance originated from a backend interoperability
+  [code]#make_*# function specified in <<sec:backend-interoperability-make>>,
+  then any required lifetime-related operations are performed on the
+  corresponding <<native-backend-object>>.
```
etomzak, line 466: The way this was worded implied that lifetime-related operations were only relevant for `T` instances created using backend interop. I've split this up to describe the general case (no backend interop) and then the backend interop case. I think it is important to address the backend interop case specifically, because a reader might not realize that objects created with `make_*()` are destroyed using RAII rules.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,271 +288,410 @@ context doesn't match the target backend.
...
+  a#), symmetric (i.e., [code]#a == b# implies [code]#b == a# and [code]#a != b#
+  implies [code]#b != a#), and transitive (i.e., [code]#a == b && b == c#
+  implies [code]#c == a#).
+
+* _Hashability:_ A specialization of [code]#std::hash# for [code]#T# must exist
+  in the host application.
+  The specialization of [code]#std::hash# must return a unique value such that
+  for two instances of [code]#T#, [code]#a# and [code]#b#, either [code]#(a ==
+  b) && std::hash(a) == std::hash(b)# evaluates to true, or [code]#(a != b) &&
+  std::hash(a) != std::hash(b)# evaluates to true.
```
etomzak, line 378: The text didn't originally say that `operator==()` and `hash()` are available on device. Should we add those?

Pennycook: I don't think we should make that change here, because it could be a breaking one.

I'd like @AerialMantis to weigh in here.  My concern is that this could be deliberate, because `accessor` is weird, and that the original intent was to allow for `accessor` implementations that compare equal on the host but do not compare equal on the device.  I'm having trouble constructing a case where that would be useful behavior, though, so it could indeed just be an oversight.

adoc/chapters/programming_interface.adoc
gmlueck: I think we should go back to the original wording:

> ... any behavior specific to `T` that is specified as performed on destruction is only performed if this instance is the last remaining host copy

I think this is referring to the "Buffer destruction rules".  The actions described in that section do not occur each time a `buffer` object is destroyed.  They happen only if the object is the last remaining host copy.

I admit the original wording was not great, but I think the rewrite makes it even less clear.

adoc/chapters/programming_interface.adoc
gmlueck: I agree with @Pennycook.  Making these operations available on device goes beyond the purported scope of this PR, which is just to reformat and clarify the text.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,271 +288,410 @@ context doesn't match the target backend.
...
+  constructible and move assignable within SYCL kernel functions.
+  Any instance of [code]#T# that is the result of a move from another instance
+  must behave as-if it were the original instance and must represent the same
+  underlying <<native-backend-object>>, where applicable.
+  Move construction and move assignment render the moved-from instance of
+  [code]#T# invalid.
+
+* _Equality:_ [code]#T# must be equality comparable in the host application.
+  If [code]#T# is a valid kernel argument, [code]#T# must also be equality
+  comparablewithin SYCL kernel functions.
```
gmlueck: I believe this statement about equality comparison in device code was NOT present in the original text.

adoc/chapters/programming_interface.adoc
```diff
@@ -288,271 +288,410 @@ context doesn't match the target backend.
...
-@ Hidden friend function @ Description
-a@
-[source]
+_Remarks:_
+
+* Conforms to the destruction requirements of <<reference-semantics>>.
+
+* On destruction of the last copy, may perform additional lifetime-related
+  operations required by any underlying <<native-backend-object>> or any other
+  implementation-defined resource managed by [code]#T#.
```
gmlueck, line 461: I think we should NOT add the phrase:

> or any other implementation-defined resource managed by [code]#T#

This was not in the original text, and I don't think the spec should be talking about implementation details like this.

adoc/chapters/programming_interface.adoc
gmlueck: Rather than adding this bullet, can we instead add some text to the section "Template functions make_*" which clarifies that objects created by the `make_` functions still conform to the common reference semantics if T is a common reference semantic type?

adoc/chapters/programming_interface.adoc
etomzak: Can you be more specific about what you think is unclear? The wording in this bullet is based on the wording in the original bullet, as well as the `~T()` wording in Table 7.

IMO, the current SYCL 2020 wording is actively misleading:

* The original states: "When any instance of T is destroyed, [...] any behavior specific to T that is specified as performed on destruction is only performed if this instance is the last remaining host copy." This conflates the concept of C++-style object destruction and SYCL-style releasing of a managed resource. When an instance of T is destroyed, that instance of T is destroyed according to C++ destruction rules. Period. Under certain circumstances, destroying an instance of T triggers further, SYCL-defined behavior WRT managed resources in addition to C++ rules. Saying, "behavior specified on destruction is only performed for destructions under some specific circumstances," is essentially rewriting the C++ standard, which we can't do.
* The original states: "On destruction of the last copy, may perform additional lifetime related operations [...] if this SYCL T instance was originally constructed using one of the backend interoperability make_* functions." In fact, the lifetime-related operations may also be performed when the T instance was constructed without using a `make_*` function.

I don't understand your reference to [§4.7.2.3 Buffer destruction rules](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:buf-sync-rules). Yes, SYCL buffers are common reference types. The spec has extra info on buffer destruction, because there are complicated interactions with lifetimes and user data. How is that relevant here?

adoc/chapters/programming_interface.adoc
etomzak: Does the WG have guidance on when and how the spec should/shouldn't make reference to implementation-defined details?

adoc/chapters/programming_interface.adoc
etomzak: I'm reluctant to introduce changes to even more spec sections in this PR. This PR has already opened far too many cans of worms. I suggest reformatting the `make_*` and `get_native` sections in a separate PR, and then moving text between the sections in a third PR if it's seen to be necessary.

adoc/chapters/programming_interface.adoc
etomzak: You're right. It might be copy-pasta.

Removed in 7d8e673

adoc/chapters/programming_interface.adoc
```diff
@@ -288,271 +288,408 @@ context doesn't match the target backend.
...
+functions in order to fulfill copy, move, and destruction requirements and
+hidden friend functions in order to fulfill equality requirements, as described
+below.
 
-A hidden friend function is a function first declared via a [code]#friend#
-declaration with no additional out of class or namespace scope declarations.
-Hidden friend functions are only visible to ADL (Argument Dependent Lookup) and
-are hidden from qualified and unqualified lookup.
-Hidden friend functions have the benefits of avoiding accidental implicit
-conversions and faster compilation.
```
keryell, line 356: This disappeared. But that seems probably OK.

adoc/chapters/programming_interface.adoc
keryell:
```c++
((a != b) && std::hash(a) != std::hash(b)) == true
```
Isn't the definition of `std::hash` magical?
In the case of `a != b` we cannot say anything about `std::hash(a)` and `std::hash(b)`.
Could we just delegate this to the C++ standard?

tomdeakin:
The WG discussed this and decided to break this into two stages: first, strictly reformat without wording changes as one PR, and second, PRs for wording changes. With that in mind, we'll close this PR.
@TApplencourt may take this on, thanks!

PR #865, opened 2025-07-01
Update to the latest Docker image

gmlueck:
This docker image creates a slightly different HTML render vs. the previous one.  It appears that the rouge source code highligher styles floating point constants slightly differntly in source code listings. In a constant like "42.0f", the "f" is now in a different CSS class. Theoretically, that should cause it to have a different font or color or something, but I can't detect any difference from looking at the web page.

There is no difference in the PDF render.


gmlueck:
Merging as editorial.

PR #879, opened 2025-07-07
Deprecate `accessor::max_size()`

Pennycook:
This function was added only to align accessor with the container interface. However, an accessor is not a container, and this has led to confusion regarding what this function should return.

---

As per discussion in last week's SYCL WG call, I've targeted this at SYCL 2020 instead of SYCL-Next, despite it being a feature that was introduced in SYCL 2020.  Whether or not this function will be removed in SYCL-Next is to be determined later, but the WG decided it is useful to signal to developers that a function should not be used as early as possible.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -9453,7 +9453,7 @@ a@
 ----
 size_type max_size() const noexcept
 ----
-   a@ Deprecated.
+   a@ Deprecated by SYCL 2020.
 
 Returns the maximum number of elements any accessor of this type would be able
 to access.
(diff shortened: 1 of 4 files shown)


etomzak:
LGTM

adoc/chapters/programming_interface.adoc
```diff
@@ -9453,8 +9453,10 @@ a@
 ----
 size_type max_size() const noexcept
 ----
-   a@ Returns the maximum number of elements any accessor of this type would be
-      able to access.
+   a@ Deprecated.
```
```suggestion
   a@ Deprecated by SYCL 2020.
```

I think we need to be clear about which version of the spec deprecates this API.  Otherwise, it won't be clear whether it is deprecated in SYCL 2020 vs. SYCL Next (once we publish SYCL Next).

Pennycook: Good point.

PR #644, opened 2024-10-17
Add sycl_khr_free_function_commands extension

Pennycook:
This extension provides an alternative mechanism for submitting commands to a device via free-functions that require developers to opt-in to the creation of event objects.

It also proposes alternative names for several commands (e.g., `launch`) and simplifies some concepts (e.g., by removing the need for the `nd_range` class).

Issue #901, opened 2025-08-11
Do the types used for kernel name require external linkage?


tahonermann:
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interfaces.bundles
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_defining_kernels
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:naming.kernels

### Issue Description

SYCL 2020 implementations that compile source code separately for the host and each device (and thus predefine the `__SYCL_DEVICE_ONLY__` macro during device compilation and do not predefine the `__SYCL_SINGLE_SOURCE__` macro) are required to correlate kernel name types used in calls to SYCL kernel invocation functions with kernel symbols (e.g., a kernel entry point function or the call operator member of the kernel function object type) emitted during device compilation. This implementation strategy produces multiple translation units for each primary source file; one for the host compilation and one for each device compilation.

Types with internal linkage are local to a translation unit and cannot, in general, be correlated across translation units. Implementations typically emit symbols that involve types with internal linkage with visibility settings that make them ineligible for reference by other translation units and therefore exempt from globally unique naming requirements. An allowance for kernel name types with internal linkage creates a significant implementation challenge since 1) such types may have names that are used in multiple translation units, but denote distinct types and thus distinct kernels, 2) the SYCL implementation might need to dynamically resolve kernel symbols (e.g., to select a separately compiled device image or to JIT generate a device image for a run-time selected device; there is a 1-N correspondence of a kernel name type known during host compilation to the kernels produced during device compilation), and 3) the SYCL implementation might not be able to correlate host and device translation units in order to form a "translation unit bundle" that restricts which device translation units need to be searched to resolve a host kernel name type (the implementation strategy used by DPC++ is to map the kernel name type to a string identifier in the (instantiated) SYCL kernel invocation function and then pass it to internal support functions present in a linked library to resolve it to a kernel symbol by searching all known device images).

Consider the following example:
- `k1.cpp`:
    ```
    #include <sycl/sycl.hpp>
    namespace { struct KN; }
    void k1(sycl::handler &h) {
      sycl::stream sout(1024, 256, h);
      h.single_task<KN>([=]{
        sout << "Hello from k1()\n";
      });
    }
    ```
- `k2.cpp`:
    ```
[...]

Issue #900, opened 2025-08-08
Remove implementation-defined generation of a kernel name type


tahonermann:
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:invokingkernels

### Issue Description

SYCL 2020 section 4.9.4.2, "SYCL functions for invoking kernels" states:
> Each function takes an optional kernel name template parameter. The user may optionally provide a [kernel name](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#kernel-name), otherwise an implementation-defined name will be generated for the kernel.

Discussion in the SYCL WG in the context of #568 and #893 has revealed that programmers expect to be able to use the type of the kernel object as the kernel name type when a kernel name type is not explicitly provided. For example, programmers expect the following example to be well-formed and to behave the same for all implementations.
```
#include <sycl/sycl.hpp>

struct kernel {
  void operator()() const {}
};

int main() {
  kernel k;
  sycl::queue q;
  q.single_task(k);
  q.wait();
 auto kid1 = sycl::get_kernel_id<kernel>();      // Ok; retrieves the kernel ID for the invocation of k.
 auto kid2 = sycl::get_kernel_id<decltype(k)>(); // Ok; retrieves the kernel ID for the invocation of k.
}
```

This expectation extends to lambda closure types and it has been reported that kokkos contains code like the following:
```
int main() {
  auto k = []{};
  sycl::queue q;
  q.single_task(k);
  q.wait();
 auto kid2 = sycl::get_kernel_id<decltype(k)>(); // Ok; retrieves the kernel ID for the invocation of k.
}
```

This expectation seems reasonable subject to the reduced feature set limitations imposed on use of lambda expressions without an explicit kernel name type and the forward declarable kernel name type concerns raised in #857 and #858. If the forward declarable kernel name type requirements were moved to the reduced feature set as suggested in #859, then there would appear to be no technical reason not to specify that the kernel object type is always used as the implicit kernel name type for full feature implementations.

tahonermann:
[...]

Issue #893, opened 2025-07-22
Can multiple kernel name types be associated with the same kernel object type?


tahonermann:
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:naming.kernels

### Issue Description

Discussion in #568 raised a concern regarding whether multiple kernel name types can be associated with the same kernel object type. This concern is tangential to the primary concern for which that issue was reported, so has been moved to this new issue.

See the code example attached to this issue. The [initial concern](https://github.com/KhronosGroup/SYCL-Docs/issues/568#issuecomment-3088647971) raised in #568 is this:
> I don't think code like this should be allowed. `MyKernel` already has a name, and I can't think of any reason why somebody would want to assign two different names to the same kernel. Assigning two names to the same kernel means that in practice it will have to be compiled twice, even though the body of the kernel will be identical.
> 
> ... assigning two different names to the _same_ lambda seems broken to me, just like the named function object case.

[Further discussion](https://github.com/KhronosGroup/SYCL-Docs/issues/568#issuecomment-3091473649) indicated that the suggested duplicate compilation is limited to device entry points (if applicable to the implementation) and SYCL invocation function template instantiations.
> The kernel name is what defines the kernel, not the kernel object type. Greg's example requires two distinct device entry points, but I don't think it necessitates multiple compilations of the kernel object type or the functions it calls. Entry points can share common code. Template instantiations would of course be performed for each invocation of the SYCL invocation function with a distinct kernel name type.

[Discussion continued](https://github.com/KhronosGroup/SYCL-Docs/issues/568#issuecomment-3096065813) with an emphasis on design:
> I meant that the _design_ seems broken: it's definitely possible to implement something that allows a user to associate multiple arbitrary type names with the same function object, I just don't understand why that's useful to anybody. It makes the implementation more complicated, means we have to specify the behavior of more weird corner cases, and potentially leads to more confusion for developers.

The code example attached to this issue attempts to demonstrate how a single kernel object type might reasonably used for multiple purposes.

[...]

PR #922, opened 2025-10-10
Add sycl_khr_launch extension

slawekptak:
_This is a new, follow-up PR to [#644](https://github.com/KhronosGroup/SYCL-Docs/pull/644), originally created by John Pennycook. All the future work related to that PR will be continued here. The reason for creating a new PR is that the PR ownership transfer is required._

This extension provides an alternative mechanism for submitting commands to a device via free-functions that require developers to opt-in to the creation of event objects.

It also proposes alternative names for several commands (e.g., launch) and simplifies some concepts (e.g., by removing the need for the nd_range class).


adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1086 @@
...
+_Constraints_:
+
+* [code]#Requirements# does not contain a [code]#kernel_bundle#; and
+* [code]#Requirements# does not contain any accessors.
+
+_Effects_: Enqueues a command barrier.
+Any commands submitted after this barrier cannot begin execution until:
+
+* All commands previously submitted to this queue have completed; and
+* All requirements in [code]#reqs# are satisfied.
```
gmlueck:
```suggestion
* All events in [code]#reqs# have completed.
```

Since we disallow accessors in `reqs`, I think the only possible requirements are from events.  We may as well be more clear about this.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1086 @@
...
+
+'''
+
+.[apititle]#event_barrier#
+[source,role=synopsis,id=api:event_barrier]
+----
+namespace sycl::khr {
+
+template <typename... Requirements>
+std::optional<event> event_barrier(const queue& q, const requirements<Requirements...>& reqs = {});
```
gmlueck: I'd like to reconsider the name of this function.  Using the word "event" in the name makes little sense.  Neither this function nor `command_barrier` above directly take any `event` objects as input.  However, *both* this function and `command_barrier` do indirectly take `event` objects via `requirements`.  In fact, both functions have exactly the same semantics regarding the input event(s) -- in both cases the barrier waits until all the events are complete.

Therefore, it is not the event semantics that distinguishes this API from `command_barrier`.  The difference is that `command_barrier` implicitly waits for all previous commands in the queue to complete, while `event_barrier` does not.

At one point we had proposed `partial_barrier` as the name for this function.  Can anyone remember why we didn't like that?  That seems like a better name to me than `event_barrier`.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1086 @@
...
+
+template <typename T, typename... Requirements>
+std::optional<event> launch_host(const queue& q, T&& hostTaskCallable); (2)
+
+}
+----
+
+_Constraints_:
+
+* [code]#Requirements# does not contain a [code]#kernel_bundle#.
```
gmlueck: John originally proposed in [this comment](https://github.com/KhronosGroup/SYCL-Docs/pull/644#discussion_r2153057798) that there should be a constraint here that `Requirements` does not have any accessors with target `target::device`.  However, this depends on what semantic we want for this `launch_host` function.

If we expect that `launch_host` has all the same features as `handler::host_task`, then we need to allow both `target::host_task` and `target::device` accessors because `target::device` accessors have a special interaction with the `interop_handle`.  However, we have noted in the past that it might be better to split out the `interop_handle` part of host tasks into a separate API.

If our goal is to provide a cleaner semantic of this `launch_host` command, then maybe we only want it to support the case where it runs pure host code and eliminate the backend interop part.  We could add that later as a separate API.  If we go this route, then we do probably want a constraint here that `Requirements` does not have any accessors with target `target::device`

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1086 @@
...
+* [code]#Requirements# does not contain any accessors.
+
+_Effects_: Enqueues an event barrier.
+Any commands submitted after this barrier cannot begin execution until all
+requirements in [code]#reqs# are satisfied.
+
+{note}If an [code]#event_barrier# is submitted with no requirements, then this
+operation may be a no-op.{endnote}
+
+'''
```
gmlueck, line 1213: Adding a comment here from #644, so it doesn't get lost ... @aelovikov-intel noted in [this comment](https://github.com/KhronosGroup/SYCL-Docs/pull/644#discussion_r2363633933) that we should add an Examples section illustrating how to use buffers and accessors with this new KHR.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1086 @@
...
+
+// Launch a basic parallel_for with a function object.
+// New form of queue::parallel_for(range, ...)
+.[apititle]#launch# (kernel function)
+[source,role=synopsis,id=api:launch]
+----
+namespace sycl::khr {
+
+template <typename KernelType, typename... Requirements>
+std::optional<event> launch(const queue& q, range<1> r, const requirements<Requirements...>& reqs, const KernelType& k); (1)
```
gmlueck: Adding a comment from #644, so it doesn't get lost ... @aelovikov-intel proposed in [this comment](https://github.com/KhronosGroup/SYCL-Docs/pull/644#discussion_r2363630744) that we should change the type of all the `const KernelType&` parameters to `KernelType&&`.

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: I think `partial_barrier` sounds good. We can also consider `command_event_barrier `and `event_barrier`. These seem more descriptive and reflect the actual behavior.

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: I have assumed the standard and interop versions integrated into one, consistent with the handler version, but I can see, that it might be cleaner to separate them. We might then have something like `launch_host_interop`, making it a separate use case, and consistent with the `interop_handle` class name. What do you think?

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: This change is done in DPC++, for the launch_grouped functions (handler-less), and if the kernel copy is required, it is moved if possible. This might be an opportunity to do some testing if needed.

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: Limit host tasks accessors to target::host_task done here: [166bd46](https://github.com/KhronosGroup/SYCL-Docs/pull/922/commits/166bd462dffffd4388a98aecd7c69b75d0373fbe)

tomdeakin:
The WG discussed this, and feel we need a solution for local memory in this KHR.

PeterTh:
Regarding local memory: to me, it seems like the least invasive strategy (as in, it doesn't depend on many other changes) that fits with the current specification of this extension would be using requirements for local accessors - since it's a natural fit with how non-local accessors are proposed to be handled. A future extension for e.g. static work group memory could then make that superfluous where it applies.

adoc/extensions/sycl_khr_free_function_commands.adoc
TApplencourt: AFAIK, It did predate the requirements. But now that we have requirement, do we still need both function?

Can we just add a `bool` to `command_barrier` like `implicit_all` or something?


If not indeed `partial_barrier` sound better.

adoc/extensions/sycl_khr_free_function_commands.adoc
TApplencourt: My 0.002$ is the same for the `local_accessor`, i should we should restrain this PR to do to much (and it doesn't already a lot). IMO It should be "easy" for people to move to this new KHR, just changing their API call and not their kernel.

>If we expect that launch_host has all the same features as handler::host_task

So Yes, I will vote for that. 

If one day we want to refactor host_task, to create a new one, we can do that with another API call. 


>If our goal is to provide a cleaner semantic of this launch_host command, 

I think it's our goal too, but maybe not in the PR.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+  khr::launch_grouped(myQueue, global, local, reqs, [=](nd_item<2> it) {
+    size_t i = it.get_global_id(0);
+    size_t j = it.get_global_id(1);
+    size_t index = i * M + j;
+    c[index] = a[index] + b[index];
+  });
+
+  // Wait for all three kernels to complete before accessing the results.
+  // This blocks the host until all previous kernels have completed.
+  e->wait();
```
gmlueck:
```suggestion
  e.wait();
```

I think.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+|Description
+
+|1
+|Initial version of this extension.
+|===
+
+== Usage example
+
+The example below rewrites the application from
+<<subsec:example.sycl.application>> to demonstrate the usage of this extension.
```
gmlueck: It seems like this example is not a direct rewrite of the code in chapter 2.  For example, the code in chapter 3 uses `buffer`, but this example uses USM.  Is that intentional?  If so, we should change this intro paragraph to mention these other changes.  However, it seems like it might be more useful if this example was a direct rewrite of the code in chapter 2 without any changes other than moving to the new APIs in this extension.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+.[apititle]#launch# (kernel function)
+[source,role=synopsis,id=api:launch]
+----
+namespace sycl::khr {
+
+template <typename KernelType, typename... Requirements>
+void launch(const queue& q, range<1> r, const requirements<Requirements...>& reqs, const KernelType& k); (1)
+
+template <typename KernelType, typename... Requirements>
+void launch(const queue& q, range<2> r, const requirements<Requirements...>& reqs, const KernelType& k); (2)void
```
gmlueck:
```suggestion
void launch(const queue& q, range<2> r, const requirements<Requirements...>& reqs, const KernelType& k); (2)
```

Looks like a typo?

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+
+}
+----
+
+_Constraints_:
+
+* [code]#Requirements# does not contain a [code]#kernel_bundle#;
+* Any accessor in [code]#Requirements# must have a target of
+[code]#target::host_task#; and
+* The host function does not make any SYCL API calls.
```
gmlueck: This cannot be a _Constraint_.  The _Constraints_ section has a very specific meaning in the C++ standard -- it is a set of constraints on the template parameters that can be checked at compile time.  If the template parameters do not meet these constraints, the compiler does not consider the function as a candidate when resolving the overload set for a call site.  The _Constraints_ are normally implemented either using the `requires` keyword from C++20 or by using SFINAE.

I think instead that this could be specified as a _Precondition_:

> _Preconditions_: The callable function `hostTaskCallable` does not call any function in the `sycl` namespace.

The C++ specification describes _Preconditions_ as "conditions that the function assumes to hold whenever it is called; violation of any preconditions results in undefined behavior".

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+Any commands submitted after this barrier cannot begin execution until:
+
+* All commands previously submitted to this queue have completed; and
+* All events (wrapped in wait_event or wait_events) in [code]#reqs# have
+  completed.
+
+{note}If a barrier is submitted to an in-order queue with no requirements, then
+this operation may be a no-op.
+If an event wrapped in signal_event was provided in [code]#reqs#, it will be
+signaled.{endnote}
```
gmlueck: This sentence about `signal_event` should not be in a non-normative `{note}`.  The non-normative notes are intended to only clarify things that are formally defined elsewhere in the spec.  I think we instead need to clarify the wording above to describe when the event is signaled:

> _Effects:_ Enqueues a command barrier.
> Any command submitted after this barrier cannot begin execution until the barrier completes.
> The barrier completes when:
>
> * All commands previously submitted to this queue have completed; and
> * All events (wrapped in `wait_event` or `wait_events`) in `reqs` have completed.
>
> If `reqs` contains a `signal_event` requirement, the event is signaled when the barrier completes.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+Any commands submitted after this barrier cannot begin execution until:
+
+* All commands previously submitted to this queue have completed; and
+* All events (wrapped in wait_event or wait_events) in [code]#reqs# have
+  completed.
+
+Once the above conditions are met, signals the provided event ([code]#e#).
+
+{note}The event is disassociated from any previous command it was
+tracking.{endnote}
```
gmlueck: Here too ... this should be normative text, not a non-normative note.  Also, I think it's important to say that the event is immediately disassociated from any previous command.  I'd suggest borrowing the wording we have in the DPC++ extension in [sycl_ext_oneapi_reusable_events](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_reusable_events.asciidoc).

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+
+template <typename... Requirements>
+void enqueue_signal_event(const queue& q, const event& e, const requirements<Requirements...>& reqs = {});
+
+}
+----
+_Constraints_:
+
+* [code]#Requirements# does not contain a [code]#kernel_bundle#; and
+* [code]#Requirements# does not contain any accessors.
```
gmlueck: Shouldn't we also add:

> * `Requirements` does not contain `signal_event`.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+namespace sycl::khr {
+
+template <typename... Requirements>
+void enqueue_wait_event(const queue& q, const event& e, const requirements<Requirements...>& reqs = {});
+
+}
+----
+_Constraints_:
+
+* [code]#Requirements# contains the [code]#signal_event# object; or
```
gmlueck: I think you mean that `Requirements` contains no other requirements other than `signal_event`.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+'''
+
+=== Event-related helper functions
+
+.[apititle]#make_event#
+[source,role=synopsis,id=api:make_event]
+----
+namespace sycl::khr {
+
+event make_event(context c);
```
gmlueck:
```suggestion
event make_event(const context& c);
```

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1161 @@
...
+
+}
+----
+_Effects_: Creates an event associated with a given context.
+The event can then be used with [code]#enqueue_signal_event# function, with
+signal_event object and as a dependency for other commands.
```
gmlueck: It's better to use _Returns_ when describing the return value of a function.  We should also specify that the event is in the "signaled" state:

> _Returns:_ An event that is associated with context `c`.  The event is in the "signaled" state.

Somewhere in this KHR we need a precise description of when the event's context needs to match the queue's context.  See the DPC++ extension section "Interaction with other event APIs" in [sycl_ext_oneapi_reusable_events](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_reusable_events.asciidoc).

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -0,0 +1,1162 @@
...
+
+.[apititle]#launch_host#
+[source,role=synopsis,id=api:launch_host]
+----
+namespace sycl::khr {
+
+template <typename T, typename... Requirements>
+void launch_host(const queue& q, const requirements<Requirements...>& reqs, T&& hostTaskCallable); (1)
+
+template <typename T, typename... Requirements>
```
gmlueck:
```suggestion
template <typename T>
```

I assume this is a copy-paste error?

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: Yes exactly - only `signal_event` is allowed, but it is optional. Do you think that the syntax "contains the `signal_event` or is empty" is not precise?

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: It seems that there is a lot in common between the reusable events extension and this one, so I've added a new section describing the interactions between the event APIs similar to "Interaction with other event APIs", but I've included not only interactions of the APIs from this KHR extension and already existing APIs, but also interactions between the APIs added in this KHR extension. This should cover all of the possible use cases.

adoc/extensions/sycl_khr_free_function_commands.adoc
gmlueck: What if `Requirements` contained both `signal_event` and `wait_event`?  According to your wording this would be allowed because "[code]#Requirements# contains the [code]#signal_event# object" is true.

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -1069,7 +1075,7 @@ signaled.{endnote}
 namespace sycl::khr {
 
 template <typename... Requirements>
-void enqueue_signal_event(const queue& q, const event& e, const requirements<Requirements...>& reqs = {});
+void enqueue_signal_event(const queue& q, event& e, const requirements<Requirements...>& reqs = {});
```
gmlueck: I'm not sure about this.  I think we want to take `event` by const reference.  The logic is the same as described here:

https://members.khronos.org/wg/SYCL/document/34373

adoc/extensions/sycl_khr_free_function_commands.adoc
```diff
@@ -1151,12 +1161,39 @@ if provided.
...
+The event is in the "signaled" state.
 
-'''
+{note}The event can be passed to the [code]#enqueue_signal_event# function or
+included in the [code]#requirements# object (wrapped in the [code]#signal_event#
+object), and as a dependency for other commands.{endnote}
+
+== Interactions between event APIs
+
+An event [code]#e# created via [code]#make_event# can be used as a command
```
gmlueck: This is kind of a style nit, but I think it's better to use a capital case italic letter when referring to an abstract event or command.  Only use code font when referring to a specific API name or a specific parameter name.  So, I would use `_E_` and `_C_` in this section instead of `[code]#e#` and `[code]#c#`.

adoc/extensions/sycl_khr_free_function_commands.adoc
slawekptak: The barrier function naming has been established, so closing this conversation.

keryell (approved):
Thanks!

tdavidcl (commented):
Very nice! This extension will solve many pain point that we have with our current SYCL usage, it looks great.

I just had kind of neatpick. In the examples i did not find guidelines related to the use of local memory or anything that require the `sycl::handler` since this extension hides it away. Is the use of free function for commands limited to kernels that do not employ local memory ? If yes are there plans to lift such limit ?

Also i don't recall seeing any named kernel in the proposed extension. Can we name them for profiling (#1002 maybe ?) ?

gmlueck:
> In the examples i did not find guidelines related to the use of local memory or anything that require the `sycl::handler` since this extension hides it away. Is the use of free function for commands limited to kernels that do not employ local memory ? If yes are there plans to lift such limit ?

Yes. the working group has discussed this.  We plan to add a KHR similar to this oneAPI extension:

https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_scratch_memory.asciidoc

And I think we might consider a KHR similar to this as well:

https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_static.asciidoc

tdavidcl:
Thanks for the clarification.

> https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_scratch_memory.asciidoc

I don't really like the approach as one has to rebuild aliases every time to use it. It looks overly complex for what it aims to achieve and also probably for people coming from other portability solutions / CUDA & HIP, ... What is the motivation behind that design rather than the second one ?

> https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_static.asciidoc

This one on the other hand would be wonderful to have. And if we want to use it as scratch we can always declare a char[] local  as one would do in e.g. CUDA. Also i would favor it since it is the closest to the other GPU backends.

gmlueck:
> I don't really like the approach as one has to rebuild aliases every time to use it. It looks overly complex for what it aims to achieve and also probably for people coming from other portability solutions / CUDA & HIP, ... What is the motivation behind that design rather than the second one ?

Why do you think it's more complex or harder to use than the CUDA API where you declare an unbounded array?  We actually designed this API as a cleaner version of the CUDA one.  One thing we don't like about the CUDA interface is that multiple unbounded array declarations alias each other, which seems very non-obvious.  For example:

```
extern __shared__ float data1[];
extern __shared__ char data2[];
```

This looks like two different arrays of shared memory (work-group local memory).  However, the two arrays `data1` and `data2` actually alias each other and point to exactly the same memory.

We think the proposed SYCL API makes this less surprising because there's just a single global function that returns a pointer to the start of the dynamically-sized local memory.

> I don't really like the approach as one has to rebuild aliases every time to use it.

What do you mean by "rebuild aliases"?  Do you mean that you need to cast the pointer to a particular type?  Would it help if we made the function templated like this:

```
template<typename T>
T* get_work_group_scratch_memory();
```

Then, you would use it like:

```
auto pscratch = sycl::khr::get_work_group_scratch_memory<int>();
```

tdavidcl:
Oh i see now why.

It's hard to always remember what is what between CUDA and SYCL.

My confusion was related to "Static Shared Memory" in CUDA. which does resemble a lot more the second extension and is the most used one in CUDA. Whereas the "Dynamic Shared Memory" variant corresponding  work_group_scratch_memory is quite rare in the type of apps i'm working with.

> Do you mean that you need to cast the pointer to a particular type? Would it help if we made the function templated like this

Yup, this syntax is much cleaner i think, with the static cast it would have made a very long line.
```c++
int* pscratch = sycl::khr::get_work_group_scratch_memory<int>();
```
which is already 65 chars so it allows for 3 level of indentation (4 space) before passing the 80 columns mark.

We are still forced to do some weird business though if we have multiple types similarely to cuda (there is probably some memory alignment to account for but not yet :sweat_smile: )
```cuda
extern __shared__ int s[];
int *integerData = s;                        // nI ints
float *floatData = (float*)&integerData[nI]; // nF floats
char *charData = (char*)&floatData[nF];      // nC chars
```
I don't know if there is a clean way to handle this ... maybe this ? But a clean wrapper that does it automatically would be wonderfull.


```c++
int *integerData = sycl::khr::get_work_group_scratch_memory<int>(); // nI ints
float *floatData = reinterpret_cast<float>(integerData + nI);   // nF floats
char *charData   = reinterpret_cast<char>(floatData + nF);      // nC chars
```

That probably goes beyond the scope of the discussion of this PR though :sweat_smile: 

PS: why name it scratch instead of local in the name since it refers to local memory ?

gmlueck:
> I don't know if there is a clean way to handle this ... maybe this ? But a clean wrapper that does it automatically would be wonderfull.

I think you'd need to do something similar to what you show.  I'm not sure about providing a "clean wrapper".  I think a wrapper like this would be very complicated, and it would be easier to just do the pointer math like you show.


> PS: why name it scratch instead of local in the name since it refers to local memory ?

I think we were trying to use the term "work-group memory" instead of "local memory" because that seemed clearer.  Therefore, we were trying to avoid using the word "local" in the API name.

tdavidcl:
> I think we were trying to use the term "work-group memory" instead of "local memory" because that seemed clearer. Therefore, we were trying to avoid using the word "local" in the API name.

I get this one, but since all of the doc and SYCL tutorials outline it as local memory (= work group local memory) in the memory hierarchy of the model wouldn't it be too confusing to introduce a new name. Also
```
int* pscratch = sycl::khr::get_local_scratch_memory<int>();
// or
int* pscratch = sycl::khr::get_local_memory_pointer<int>();
```
is shorter :stuck_out_tongue:

tahonermann:
I have implementation concerns regarding this proposal. In particular, I'm concerned about the ability for implementations that use separate host and device compilation steps (like DPC++) to correlate kernel definitions across host and device compilation in a way that doesn't results in spurious kernel identification conflicts. See #893 for examples where the type of a named function object is insufficient to differentiate use in distinct kernel invocation calls.

@TApplencourt, I'd like to resume kernel naming/identification discussions before this proposal moves forward.

TApplencourt:
Totally agree, the naming/identification discussions are support important! 

We should see if we can use `compile-time property` if we can solve your issue. A presentation to help drive discusion will be super nice _nudge, nudge_

For `sycl_khr_free_function` my naive recollection was that it was possible to implement it as a "layered" implementation on top of "old submit mechanism" -- minus compile time property. I guess I was wrong :)

tahonermann:
@TApplencourt,
> We should see if we can use compile-time property if we can solve your issue. A presentation to help drive discusion will be super nice _nudge, nudge_

I'm happy to provide a simple presentation of my concerns if you want to schedule it for a future SYCL WG meeting (not tomorrow, please).

> For `sycl_khr_free_function` my naive recollection was that it was possible to implement it as a "layered" implementation on top of "old submit mechanism" -- minus compile time property. I guess I was wrong :)

That isn't wrong; it can be layered on top, but naively doing so preserves the many problems we already have with kernel naming. If we are going to provide new kernel invocation interfaces, I'd like to try to fix as many of those issues as we can. I would put #893 at the top of the list of issues to resolve (especially with respect to class types that can be used as multiple kinds of kernels as in the examples with overloaded member call operators). Next up would be #900 and #858 since the inability to specify an explicit kernel name imposes a requirement on the implementation to do so which prevents some implementation strategies currently in use (e.g., use of a SYCL unaware host compiler in conjunction with a device compiler). Issue #901 would then be next on the list.

TApplencourt:
Oh I see. We should take the opportunity of having a new launch API to fix implementation problem here. That make sense.

>SYCL WG meeting (not tomorrow, please).

You start to know me, and avoid my traps :( 
Just let me know when you are ready :) And thanks again for tackling those issues!

gmlueck:
I confirm that the contribution in this PR (PR#922) by GitHub user Pennycook was made in the course of their employment at Intel at the time of contribution, and I am authorized to confirm on the company's behalf that this contribution is covered under Intel's Khronos Specification Mixed Repository Contributor License Agreement.

Signed: Gregory Lueck, SYCL Language Architect, Intel Corporation
Date: September 17, 2026

nliber (approved):

keryell:
> I confirm that the contribution in this PR (PR#922) by GitHub user Pennycook was made in the course of their employment at Intel at the time of contribution, and I am authorized to confirm on the company's behalf that this contribution is covered under Intel's Khronos Specification Mixed Repository Contributor License Agreement.
> 
> Signed: Gregory Lueck, SYCL Language Architect, Intel Corporation Date: September 17, 2026

I have asked my colleague @Pennycook to sign the CLA.

PR #925, opened 2025-10-30
Clarify async error handling priorities

### LINKED ISSUE #908: queue and async exception handler lifecycles.
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

4.13.1.1 Asynchronous error handler.

4.6.5 Queue class

### Issue Description

Upon ~queue destruction, a queue created with an async exception handler will have that handler called with any unconsumed errors.
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsubsec:exception.async

Also, a queue may be destroyed even when there are uncompleted commands that have been submitted to the queue.
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:queue-class

Given those, what is the exact expectation here:
```
sycl::event e;
{ // open a closure
  std::vector<exception> vec;
  auto async_exception_handler [&vec](sycl::exception_list el){ /* rethrow */ el.insert(someException); };
  sycl::queue q(async_exception_handler);
  e = q.submit( ... );
} // end closure.  
e.wait_and_throw();  // <-- trouble
```
If an error occurs immediately, it'll be recorded when the ~queue destructor occurs at the end of the closure.  But if the error occurs later, then the async exception handler will be called on the line with // trouble, when the exception handler and the vec are out of scope and no longer valid. 

Is the user supposed to anticipate the lifecycle needs and make sure the async handler's lifecycle is longer than the queue it is passed to?  Or once the ~queue is destroyed, then any event associated with an operation it set in motion no longer to be called?

### Code Example (Optional)

_No response_

gmlueck:
The behavior was previously unclear in the case when a call to `event::wait_and_throw` reports an async error and the queue no longer exists.

Closes #908

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17444,10 +17444,9 @@ The queue and context classes can optionally take an asynchronous handler object
 <<async-handler>> on construction, which is a callable such as a function class
 or lambda, with an [code]#exception_list# as a parameter.
 Invocation of an <<async-handler>> may be triggered by the queue member
-functions [code]#queue::wait_and_throw()# or
-[code]#queue::throw_asynchronous()#, by the event member function
-[api]#event::wait_and_throw#, or automatically on destruction of a queue or
-context that contains unconsumed asynchronous errors.
+functions [api]#queue::wait_and_throw# or [api]#queue::throw_asynchronous#, by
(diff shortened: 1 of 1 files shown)


tomdeakin:
No CTS change required.

PR #927, opened 2025-10-30
[KHR] add sycl_khr_queue_empty_query

gmlueck:
Cherry pick #700 from main
Cherry pick #817 from main


gmlueck:
Merging because:

* The WG already approved #700
* Change #817 was editorial
* The Khronos board has ratified sycl_khr_queue_empty_query, so we can now publish it

PR #933, opened 2025-11-03
Removing contradictory statement about context

gmlueck:
This sentence is contradictory because we say in section 4.6.2 "Platform class":

> Each platform has an associated default context which contains all of
> the root devices in the platform.

Therefore, it must be that all devices in the same platform *can* be grouped into the same context.


tomdeakin:

PR #814, opened 2025-05-13
[KHR] Provide an alternative to <sycl/sycl.hpp>

AlexeySachkov:
This PR aims to address KhronosGroup/SYCL-Docs#780 by introducing a new KHR extension that document more headers that SYCL applications can include for better control of what they pay for in terms of compile time.

PR #952, opened 2026-01-16
Render preprocessing statements in different color

gmlueck:
@KornevNikita noted in #814 that comments and preprocessing statements have the same color in our source code listings, which looks ugly. Change the color of preprocessing statements to be something different.


gmlueck:
Merging as editorial

Issue #956, opened 2026-02-02
Clarify whether SYCL_EXTERNAL behaves like GNU-style or C++-style attribute


illuhad:
### Specification Version

SYCL 2020 (Revision 11)

### Section Number(s)

5.10.1

### Issue Description

Currently, as far as I can see, we do not explicitly say whether `SYCL_EXTERNAL` should follow GNU-style (`__attribute__((attr))`) or C++-style (`[[attr]]`) placement rules.

clang does not accept these in exactly the same places.

The spec currently only says 

> In order to declare a function with SYCL_EXTERNAL, the macro name SYCL_EXTERNAL must appear before the function declaration. If the function is also decorated with C++ attributes that appear before the declaration, the SYCL_EXTERNAL may appear before, after, or between these attributes.

Which *could* be interpreted as `SYCL_EXTERNAL` following C++-style attributes, but it does not seem very clear to me.

This issue has immediate practical relevance:
* AdaptiveCpp implements `SYCL_EXTERNAL` using `[[attr]]` syntax
* DPC++ implements `SYCL_EXTERNAL` using `__attribute__((attr))` syntax.

which already causes ecosystem friction. For example, the boost C++ libraries, when they detect a SYCL compiler, enable `SYCL_EXTERNAL` on some functions in such a way that only works with DPC++/GNU-style attributes.
Because of this, it is currently not possible to use recent versions of boost with AdaptiveCpp, even if kernels don't even use boost functionality since incorrect attribute usage is diagnosed early by clang.

See https://github.com/AdaptiveCpp/AdaptiveCpp/issues/1957

@TApplencourt Can we put this on the agenda?

tahonermann:
The in-progress Clang implementation uses the C++11 attribute syntax. See https://godbolt.org/z/zrYT6cMsE.

I would argue that the Boost code violates the SYCL specification. The example in [SYCL 2020 section 5.10.1, "SYCL functions and member functions linkage"](https://eel.is/c++draft/dcl.spec.general#nt:decl-specifier), indicates that `SYCL_EXTERNAL` is to be placed before `extern`. Since `extern` and `inline` occupy the same place in the C++ grammar (see *decl-specifier* in [\[dcl.spec.general\]](https://eel.is/c++draft/dcl.spec.general#nt:decl-specifier)), it is reasonable to conclude that the same position is required relative to the `inline` specifier.

[...]

PR #959, opened 2026-02-12
Fix example and definition of SYCL_EXTERNAL

TApplencourt:
Fix issue #956


illuhad (approved):
Looks good to me -- let's see what the C++ lawyers think :-)

nliber (approved):

adoc/code/sycl-external.cpp
gmlueck, line 1: I thought the formatting of this example looked pretty ugly in the HTML and PDF rendering.  How about this (19fe6531fd556e21c7f71b1de84dbfcf79e27bde)?  With the new format, each declaration is on its own line.

Apologies for pushing to your PR, but I thought it was a lot easier to suggest a new format via a commit than trying to describe it in a comment.

adoc/chapters/device_compiler.adoc
```diff
@@ -888,10 +888,11 @@ device code.
 The macro has no effect when the function is called from host code.
 
 In order to declare a function with [code]#SYCL_EXTERNAL#, the macro name
-[code]#SYCL_EXTERNAL# must appear before the function declaration.
-If the function is also decorated with {cpp} attributes that appear before the
-declaration, the [code]#SYCL_EXTERNAL# may appear before, after, or between
-these attributes.
+[code]#SYCL_EXTERNAL# must appear after the template head (i.e., template <...>)
```
gmlueck:
```suggestion
[code]#SYCL_EXTERNAL# must appear after the template head (i.e., [code]#+template <...>+#)
```

Let's use code format here.  (Note the `+` escape so that the three dots are rendered verbatim instead of an ellipsis.)

adoc/chapters/device_compiler.adoc
```diff
@@ -888,10 +888,11 @@ device code.
...
+if present, and before the function's return type or declaration specifiers
+(such as inline or static).
```
gmlueck:
```suggestion
(such as [code]#inline# or [code]#static#).
```

Here too I think.

adoc/code/sycl-external.cpp
gmlueck: Oh, I see.  This format causes the `clang-format` CI to fail.  I think we should find some way to override the `clang-format` in this case.  Do you know how to do that?

TApplencourt: Hum, the good thing about clang-format is that we don't need argue about style. I'm afraid to open a can of worm if we allow exception...  But I agree it's ugly, we can try to fix the clang-format potentially.

Adding black list is a little hard as we do in the CI
```
git diff -U0 --no-color ${{ github.event.pull_request.base.sha }} ./adoc/code/ | ./adoc/scripts/clang-format-diff.py -p1
```
So we don't know where the diff come from.

TApplencourt: But thinking 10s more, I suppose we can ask `git diff` to only print the diff on some file excluding other if we want to go this path (even if it's a 2 pass process).

nliber: It's hard to render it in my head. :-)

gmlueck: @TApplencourt The exclusion list seems easier to me, but I'm happy with either option so long as `clang-format` doesn't get in the way of formatting our examples as we want them.

TApplencourt: It's just a "long line" problem. We can increase the `ColumnLimit` to `100` which is the new "norm" used by lot of project (LLVM for example) from the old default of `80`.

I will try and see how invasive the code-change is. But we apply the `clang-diff` only on `diff` so old code will not be impacted AFAIK.

TApplencourt: The most invasive change is the `AlwaysBreakTemplateDeclarations: MultiLine` for Template.

If applied everywhere (witch we are not doing just to be super clear, we just format the diff), this new clang-format will transform (in `queue.hpp`):
```
   template <typename T>
   event copy(const T* src, T* dest, std::size_t count);
```
 Into
 ```
   template <typename T> event copy(const T* src, T* dest, std::size_t count);
   ```
   
   If we want to avoid that, we can just change the example of SYCL_EXTERNAL to have
   ```
   template <typename T>
   SYCL_EXTERNAL void D();    // OK
   template <> 
   SYCL_EXTERNAL inline void D<int>();  // OK
```

In this case we can go back to our old `AlwaysBreakTemplateDeclarations: True`. I don't mind either way, I'm just in team "we should automatically format, it make the spec look coherent".

gmlueck: > If we want to avoid that, we can just change the example of SYCL_EXTERNAL to have

I think this is OK, but we should align the `// OK` comments on those two lines.

TApplencourt: Doesn't seem possible when the comments are not directly on contiguous line.

gmlueck (approved):

PR #960, opened 2026-02-12
[KHR] Prefetch host

TApplencourt:
Replace old https://github.com/KhronosGroup/SYCL-Docs/pull/280

Add `prefetch_host` with the new format. The text and layout are based of the `prefetch` code to keep consistency.


illuhad (approved):
LGTM!

Sidenote: AdaptiveCpp already implements this exact API; the only difference is that we don't have yet the `khr_` prefix and the right feature test macro.
https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/extensions.md#acpp_ext_prefetch_host

adoc/extensions/sycl_khr_prefetch_host.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+The example below demonstrates the usage of this extension.
+
+[source,,linenums]
+----
+#include <sycl/sycl.hpp>
+
+int main() {
+  sycl::queue Q;
+  int *A = Q.malloc_shared<int>(1,Q);
```
gmlueck, line 99: There is no such member function to `queue`.

adoc/extensions/sycl_khr_prefetch_host.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+|1
+|Initial version of this extension.
+|===
+
+[[sec:khr-prefetch-host-handler]]
+== Extensions to the handler class
+
+This extension adds the following function to the [code]#sycl::handler# class.
+
```
gmlueck, line 30: I think we should show a high level synopsis here of the additions to the `handler` class like:

```
namespace sycl {
class handler {
  void khr_prefetch_host(void* ptr, size_t numBytes);
  // ...
};
}
```

This is consistent with the other KHR's that add member functions:

* https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:khr-default-context-platform
* https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:khr-queue-flush-queue

adoc/extensions/sycl_khr_prefetch_host.adoc
```diff
@@ -0,0 +1,103 @@
...
+free to migrate the allocation to the host.
+{endnote}
+
+'''
+
+[[sec:khr-prefetch-host-queue]]
+== Extensions to the queue class
+
+This extension adds the following function to the [code]#sycl::queue# class.
+
```
gmlueck, line 57: Also here, add a high-level synopsis of the member functions added to `queue`.

adoc/extensions/sycl_khr_prefetch_host.adoc
TApplencourt: I did it but then remove it due to 
- https://github.com/KhronosGroup/SYCL-Docs/blob/main/adoc/extensions/sycl_khr_queue_empty_query.adoc

Can put it back if you prefer

adoc/extensions/sycl_khr_prefetch_host.adoc
TApplencourt: I'm ashamed : ( Good catch!

adoc/extensions/sycl_khr_prefetch_host.adoc
gmlueck: I think sycl_khr_queue_empty_query is the outlier.  We should add the high level synopsis there too to make them all consistent.

nscottnichols:
As a SYCL user, I'm in favor of accepting this sooner than other larger extensions (that are in some sense more disruptive to the specification) since it is small and self contained. As others have mentioned, adding this to the specification will speed up adoption.

gmlueck:
For those who are asking for this KHR to move forward in order to speed up adoptions, which SYCL implementation(s) are you using?

If this includes DPC++, then I wanted to explain my hesitancy.  DPC++ currently provides a way to prefetch to host, but it has no effect when using the Level Zero backend because there is not yet support in Level Zero to do this prefetch.  Therefore, I don't see a lot of value for DPC++ to implement a standalone "prefetch host" KHR in the short term because it would just be a no-op anyway on Level Zero.  By the time we have Level Zero support, I expect that the WG will have approved #922, so I'd just as soon wait until then to have a prefetch-host KHR.

TApplencourt:
@nscottnichols , @tdavidcl  if you don't mind chiming in : ) Thanks!

adoc/extensions/sycl_khr_prefetch_host.adoc
```diff
@@ -0,0 +1,103 @@
...
+[[sec:khr-prefetch-host-queue-member-funcs]]
+=== Member functions
+
+.[apidef]#queue::khr_prefetch_host#
+[source,role=synopsis,id=api:queue-khr-prefetch-host]
+----
+event khr_prefetch_host(void* ptr, size_t numBytes)                                      (1)
```
illuhad, line 64: Reminder: Change to `const void*`

TApplencourt:
WG:
- Agree to go forward for the old format
- Need to resolve current comment (`const void`) and others

tdavidcl:
> For those who are asking for this KHR to move forward in order to speed up adoptions, which SYCL implementation(s) are you using?

Both in equal amounts in my case

> If this includes DPC++, then I wanted to explain my hesitancy. DPC++ currently provides a way to prefetch to host, but it has no effect when using the Level Zero backend because there is not yet support in Level Zero to do this prefetch. Therefore, I don't see a lot of value for DPC++ to implement a standalone "prefetch host" KHR in the short term because it would just be a no-op anyway on Level Zero. By the time we have Level Zero support, I expect that the WG will have approved #922, so I'd just as soon wait until then to have a prefetch-host KHR.

Do you know the status for HIP and CUDA too.

In any case in term of personal preferences I would rather have a no-op implementation for the time being with smaller features rather than larger ones. In general SYCL is quite large and intertwined so any efforts toward something with smaller components is valuable since I think that this is required for larger adoption (opinionated view on it 😄).

PR #962, opened 2026-02-18
Clarify host concurency access

### LINKED ISSUE #951: clarify concurrent access and aspect::usm_host_allocations
### Specification Version

SYCL 2020 (Revision 11)

### Section Number(s)

4.6.4.5. Aspects

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-aspects

### Issue Description

The current spec description for `aspect::usm_host_allocations` is very clear that it does not provide atomic access to host allocations, but less clear that it also does not support concurrent access to host allocations.  While this can be inferred from the description of `aspect::usm_atomic_host_allocations`, which explicitly mentions both concurrent access and atomic modification, it would be much clearer to explicitly say in the description of `aspect::usm_host_allocations` that concurrent access requires `aspect::usm_atomic_host_allocations`, similar to the text for `aspect::usm_shared_allocations`.

Specifically, I suggest changing this text:

> The device only supports atomic modification of a host allocation if aspect::usm_atomic_host_allocations is also supported.

to something like:

> Concurrent access and atomic modification of a host allocation is only supported if aspect::usm_atomic_host_allocations is also supported.

### Code Example (Optional)

_No response_

TApplencourt:
Fix #951

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3596,7 +3596,7 @@ Indicates that the device supports explicit USM allocations as described in
 Indicates that the device can access USM memory allocated via
 [code]#usm::alloc::host#.
 Concurrent access and atomic modification of a host allocation is only supported
-if [code]#aspect::usm_atomic_host_allocations# is also supported.
+if [api]#aspect::usm_atomic_host_allocations# is also supported.
 (See <<sec:usm>>.)
 --
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -3595,8 +3595,8 @@ Indicates that the device supports explicit USM allocations as described in
 --
 Indicates that the device can access USM memory allocated via
 [code]#usm::alloc::host#.
-The device only supports atomic modification of a host allocation if
-[api]#aspect::usm_atomic_host_allocations# is also supported.
+Concurrent access and atomic modification of a host allocation is only supported
+if [code]#aspect::usm_atomic_host_allocations# is also supported.
 (See <<sec:usm>>.)
 --
 


adoc/chapters/programming_interface.adoc
```diff
@@ -3595,8 +3595,8 @@ Indicates that the device supports explicit USM allocations as described in
 --
 Indicates that the device can access USM memory allocated via
 [code]#usm::alloc::host#.
-The device only supports atomic modification of a host allocation if
-[api]#aspect::usm_atomic_host_allocations# is also supported.
+Concurrent access and atomic modification of a host allocation is only supported
+if [code]#aspect::usm_atomic_host_allocations# is also supported.
```
```suggestion
if [api]#aspect::usm_atomic_host_allocations# is also supported.
```

This will style like `[code]` and also make it a hyperlink.  Note that the previous wording used `[api]` also.

adoc/chapters/programming_interface.adoc
gmlueck, line 3610: Since we are clarifying the USM host aspects ... Does `usm_atomic_host_allocations` also mean that atomic operations are allowed between two devices?  The current wording only allows atomics between the host and one device.

Should the wording be like this, similar to `aspect::usm_atomic_shared_allocations`:

> The host and other devices in the same context that also support this capability may concurrently access and atomically modify host allocations

Tagging @bashbaug since he submitted the issue that spawned this PR.

If this is *not* the intent, it might be work a clarifying note to emphasize that device-to-device atomics and concurrent accesses are not guaranteed by this aspect.

TApplencourt: We say:

>If a device does support concurrent access, both the host and and the device may atomically modify the same data inside an allocation. Allocations, or pieces of allocations, are now free to migrate to different devices in the same context that also support this capability. Additionally, many devices that support concurrent access may support a working set of shared allocations larger than device memory. Users may query whether a device supports concurrent access with atomic modification of shared allocations through the aspect aspect::usm_atomic_shared_allocations. See [Section 4.6.4.5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-aspects) for more details.

I'm a little afraid that if we put too much on this aspect, it will start to make it hard to implement. 

Level zero spec not helpfull:
- https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/api.html#ze-memory-access-cap-flags-t

Cuda have a lot of info in their memory model:
- https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/memory_model.html#atomicity

TApplencourt: Regardless should we merge this PR, and open another issue to discuss concurency? So we don't stall this one.  One clarification merged is better than 2 not merged :D

bashbaug: > Should the wording be like this [...]

Yes, I think so, but I agree this should be discussed in its own issue and fixed in its own PR.

The clarifications to aspect::usm_host_allocations in this PR LGTM - thanks!

TApplencourt: https://github.com/KhronosGroup/SYCL-Docs/issues/964 For you @bashbaug  : ) Thanks!

gmlueck (approved):

illuhad (approved):

TApplencourt:
\o/ Thanks for all the reviewers! 
- Clarification to SYCL2020
- No CTS change required
- Ready to be merged

PR #965, opened 2026-02-19
Remove clang-format

TApplencourt:
You will be missed clang-format : ( 

But what the WG want, we do!


TApplencourt:
(For futur work: All compiler have `-fsyntax-only`, if people are motivated it can be a path forward to verify code validify. But not sure how it will work with the `sycl.hpp` which is requried but lot of example but that we don't provide.)

PR #975, opened 2026-02-21
Fix old formatting mistake

gmlueck:
I noticed this while updating the copyright dates.

Fix an old formatting mistake with this synopsis.  This appears to have been introduced in #269 when clang format was run over all the synopsis headers.


gmlueck:
Merging as editorial

PR #986, opened 2026-03-11
Clarify "atomic_ref" with private address space

gmlueck:
The spec was not clear about this before.  The OpenCL SPIR-V environment specification says that atomic operations on objects in the private address space produce undefined behavior.  Therefore, it seems sensible to make the same clarification in SYCL.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -20648,6 +20648,11 @@ operations in device code with a syntax similar to the {cpp} standard
 [code]#std::atomic_ref#.
 The [code]#sycl::atomic_ref# class must not be used in host code.
 
+The referenced object of type [code]#T# must not reside in the private address
+space.
+If [code]#atomic_ref# is constructed with a reference to an object in the
+private address space, the behavior is undefined.
+
 Unlike [code]#std::atomic_ref#, [code]#sycl::atomic_ref# does not provide a
 default memory ordering for its operations.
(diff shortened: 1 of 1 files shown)


gmlueck:
The relevant text from the OpenCL SPIR-V Environment spec is here:

> For all Atomic Instructions:
>
> * Only 32-bit integer types are supported for the Result Type and/or type of Value.
> 
> * The Pointer operand must be a pointer to the Function, Workgroup, or CrossWorkgroup Storage Classes. Note that an Atomic Instruction on a pointer to the Function Storage Class is valid, but does not have defined behavior.
>
> * For OpenCL environments that support and declare the GenericPointer capability, the Pointer operand may be a pointer to the Generic Storage Class, however behavior is still undefined if the Generic pointer represents a pointer to the Function Storage Class.

(https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#validation-rules)

bader:
> The OpenCL SPIR-V environment specification says that atomic operations on objects in the address space produce undefined behavior.

the address space -> the **private** address space?

TApplencourt (approved):

VerenaBeckham (approved):

TApplencourt:
- Check, If in `CTS` we test this (I guess not, but just double check -- deadline 05/21). 
- Agree to merge as a clarification
- Cuda: May Working (in case people write a Cuda backend specification one day)

TApplencourt:
We can merge this one.

PR #987, opened 2026-03-11
Add const to prefetch and memadvice

### LINKED ISSUE #963: Prefetch/ Memadvice should be `const`
### Specification Version

SYCL 2020 (Revision 11)

### Section Number(s)

somehere

### Bug Description

```

  event prefetch(void* ptr, std::size_t numBytes);
  event prefetch(void* ptr, std::size_t numBytes, event depEvent);
  event prefetch(void* ptr, std::size_t numBytes,
                 const std::vector<event>& depEvents);

  /* -- Shortcut functions: mem_advise -- */

  event mem_advise(void* ptr, std::size_t numBytes, int advice);
  event mem_advise(void* ptr, std::size_t numBytes, int advice, event depEvent);
  event mem_advise(void* ptr, std::size_t numBytes, int advice,
                   const std::vector<event>& depEvents);
```

Adabtivecpp: already implemented as const
DPCP: Maybe?

- For sure will clarify into launch_queue khr
- Maybe clarification fix into SYCL 2020

TApplencourt:
Fix #963

Implementation are correct, spec should follow : )

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4627,11 +4627,11 @@ queue.
 .[apidef]#queue::prefetch#
 [source,role=synopsis,id=api:queue-prefetch]
 ----
-event prefetch(void* ptr, std::size_t numBytes)                                      (1)
+event prefetch(const void* ptr, std::size_t numBytes)                                      (1)
 
-event prefetch(void* ptr, std::size_t numBytes, event depEvent)                      (2)
+event prefetch(const void* ptr, std::size_t numBytes, event depEvent)                      (2)
 
-event prefetch(void* ptr, std::size_t numBytes, const std::vector<event>& depEvents) (3)
+event prefetch(const void* ptr, std::size_t numBytes, const std::vector<event>& depEvents) (3)
 ----
 
 _Effects (1):_ Equivalent to calling [api]#queue::submit# with a command group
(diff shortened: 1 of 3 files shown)


TApplencourt:
(I don't know what magic @illuhad have done, but I cannot add him as a reviewer... Smart guy!)

gmlueck (approved):
LGTM!

Do we need to change the CTS to pass a `const` pointer to these APIs?

TApplencourt:
>Do we need to change the CTS to pass a const pointer to these APIs?

Oh yes good idea; it should make the CST  fail with an implementation implementing the old specification. 

Not sure, if it's a blocking change but should be better indeed.

illuhad (approved):
LGTM!

@TApplencourt I think github only allows requesting reviews from people who have write permissions, which, for whatever reason, I don't have for SYCL-Docs. That's also why my review checkmarks are not green either.

gmlueck:
> I think github only allows requesting reviews from people who have write permissions, which, for whatever reason, I don't have for SYCL-Docs. That's also why my review checkmarks are not green either.

I think it is because you were not part of the [sycl-committee](https://github.com/orgs/KhronosGroup/teams/sycl-committee/members) group.  I just invited you, and it looks like you are a member now.

illuhad:
Yep, thanks @gmlueck :-)

TApplencourt:
WG: Merge as clarification.

TApplencourt:
CTS: Merged, can merge as clarification in SYCL2020

PR #988, opened 2026-03-13
Avoid some undefined attribute warnings

gmlueck:
I'd like to enable `attribute-missing=warn` in the Asciidoctor build in order to catch cases where we forgot to define an attribute.  However, doing so causes false warnings for things in `[code]` style that look like attributes.  Add escaping to these, which avoids the warning.

Unfortunately, I still cannot enable `attribute-missing=warn` because there are still the following false warnings:

```
asciidoctor: WARNING: skipping reference to missing attribute: n
asciidoctor: WARNING: skipping reference to missing attribute: objectbundle
asciidoctor: WARNING: skipping reference to missing attribute: objectbundle
```

I have no idea where these are coming from.  In fact, the word "objectbundle" doesn't even exist in our repo.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1675,22 +1675,23 @@ Some multi-dimensional objects can be viewed in a linear form.
 When this happens, the right-most term in the object's range varies fastest in
 the linearization.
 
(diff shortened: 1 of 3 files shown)


gmlueck:
Merging as editorial

PR #996, opened 2026-04-04
Fix kernel argument definition for named function objects

### LINKED ISSUE #915: Definition of "kernel argument" is wrong for named kernel object
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

[section 4.12.4.](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:kernel.parameter.passing) "Rules for parameter passing to kernels"

### Bug Description

During a conversation at the Philadelphia F2F, we decided that this statement in that section is not quite right:

> If the kernel is a named function object, the `operator()` member function (or other member functions that it calls) may reference member variables inside the same named function object. Any such member variables become parameters to the kernel.

This should be changed to say that all member variables of the named function object are kernel arguments.  If an implementation can prove that certain member variables are not used in the kernel, then it can avoid passing them.  This is essentially dead argument elimination.

During our conversation, we noted that the compiler cannot always prove whether a member variable of the kernel object is used in device code.  Consider the case when the `operator()` calls some `SYCL_EXTERNAL` member function that is defined in another translation unit.  In such a case, the compiler cannot know which member variables are used in the `SYCL_EXTERNAL` function.

ssam18:
Fixes #915.

The previous wording implied that only member variables referenced by `operator()` become kernel parameters, which breaks down when `operator()` calls a `SYCL_EXTERNAL` function defined in another translation unit — the compiler has no way to know which members are used there. The corrected text states that all member variables of the named function object are kernel parameters, and that an implementation may drop any it can prove are unused (dead argument elimination). This also aligns with how DPC++ already handles type checking for kernel objects.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17667,10 +17667,10 @@ a device copyable type in that case and the specialization is ignored.
 
 A SYCL application passes parameters to a kernel in different ways depending on
 whether the kernel is a named function object or a lambda expression.
-If the kernel is a named function object, the [code]#operator()# member function
-(or other member functions that it calls) may reference member variables inside
-the same named function object.
-Any such member variables become parameters to the kernel.
+If the kernel is a named function object, all member variables of the named
+function object become parameters to the kernel.
+An implementation may omit passing a member variable that it can prove is not
+used by the kernel, which is a form of dead argument elimination.
 If the kernel is a lambda expression, any variables captured by the lambda
 become parameters to the kernel.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17667,10 +17667,8 @@ a device copyable type in that case and the specialization is ignored.
 
 A SYCL application passes parameters to a kernel in different ways depending on
 whether the kernel is a named function object or a lambda expression.
-If the kernel is a named function object, all member variables of the named
+If the kernel is a named function object, all non-static member variables of the named
 function object become parameters to the kernel.
-An implementation may omit passing a member variable that it can prove is not
-used by the kernel, which is a form of dead argument elimination.
 If the kernel is a lambda expression, any variables captured by the lambda
 become parameters to the kernel.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -17667,8 +17667,8 @@ a device copyable type in that case and the specialization is ignored.
 
 A SYCL application passes parameters to a kernel in different ways depending on
 whether the kernel is a named function object or a lambda expression.
-If the kernel is a named function object, all non-static member variables of the named
-function object become parameters to the kernel.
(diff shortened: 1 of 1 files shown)


ssam18:
I have signed the Contribution License Agreement, Still CI/CD workflow is pending on that. When I click, it shows you already have signed. Not sure how to fix it.

TApplencourt:
@gmlueck said "I volunteered to draft a PR for this." Why are you not consulting with him before writing a PR? It's kind of rude... We are all human with feeling.

adoc/chapters/programming_interface.adoc
```diff
@@ -17667,10 +17667,10 @@ a device copyable type in that case and the specialization is ignored.
...
 A SYCL application passes parameters to a kernel in different ways depending on
 whether the kernel is a named function object or a lambda expression.
-If the kernel is a named function object, the [code]#operator()# member function
-(or other member functions that it calls) may reference member variables inside
-the same named function object.
-Any such member variables become parameters to the kernel.
+If the kernel is a named function object, all member variables of the named
+function object become parameters to the kernel.
+An implementation may omit passing a member variable that it can prove is not
+used by the kernel, which is a form of dead argument elimination.
```
gmlueck: I don't think this second sentence should be added to the spec.  It is not the responsibility of the spec to point out possible optimizations in the implementation.

adoc/chapters/programming_interface.adoc
```diff
@@ -17667,10 +17667,10 @@ a device copyable type in that case and the specialization is ignored.
 
 A SYCL application passes parameters to a kernel in different ways depending on
 whether the kernel is a named function object or a lambda expression.
-If the kernel is a named function object, the [code]#operator()# member function
-(or other member functions that it calls) may reference member variables inside
-the same named function object.
-Any such member variables become parameters to the kernel.
+If the kernel is a named function object, all member variables of the named
```
```suggestion
If the kernel is a named function object, all non-static member variables of the named
```

gmlueck:
> @gmlueck said "I volunteered to draft a PR for this." Why are you not consulting with him before writing a PR? It's kind of rude... We are all human with feeling.

No worries.  Glad for the help!

ssam18:
> @gmlueck said "I volunteered to draft a PR for this." Why are you not consulting with him before writing a PR? It's kind of rude... We are all human with feeling.

My intension is to not hurt your feeling. I am here to help and expedite the work for this group. You can collaborate here in this PR and take the credit, I don't mind.

TApplencourt:
- Just need to reflow, and agree to merge.

TApplencourt (approved):

TApplencourt:
Maybe add a note to said that user can always by pass by specialization the trait to true? (to behave like cuda, jsut opt-out of the diagnostic)

gmlueck (approved):
Please fix reflow failure.

illuhad (approved):

ssam18:
> Please fix reflow failure.

Done.

TApplencourt:
WG: Will merge Clarification 2020

Thanks a lot! Sorry for the delay.

PR #998, opened 2026-04-04
Add converting constructors to multi_ptr for generic address space

### LINKED ISSUE #894: Should `multi_ptr` with an address space be implicitly convertible to a generic pointer?
### Specification Version

SYCL 2020 (Revision 10)

### Section Number(s)

- Section 4.7.7.1. Multi-pointer class, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:multiptr

### Issue Description

The specialization of `multi_ptr` for the generic address space provides two additional overloads of `operator=` that allow assignment from a `multi_ptr` with any address space:

```c++
  // Available only when:
  //   (Space == access::address_space::generic_space &&
  //    AS != access::address_space::constant_space)
  template <access::address_space AS, access::decorated IsDecorated>
  multi_ptr& operator=(const multi_ptr<value_type, AS, IsDecorated>&);

  // Available only when:
  //   (Space == access::address_space::generic_space &&
  //    AS != access::address_space::constant_space)
  template <access::address_space AS, access::decorated IsDecorated>
  multi_ptr& operator=(multi_ptr<value_type, AS, IsDecorated>&&);
```

The same specialization does does not provide any implicit constructors for this case, and the only way to construct a `multi_ptr` for the generic address space is to do so explicitly:

```c++
  // Constructors
  multi_ptr();
  multi_ptr(const multi_ptr&);
  multi_ptr(multi_ptr&&);
  explicit multi_ptr(
      typename multi_ptr<ElementType, Space, access::decorated::yes>::pointer);
  multi_ptr(std::nullptr_t);
```

This inconsistency may lead to some confusion, and it makes generic `multi_ptr` harder to use.  Is it intentional?

### Code Example (Optional)

```c++
#include <sycl/sycl.hpp>

void foo(sycl::multi_ptr<float, sycl::access::address_space::generic_space> GenericPointer) {
  /* some function that doesn't care about the address space */
}

void bar(sycl::multi_ptr<float, sycl::access::address_space::global_space> GlobalPointer)
{

    // This fails to compile, because there's no implicit conversion defined.
    {
        foo(GlobalPointer);
    }

    // This also fails, because the = here requires a converting constructor.
    {
        sycl::multi_ptr<float, sycl::access::address_space::generic_space> GenericPointer = GlobalPointer;
        foo(GenericPointer);
    }

    // This is (surprisingly) fine, because the = here is an overloaded assignment.
    {
        sycl::multi_ptr<float, sycl::access::address_space::generic_space> GenericPointer;
        GenericPointer = GlobalPointer;
        foo(GenericPointer);
    }
}
```

ssam18:
The generic_space specialization already had converting copy and move `operator=` overloads accepting any non-constant address space, but had no matching constructors. This meant assigning to an already-constructed `generic_ptr` worked, while direct initialization and passing as a function argument did not. The fix adds the two missing constructors with the same constraints as the existing assignment operators, making the interface consistent. Fixes #894.

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -10387,6 +10387,28 @@ multi_ptr(std::nullptr_t)
 a@
 [source]
 ----
+template <access::address_space AS, access::decorated IsDecorated>
+multi_ptr(const multi_ptr<value_type, AS, IsDecorated>&)
+----
+   a@ Available only when:
+      [code]#(Space == access::address_space::generic_space && AS != access::address_space::constant_space)#.
+
+Constructs a [code]#generic_ptr# by copying the value of the right hand side [code]#multi_ptr#.
+
+a@
+[source]
+----
+template <access::address_space AS, access::decorated IsDecorated>
+multi_ptr(multi_ptr<value_type, AS, IsDecorated>&&)
+----
+   a@ Available only when:
+      [code]#(Space == access::address_space::generic_space && AS != access::address_space::constant_space)#.
+
+Constructs a [code]#generic_ptr# by moving the value of the right hand side [code]#multi_ptr#.
+
+a@
+[source]
+----
 template <typename AccDataT, int Dimensions,
           access_mode Mode,
           access::placeholder IsPlaceholder>
--- a/adoc/headers/multipointer.h
+++ b/adoc/headers/multipointer.h
@@ -65,6 +65,18 @@ class multi_ptr {
   multi_ptr(std::nullptr_t);
 
   // Available only when:
+  //   (Space == access::address_space::generic_space &&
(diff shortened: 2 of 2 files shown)


TApplencourt:
The issue said:

> This inconsistency may lead to some confusion, and it makes generic multi_ptr harder to use. Is it intentional?

Do you think it's intentional? We know how to fix it, it was not the problem...

ssam18:
@TApplencourt 
The inconsistency looks unintentional as operator= already implicitly accepts any non-constant address space for generic_space, so it is reasonable to expect the constructors to do the same. Having assignment work but not construction or implicit conversion when passing as a function argument feels like an oversight. 
That said, if there is  context I am missing that made this an intentional omission, I am happy to revisit.

PR #1005, opened 2026-04-28
Clarify Unified Addressing

VerenaBeckham:
Unified Addressing should only be within a single context (and the host).

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11174,12 +11174,12 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
 
 === Unified addressing
 
-Unified Addressing guarantees that all devices in a given SYCL <<context>>
-will use a unified address space.
+Unified Addressing guarantees that all devices in a given SYCL <<context>> will
+use a unified address space.
 Pointer values in the unified address space will always refer to the same
 location in memory.
-The unified address space encompasses the host and all devices in a given
-SYCL <<context>>.
+The unified address space encompasses the host and all devices in a given SYCL
+<<context>>.
 Note that this does not require addresses in the unified address space to be
 accessible on all devices, just that pointer values will be consistent.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11174,10 +11174,12 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
 
 === Unified addressing
 
-Unified Addressing guarantees that all devices will use a unified address space.
+Unified Addressing guarantees that all devices in a given SYCL <<context>>
+will use a unified address space.
 Pointer values in the unified address space will always refer to the same
 location in memory.
-The unified address space encompasses the host and one or more devices.
+The unified address space encompasses the host and all devices in a given
+SYCL <<context>>.
 Note that this does not require addresses in the unified address space to be
 accessible on all devices, just that pointer values will be consistent.
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11179,16 +11179,16 @@ device code, providing the following guarantees for pointer values to objects
 that have overlapping lifetimes:
 
   * A USM pointer is different from all non-USM memory addresses on the host.
-  * A USM host memory pointer allocated from context C has a value that is
+  * A USM host memory pointer allocated from context _C_ has a value that is
     different from all other USM host memory pointers (regardless of context),
     different from all USM shared memory pointers (regardless of context), and
-    different from all USM device memory pointers allocated from context C.
-  * A USM shared memory pointer allocated from context C has a value that is
+    different from all USM device memory pointers allocated from context _C_.
+  * A USM shared memory pointer allocated from context _C_ has a value that is
     different from all other USM shared memory pointers (regardless of context),
     different from all USM host memory pointers (regardless of context), and
-    different from all USM device memory pointers allocated from context C.
-  * A USM device memory pointer allocated from context C has a value that is
-    different from all other USM pointers allocated from context C.
+    different from all USM device memory pointers allocated from context _C_.
+  * A USM device memory pointer allocated from context _C_ has a value that is
+    different from all other USM pointers allocated from context _C_.
 
 === Kinds of unified shared memory
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11174,9 +11174,8 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
 
 === Unified addressing
 
-USM allocations have pointer values that are unique across both host code and
-device code, providing the following guarantees for pointer values to objects
-that have overlapping lifetimes:
+The following guarantees apply to the pointer values of USM allocations for
+objects with overlapping lifetimes:
 
   * A USM pointer is different from all non-USM memory addresses on the host.
   * A USM host memory pointer allocated from context _C_ has a value that is
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -11174,15 +11174,21 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
 
 === Unified addressing
 
-Unified Addressing guarantees that all devices in a given SYCL <<context>> will
-use a unified address space.
-Pointer values in the unified address space will always refer to the same
-location in memory.
-The unified address space encompasses the host and all devices in a given SYCL
-<<context>>.
-Note that this does not require addresses in the unified address space to be
-accessible on all devices, just that pointer values will be consistent.
+USM allocations have pointer values that are unique across both host code and
+device code, providing the following guarantees for pointer values to objects
+that have overlapping lifetimes:
 
+  * A USM pointer is different from all non-USM memory addresses on the host.
+  * A USM host memory pointer allocated from context C has a value that is
+    different from all other USM host memory pointers (regardless of context),
+    different from all USM shared memory pointers (regardless of context), and
+    different from all USM device memory pointers allocated from context C.
+  * A USM shared memory pointer allocated from context C has a value that is
+    different from all other USM shared memory pointers (regardless of context),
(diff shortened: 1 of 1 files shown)


adoc/chapters/programming_interface.adoc
```diff
@@ -11174,10 +11174,12 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
...
-Unified Addressing guarantees that all devices will use a unified address space.
+Unified Addressing guarantees that all devices in a given SYCL <<context>> will
+use a unified address space.
 Pointer values in the unified address space will always refer to the same
 location in memory.
-The unified address space encompasses the host and one or more devices.
+The unified address space encompasses the host and all devices in a given SYCL
+<<context>>.
 Note that this does not require addresses in the unified address space to be
 accessible on all devices, just that pointer values will be consistent.
```
gmlueck: I've always found this section to be unclear, so maybe we can take this as an opportunity to clarify it.  Looking closely at three of the four sentences in this section:

* The first sentence defines "Unified Addressing" as "use a unified address space".  That seems like an unhelpful circular definition.
* The second sentence says that "pointer values [...] refer to the same location in memory".  It's not clear to me what this means.  Isn't this always the case for pointer values?  The value of a pointer is the address of a memory location.  How could a pointer refer to a different location in memory if its value doesn't change?
* The fourth sentence says that "pointer values will be consistent".  What does "consistent" mean here?

Are we trying to say the following?

* The address of any USM buffer is different from all non-USM memory addresses on the host.
* The address of any USM host memory buffer allocated from context _C_ has an address that is different from all other USM host memory buffers (regardless of context), different from all USM shared memory buffers (regardless of context), and different from all USM device memory buffers allocated from context _C_.
* The address of any USM shared memory buffer allocated from context _C_ has an address that is different from all other USM shared memory buffers (regardless of context), different from all USM host memory buffers (regardless of context), and different from all USM device memory buffers allocated from context _C_.
* The address of any USM device memory buffer allocated from context _C_ has an address that is different from all other USM device memory buffers allocated from context _C_, different from all USM host memory buffers (regardless of context), and different from all USM shared memory buffers (regardless of context).

Tagging @bashbaug because the SYCL definition of "unified addressing" should match OpenCL.

hdelan:
Perhaps [this section](https://eel.is/c++draft/basic.memobj#intro.object-10) of the C++ spec could be extended to USM buffers within SYCL contexts, but not across different contexts.

gmlueck:
> Perhaps [this section](https://eel.is/c++draft/basic.memobj#intro.object-10) of the C++ spec could be extended to USM buffers within SYCL contexts, but not across different contexts.

I think that section of the C++ spec is mostly talking about nested objects.  How do you think it could be extended to clarify our situation?

FWIW, I considered writing my proposed text in terms of C++ object addresses, but I changed it to refer to USM buffer addresses instead.  I did this because the memory returned from the USM allocation routines is raw memory that does not contain any C++ objects.

hdelan:
It says that except in the mentioned cases

> [two objects] have distinct addresses and occupy disjoint bytes of storage

Maybe a better place to extend is [here](https://eel.is/c++draft/expr.eq#3) but again this refers to objects rather than memory as you say.

adoc/chapters/programming_interface.adoc
VerenaBeckham: I think I would agree with these statements.
However, I'm not sure I like the term "USM buffer". Since a "buffer" is overloaded in SYCL-speak (buffer/accessors) this seems confusing and I couldn't find any other reference to the term in the spec. I think if we need this term we should define it in this section first, but at the end of the day USM gives you pointers, so can we reword it in terms of pointers without losing clarity? Perhaps we could spell it out in terms of "pointer returned by a SYCL malloc_X function", if that's what we mean?

gmlueck: > Perhaps we could spell it out in terms of "pointer returned by a SYCL malloc_X function", if that's what we mean?

Yes, that's what I mean.  If we're being pedantic, you can also get a USM pointer from the `usm_allocator`.  Perhaps we need to define a term "USM pointer" to mean a pointer returned from any of the USM allocation functions.

gmlueck: How about this for a proposed rewrite of the "Unified addressing" section:

> USM allocations have pointer values that are unique across both host code and device code, providing the following guarantees for pointer values to objects that have overlapping lifetimes:
>
> * A USM pointer is different from all non-USM memory addresses on the host.
> * A USM host memory pointer allocated from context _C_ has a value that is different from all other USM host memory pointers (regardless of context), different from all USM shared memory pointers (regardless of context), and different from all USM device memory pointers allocated from context _C_.
> * A USM shared memory pointer allocated from context _C_ has a value that is different from all other USM shared memory pointers (regardless of context), different from all USM host memory pointers (regardless of context), and different from all USM device memory pointers allocated from context _C_.
> * A USM device memory pointer allocated from context _C_ has a value that is different from all other USM device memory pointers allocated from context _C_, different from all USM host memory pointers (regardless of context), and different from all USM shared memory pointers (regardless of context).

TApplencourt:
- What is a USM Pointer?
 "An USM pointer, is a pointer who is returned by sycl::malloc_* like"

- Can do that in another PR for "what is allocate USM Pointer"?
  -  Thomas will open an Issue

- Verena will update this PR with greg test

adoc/chapters/programming_interface.adoc
bashbaug: ~~This mostly looks good to me, but I believe a USM host pointer needs to be different from all USM device pointers regardless of context.  Same goes for USM shared pointers, which also need to be different from all USM device pointers regardless of context.  Basically, the only USM pointers that are allowed to alias are USM device pointers, and only for USM device pointers from different contexts.~~

~~Note: I don't think we can go so far as to *guarantee* this, because a USM allocation from a buggy implementation could alias a USM host or USM shared allocation from a good implementation, but if this happens it's a bug in the other implementation.~~

(The above text is not quite correct.  It is probably correct for at least some implementations, but there are no guarantees.  See the [reply below](https://github.com/KhronosGroup/SYCL-Docs/pull/1005#discussion_r3269604253) for details.)

Another way of thinking about this is: For any given context and any given pointer, it is possible to determine with 100% certainty whether the pointer is a USM pointer allocated against the context, and if it is, what type of USM does the pointer point to.  If this were not the case, it would be impossible to implement `get_pointer_type`.

hdelan: AFAIU we would only require USM host pointers to never alias any USM device pointers if we want a USM host pointer to be able to memcpy to and from a device pointer in another context. Do you think we need to require cross-context memcpy in general? Let me know if I am missing something.

gmlueck: @bashbaug and I reviewed my proposed wording above, and we think it is not quite right.  Here is what it should be:

> USM allocations have pointer values that are unique across both host code and device code, providing the following guarantees for pointer values to objects that have overlapping lifetimes:
>
> * A USM pointer is different from all non-USM memory addresses on the host.
> * A USM host memory pointer allocated from context _C_ has a value that is different from all other USM host memory pointers (regardless of context), different from all USM shared memory pointers (regardless of context), and different from all USM device memory pointers allocated from context _C_.
> * A USM shared memory pointer allocated from context _C_ has a value that is different from all other USM shared memory pointers (regardless of context), different from all USM host memory pointers (regardless of context), and different from all USM device memory pointers allocated from context _C_.
> * A USM device memory pointer allocated from context _C_ has a value that is different from all other USM pointers allocated from context _C_.

TApplencourt: Verena agree to update the PR with this text

VerenaBeckham: Updated.

adoc/chapters/programming_interface.adoc
```diff
@@ -11174,13 +11174,21 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
...
-The unified address space encompasses the host and one or more devices.
-Note that this does not require addresses in the unified address space to be
-accessible on all devices, just that pointer values will be consistent.
-
+USM allocations have pointer values that are unique across both host code and
+device code, providing the following guarantees for pointer values to objects
+that have overlapping lifetimes:
+
+  * A USM pointer is different from all non-USM memory addresses on the host.
+  * A USM host memory pointer allocated from context C has a value that is
```
```suggestion
  * A USM host memory pointer allocated from context _C_ has a value that is
```

Kind of a nit, but we usually use italics in cases like this where a letter is used to represent a specific instance of an object.  Same for the other C's below.

gmlueck (commented):
Looks good.  Just a small formatting nit below.

adoc/chapters/programming_interface.adoc
VerenaBeckham: Done.

adoc/chapters/programming_interface.adoc
```diff
@@ -11174,13 +11174,21 @@ include::{code_dir}/usm_device.cpp[lines=4..-1]
...
 === Unified addressing
 
-Unified Addressing guarantees that all devices will use a unified address space.
-Pointer values in the unified address space will always refer to the same
-location in memory.
-The unified address space encompasses the host and one or more devices.
-Note that this does not require addresses in the unified address space to be
-accessible on all devices, just that pointer values will be consistent.
-
+USM allocations have pointer values that are unique across both host code and
```
VerenaBeckham: I'm still not quite sure about the meaning of the first half of this first sentence. Perhaps we can discuss in the call.

gmlueck: I meant it only as an introduction to the bullet points below.  For example, the first bullet says that a USM pointer is different from a the address of any object on the host --- therefore that USM pointer is unique across both host and device.  The other bullets provide other uniqueness guarantees.

If you think that part of the sentence adds confusion, I'm open to rewording it.

TApplencourt: "Pointer values to objects USM that have overlapping lifetimes have the following qarenty:"  ~

VerenaBeckham: I have had a go at rewording.

gmlueck (approved):

TApplencourt (approved):

TApplencourt:
WG: The WG agree to merge this PR at a clarification of 2020.

PR #1015, opened 2026-06-10
Refactor anatomy.cpp for USM and in_order queue

MikeL270:
1. Updated `anatomy.cpp` to use `malloc_device`. We did not use `malloc_shared` in order to demonstrate the in_order behavior.
2. Modified accompanying description to reference the updated code and clarify that the command group scope is created implicitly because of the `parallel_for` queue shortcut.  

Thank you for taking the time to review my pr and do not hesitate to reach out for clarification if needed.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -79,7 +79,7 @@ the SYCL features that will be used.
 A SYCL application runs on a <<sec:platformmodel, SYCL Platform>>.
 The application is structured in three scopes which specify the different
 sections; <<application-scope>>, <<command-group-scope>> and <<kernel-scope>>.
-The <<kernel-scope>> specifies a single kernel function that will be, or has
+The <<kernel-scope>> specifies a <<sycl-kernel-function>> that will be, or has
 been, compiled by a <<device-compiler>> and executed on a <<device>>.
 In this example <<kernel-scope>> is defined by lines 18 to 19, the body of the
 lambda passed to [code]#parallel_for#.
@@ -87,7 +87,8 @@ The <<command-group-scope>> specifies a unit of work which is comprised of a
 <<sycl-kernel-function>> and any associated requirements.
 In this example the kernel is launched using the [code]#parallel_for# queue
 shortcut on lines 17 to 20, which implicitly defines the <<command-group-scope>>
-around the kernel invocation.
+around the <<kernel-scope>>.
+All queue shortcuts can be found in <<sec:queue-class>>.
 The <<application-scope>> specifies all other code outside of a
 <<command-group-scope>>.
 These three scopes are used to control the application flow and the construction
@@ -115,21 +116,17 @@ In the case of [code]#parallel_for# the <<sycl-kernel-function>> will be
 executed over the given range from 0 to 1023.
 The different member functions to execute kernels can be found in
 <<subsec:invokingkernels>>.
-
-In this example the [code]#parallel_for# queue shortcut on lines 17 to 20 is
-used to implicitly define a command group containing a single
-<<sycl-kernel-function>>. All queue shortcuts can be found in <<sec:queue-class>>.
-
 All the requirements for a kernel to execute are defined in this
 <<command-group-scope>>, as described in <<sec:executionmodel>>.
-In this case [code]#myQueue# is constructed on line 11 with the
-[code]#property::queue::in_order# property, which causes commands submitted to
-the queue to execute in the order they are enqueued.
-No <<device>> is specified, so the queue selects the best underlying device to
-execute on, leaving the decision up to the runtime.
+Additionally, in this example, [code]#myQueue# is constructed on line 11 with
+the [code]#property::queue::in_order# property, which causes commands submitted
+to the queue to execute in the order they were enqueued.
+No <<device>> is specified in the queue constructor, so the constructor selects
+the best underlying device to execute on, leaving the decision up to the
+runtime.
 
 In SYCL, data that is required within a <<sycl-kernel-function>> must be
-contained within a <<usm>>, <<image>>, or <<buffer>> allocation, as described in
+contained within a <<usm>> allocation, <<image>>, or <<buffer>> as described in
 <<sec:memory.model>>.
 In this example we use a USM device allocation, created with
 [code]#malloc_device# on line 14, which returns a pointer to memory accessible
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -81,10 +81,13 @@ The application is structured in three scopes which specify the different
 sections; <<application-scope>>, <<command-group-scope>> and <<kernel-scope>>.
 The <<kernel-scope>> specifies a single kernel function that will be, or has
 been, compiled by a <<device-compiler>> and executed on a <<device>>.
-In this example <<kernel-scope>> is defined by lines 25 to 26.
+In this example <<kernel-scope>> is defined by lines 18 to 19, the body of the
+lambda passed to [code]#parallel_for#.
 The <<command-group-scope>> specifies a unit of work which is comprised of a
-<<sycl-kernel-function>> and <<accessor,accessors>>.
-In this example <<command-group-scope>> is defined by lines 20 to 28.
+<<sycl-kernel-function>> and any associated requirements.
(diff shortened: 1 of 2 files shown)


gmlueck:
Please retarget this PR to the "main" branch.  Our policy is to make all changes on the "main" branch unless the change is somehow not applicable to future versions of the spec.  In this case, the change clearly is applicable to future versions.

Once the PR is merged, I will cherry-pick it to the "sycl-2020" branch if the WG approves it as a SYCL 2020 clarification (which I expect would be the case for this PR).

adoc/chapters/architecture.adoc
gmlueck: This is actually line 2, though I realize the same problem existed before your PR.

adoc/chapters/architecture.adoc
```diff
@@ -79,12 +79,16 @@ the SYCL features that will be used.
 A SYCL application runs on a <<sec:platformmodel, SYCL Platform>>.
 The application is structured in three scopes which specify the different
 sections; <<application-scope>>, <<command-group-scope>> and <<kernel-scope>>.
```
gmlueck: I think it is confusing to talk about command group scope here because this example doesn't have that scope.  If the purpose of this example is to illustrate a simple example, I think we should remove all the wording about the three scopes.  Personally, I think it would be better to rewrite this section describing the example program entirely.  The key points to make are:

* SYCL applications have both host code and device code.  Explain that the body of the lambda passed to `parallel_for` is the device code, which is called a kernel.
* The paragraph about creating an instance of a kernel which is executed over a range is good, so we should keep that.
* Since this application uses USM device memory, we should explain that this memory is not accessible on the host, so it needs to be copied back after the kernel completes.

I do not think the purpose of this example is to illustrate all SYCL features.  Therefore, I'd remove all references to features that are not illustrated by this example.  Therefore, remove references to naming a kernel via a template parameter, references to buffers or images, references to "requirements", references to the platform, etc.

MikeL270:
> Please retarget this PR to the "main" branch. Our policy is to make all changes on the "main" branch unless the change is somehow not applicable to future versions of the spec. In this case, the change clearly is applicable to future versions.
> 
> Once the PR is merged, I will cherry-pick it to the "sycl-2020" branch if the WG approves it as a SYCL 2020 clarification (which I expect would be the case for this PR).

Done! I will work on addressing the rest of your feedback shortly. Thank you for the guidance!

adoc/code/anatomy.cpp
```diff
@@ -3,37 +3,36 @@
...
-
-      // Enqueue a parallel_for task with 1024 work-items
-      cgh.parallel_for(1024, [=](id<1> idx) {
-        // Initialize each buffer element with its own rank number starting at 0
-        writeResult[idx] = idx;
-      });  // End of the kernel function
-    });    // End of our commands for this queue
-  }  // End of scope, so we wait for work producing resultBuf to complete
+  // Allocate host memory to store the results
+  std::vector<int> dataHost(1024);
```
EwanC: This is a nitpick, but rather than repeating the literal `1024`  5 times in the code, I think it'd be better to define a variable and reuse that, e.g.  `const int elements = 1024`

MikeL270: I agree with your feedback. I implemented this with `size_t` in 47feb8a32bba6878ae903eaa33574f8827022ff3.

I did this to remain consistent with other examples I have seen.

adoc/chapters/architecture.adoc
```diff
@@ -72,70 +72,51 @@ run in parallel on any device available.
...
+At line 14 we use [code]#malloc_device# to allocate memory on the <<device>>
+associated with our queue and store a universal shared memory <<usm>> pointer to
+the memory in [code]#dataDevice#.
+This memory is only accessible on the device.
+
+In lines 17 to 20, [code]#parallel_for# does three conceptual things:
+
+  1. The <<sycl-kernel-function>> (defined here as a lambda), passed as the
+    second argument to [code]#parallel_for#, is compiled by a
+    <<device-compiler>> into a <<kernel>> that can be ran on the <<device>>
```
gmlueck:
```suggestion
    <<device-compiler>> into a <<kernel>> that can be run on the <<device>>
```

adoc/chapters/architecture.adoc
```diff
@@ -72,70 +72,51 @@ run in parallel on any device available.
...
-on line 21.
-The <<buffer>> is used to keep track of access to the data and the <<accessor>>
-is used to request access to the data on a queue, as well as to track the
-dependencies between <<sycl-kernel-function>>.
-In this example the <<accessor>> is used to write to the data buffer on line 26.
-
+At line 2 we [code]#{hash}include# the SYCL header files, which provide all of
+of the SYCL features that will be used.
+
+At line 11 we instantiate a <<queue>> with the [code]#in_order# property.
```
gmlueck: Check all these line numbers, I think there are mostly wrong.  For example, the queue is constructed on line 14.

adoc/chapters/architecture.adoc
```diff
@@ -72,70 +72,51 @@ run in parallel on any device available.
...
+A queue is bound to a <<device>>, if you do not specify a device with a selector
+function, the SYCL runtime will choose one automatically.
+A queue will execute commands submitted to it on it's associated device.
+In our case commands will be executed in a first in first out (FIFO) fashion
+because of the [code]#in_order# property passed to our queue's constructor.
+The commands submitted to [code]#myQueue# are executed asynchronously from the
+host code.
+
+At line 14 we use [code]#malloc_device# to allocate memory on the <<device>>
+associated with our queue and store a universal shared memory <<usm>> pointer to
```
gmlueck:
```suggestion
associated with our queue and store a unified shared memory (<<usm>>) pointer to
```

adoc/chapters/architecture.adoc
```diff
@@ -72,70 +72,51 @@ run in parallel on any device available.
...
+  1. The <<sycl-kernel-function>> (defined here as a lambda), passed as the
+    second argument to [code]#parallel_for#, is compiled by a
+    <<device-compiler>> into a <<kernel>> that can be ran on the <<device>>
+    associated with [code]#myQueue#.
+  2. A command that invokes the <<kernel>> on the <<device>> is submitted to
+    [code]#myQueue#, and the <<kernel>> is executed asynchronously from the
+    host.
+  3. The command decomposes the range 0 to 1023 into <<work-item, work-items>>.
+    For each <<work-item>> the <<sycl-kernel-function>> is invoked once on the
+    device.
```
gmlueck:
```suggestion
    device, and these invocations may be executed in parallel.
```

This seems worth noting since it is key reason that offload compute is faster than just doing the computation on the host.

gmlueck (commented):
Thanks, this looks much improved!  Just a few comments below.

adoc/chapters/architecture.adoc
MikeL270: Fixed in fec597b5800dad9c12c54b3457e6409d91bf1afd.

gmlueck (approved):

adoc/code/anatomy.cpp
```diff
@@ -3,37 +3,39 @@
...
-    });    // End of our commands for this queue
-  }  // End of scope, so we wait for work producing resultBuf to complete
+  // Declare number of work items
+  constexpr size_t N = 1024;
+
+  // Allocate host memory to store the results
+  std::vector<int> dataHost(N);
+
+  // Create an in order queue to enqueue work to the default device
+  queue myQueue{property::queue::in_order()};
```
illuhad: This issue was already in the original code, but I think we should avoid `using namespace sycl` especially when also using STL containers like `vector`. As soon as something pulls in `std::queue`, we might get a clash with the SYCL queue in ADL scenarios.

TApplencourt: Agree. And for `hello-world`, It's specially important to show what is part of `sycl` and what is not.

MikeL270: Fixed in 5dec0eabe33775d4666f297d796a77cb79cb8547

illuhad (approved):

TApplencourt:
2 Approval, we can merge. As Clarificition of SYCL 2020. 
Thanks!

PR #1023, opened 2026-06-25
Clarify "get_wait_list" for in-order queue

### LINKED ISSUE #1017: Clarify "event::get_wait_list" for in-order queues
### Specification Version

SYCL 2020 (Revision 11)

### Section Number(s)

[Section 4.6.6. "Event class"](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:event-class)

### Issue Description

What do we expect `event::get_wait_list` to return for an event the comes from a command on an in-order queue?  Consider the following code snippet:

```
queue q{property::queue::in_order{});
event e1 = q.single_task([=]{});
event e2 = q.single_task([=]{});
std::vector<event> evts = e2.get_wait_list();
```

Do we expect `evts` to contain `e1`?  The description of `get_wait_list` says:

> Only direct dependencies are returned

However, it's not clear what a "direct dependency" is.  There is no explicit event dependency created here.  Instead, the dependency is implicit because the queue is in-order.

Once we finish the KHR proposed in #922, users will be able to write code like this:

```
queue q{property::queue::in_order{});
khr::launch_task(q, [=]{});
event e2 = q.single_task([=]{});
std::vector<event> evts = e2.get_wait_list();
```

Now, it's even less clear.  Do we expect the implementation to synthesize an event that corresponds to the first call to `launch_task` and then return that event in `evts`?  This would be very hard to do efficiently because the purpose of `launch_task` is to allow the implementation to optimize the submit path by *not* creating an event.

In fact, both cases above are hard for DPC++ to do efficiently because of the way we optimize event management for in-order queues.  Even this case would require some extra work in our implementation:
```
queue q{property::queue::in_order{});
event e1 = q.single_task([=]{});
event e2 = q.single_task(e1, [=]{});   // Redundantly list "e1" as a dependency even though queue is in-order
std::vector<event> evts = e2.get_wait_list();
```

Would we expect `evts` to contain `e1` in this case?  If so, we will need to do some extra bookkeeping just to return `e1` in this silly case.

From our perspective, it would be ideal to define `get_wait_list` such that it can return an empty list in all three of the cases I list above.

It's not clear to me what use case is solved by having `get_wait_list` at all.  Do we know if anyone is using it?  If so, what are they using it for?

### Code Example (Optional)

_No response_

gmlueck:
The working group does not know of any use case for `event::get_wait_list`, and the semantics seem unclear when the queue is in-order.  Therefore, we decided to deprecate this API and leave the behavior up to the implementation.

Closes #1017

### DIFF
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4888,12 +4888,17 @@ _Returns:_ The <<backend>> associated with this event.
 std::vector<event> get_wait_list()
 ----
 
+Deprecated by SYCL 2020.
+
 _Returns:_ The list of events that this event waits for in the dependence graph.
 Only direct dependencies are returned, and not transitive dependencies that
 direct dependencies wait on.
 
 _Remarks:_ Whether already completed events are included in the returned list is
 implementation-defined.
+If the event is associated with a command submitted to an in-order queue, then
+it is implementation-defined whether this function returns any dependent events
+or if it returns an empty vector.
 
 '''
 
--- a/adoc/chapters/programming_interface.adoc
+++ b/adoc/chapters/programming_interface.adoc
@@ -4888,7 +4888,7 @@ _Returns:_ The <<backend>> associated with this event.
 std::vector<event> get_wait_list()
 ----
 
-Deprecated by SYCL 2020.
+Deprecated by SYCL {SYCL_VERSION}.
 
 _Returns:_ The list of events that this event waits for in the dependence graph.
 Only direct dependencies are returned, and not transitive dependencies that


TApplencourt (approved):

TApplencourt:
@VerenaBeckham If I remenber correctly, you suggested to  wait for Sycl NEXT for the deprecation?

VerenaBeckham:
> @VerenaBeckham If I remenber correctly, you suggested to wait for Sycl NEXT for the deprecation?

Yes, I would say that we don't usually deprecate features in hindsight and if we do that then there will not be a lot of time for people to give feedback before the feature is removed (a few months at best). It seems the implementation burden is 0, so I don't see any rush in removing the feature.

gmlueck:
This PR is against the SYCL-Next branch.  I suggest we approve this PR (including the deprecation note).  However, when we cherry-pick it to the SYCL-2020 branch, we do not cherry-pick the note about deprecation.

VerenaBeckham:
> This PR is against the SYCL-Next branch. I suggest we approve this PR (including the deprecation note). However, when we cherry-pick it to the SYCL-2020 branch, we do not cherry-pick the note about deprecation.

But we would have to reword the deprecation notice to refer to SYCL Next? (When we agree on a name for the release we'd have to search & replace.)

gmlueck:
> But we would have to reword the deprecation notice to refer to SYCL Next? (When we agree on a name for the release we'd have to search & replace.)

Oh, good point.  Fixed in 1e2fcdb802a9ff0664c7739cc6550ad616deaf5b.  We use the asciidoc attribute `SYCL_VERSION` for this purpose.  It expands to "NEXT" currently, and we will replace the attribute definition with some number when we're ready to publish the spec.

VerenaBeckham (approved):

TApplencourt:
Thanks a lot.

- No CTS Change Required (as no `get_wait_list` test where using `In-Order-queue`)
- Merge in SYCL Next
- For sycl 2020, don't add the deprecation warning.

PR #1032, opened 2026-07-28
Clarify atomics memory model

VerenaBeckham:
Intended to replace https://github.com/KhronosGroup/SYCL-Docs/pull/849. Based on text from OpenCL: https://gitlab.khronos.org/opencl/OpenCL-Docs/-/merge_requests/226.

### DIFF
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1086,6 +1086,22 @@ values:
     the memory allocation containing the referenced object, as defined by the
     capabilities of <<buffer,buffers>> and <<usm>>.
 
+The memory scopes are listed above from narrowest
+([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
+
+The complete set of memory scopes is not guaranteed to be supported by every
+device.
+The set of supported memory scopes can be queried via the information
+descriptors for the [code]#sycl::device# and [code]#sycl::context# classes.
+
+The widest scope that can be applied to an atomic operation corresponds to the
+set of work-items which can access the associated memory location.
+For example, the widest scope that can be applied to atomic operations in
+work-group local memory is [code]#sycl::memory_scope::work_group#.
+If a wider scope is supplied, the behavior is as-if the narrowest scope
+containing all work-items which can access the associated memory location was
+supplied.
+
 Atomic operations on the same memory location must have inclusive scope,
 otherwise the behaviour is undefined.
 Atomic operations _A_ and _B_ are defined to have an inclusive scope if:
@@ -1106,22 +1122,6 @@ Atomic operations _A_ and _B_ are defined to have an inclusive scope if:
 In SYCL 2020 the memory scope of _A_ and _B_ must additionally be the same for
 the operations to have an inclusive scope.
 From SYCL {SYCL_VERSION} the memory scope of _A_ and _B_ may be different.
-
-The memory scopes are listed above from narrowest
-([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
-
-The complete set of memory scopes is not guaranteed to be supported by every
-device.
-The set of supported memory scopes can be queried via the information
-descriptors for the [code]#sycl::device# and [code]#sycl::context# classes.
-
-The widest scope that can be applied to an atomic operation corresponds to the
-set of work-items which can access the associated memory location.
-For example, the widest scope that can be applied to atomic operations in
-work-group local memory is [code]#sycl::memory_scope::work_group#.
-If a wider scope is supplied, the behavior is as-if the narrowest scope
-containing all work-items which can access the associated memory location was
-supplied.
 
 [NOTE]
 ====
--- a/adoc/chapters/architecture.adoc
+++ b/adoc/chapters/architecture.adoc
@@ -1086,6 +1086,24 @@ values:
     the memory allocation containing the referenced object, as defined by the
     capabilities of <<buffer,buffers>> and <<usm>>.
 
+Potentially concurrent conflicting actions with different memory scopes may lead
+to a data race, resulting in undefined behavior.
+An atomic operation _A_ with scope _S~1~_ operating on the same memory location
+as atomic operation _B_ with scope _S~2~_ is a data race if:
+
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~1~_; or
+* The work-items which executed _A_ and _B_ are not both in the same group of
+ work-items associated with scope _S~2~_.
+
+An atomic operation _A_ with scope _S~1~_ can only synchronize with another
+atomic operation _B_ with scope _S~2~_ if:
+
+* The work-items which executed _A_ and _B_ are both in the same group of
+  work-items associated with scope _S~1~_; and
+* The work-items which executed _A_ and _B_ are both in the same group of
+  work-items associated with scope _S~2~_.
+
 The memory scopes are listed above from narrowest
 ([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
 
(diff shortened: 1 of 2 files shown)


adoc/chapters/architecture.adoc
```diff
@@ -1086,6 +1086,27 @@ values:
...
+otherwise the behaviour is undefined.
+Atomic operations _A_ and _B_ are defined to have an inclusive scope if:
+
+  * _A_ and _B_ are executed by work-items within the same sub-group and the
+    memory scope of _A_ and _B_ is [code]#sycl::memory_scope::sub_group# or
+    wider.
+  * _A_ and _B_ are executed by work-items within the same work-group and the
+    memory scope of _A_ and _B_ is [code]#sycl::memory_scope::work_group# or
+    wider.
+  * _A_ and _B_ are executed by work-items in the same kernel instance, and the
```
VerenaBeckham, line 1115: OpenCL also talks of "child-kernel of the kernel instance" - I presume we don't have this feature?

adoc/chapters/architecture.adoc
```diff
@@ -1106,10 +1127,9 @@ supplied.
 ====
 The addition of memory scopes to the {cpp} memory model modifies the definition
 of some concepts from the {cpp} core language.
-For example: data races, the synchronizes-with relationship and sequential
-consistency must be defined in a way that accounts for atomic operations with
-differing (but compatible) scopes, in a manner similar to the <<opencl20, OpenCL
-2.0 specification>>.
+For example: sequential consistency must be defined in a way that accounts for
+atomic operations with differing (but compatible) scopes, in a manner similar to
+the <<opencl20, OpenCL 2.0 specification>>.
```
VerenaBeckham: I don't know if there is a better way to refer to the OpenCL spec, which I believe is also a combined spec? Or should we just add a reference to the 3.1 spec?

TApplencourt: I think a link to `3.3.7. Memory Consistency Model for OpenCL 2.x` maybe the cleaner indeed .

adoc/chapters/architecture.adoc
TApplencourt: Yes, we don't have this feature AFAIK.

--- 
From the OpenCL spec glossary: https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#_glossary


> Device-side enqueue

> A mechanism whereby a kernel-instance is enqueued by a kernel-instance running on a device without direct involvement by the host program. This produces nested parallelism; i.e. additional levels of concurrency are nested inside a running kernel-instance. The kernel-instance executing on a device (the parent kernel) enqueues a kernel-instance (the child kernel) to a device-side command-queue. Child and parent kernels execute asynchronously though a parent kernel does not complete until all of its child-kernels have completed.

adoc/chapters/architecture.adoc
```diff
@@ -1086,6 +1086,27 @@ values:
...
+    scope of _A_ and _B_ is [code]#sycl::memory_scope::device# or wider.
+  * _A_ and _B_ are executed by a host thread or by work-items, and the memory
+    scope of _A_ and _B_ is [code]#sycl::memory_scope::system#.
+
+In SYCL 2020 the memory scope of _A_ and _B_ must additionally be the same for
+the operations to have an inclusive scope.
+From SYCL {SYCL_VERSION} the memory scope of _A_ and _B_ may be different.
+
 The memory scopes are listed above from narrowest
 ([code]#memory_scope::work_item#) to widest ([code]#memory_scope::system#).
```
gmlueck: The new text that you added disrupts the flow of this section.  For example, this paragraph starts with "The memory scopes are listed above ...", and it refers to the list of memory scopes starting on line 1076.  That was clear before because this list used to be immediately before this paragraph.  However, your new text is new in between, so the word "above" is no longer clear.

I suggest moving your new text to the place I indicate below.

adoc/chapters/architecture.adoc
gmlueck: I would add the new text here, immediately before the NOTE.

adoc/chapters/architecture.adoc
VerenaBeckham: To be honest, I'm not sure what this sentence is trying to convey other than a TODO for the SYCL WG. Similarly, the following sentence is a promise of something we are still not doing. Is anyone opposed to just removing both lines?

adoc/chapters/architecture.adoc
VerenaBeckham: Good point. Done.

adoc/chapters/architecture.adoc
gmlueck: I agree that this whole NOTE is a TODO for the WG.  What do we need to do to address it?  Since this PR is already addressing some memory model things, can it be expanded to address the remaining issues in this NOTE?

adoc/chapters/architecture.adoc
```diff
@@ -1102,16 +1102,31 @@ If a wider scope is supplied, the behavior is as-if the narrowest scope
...
 ====
 The addition of memory scopes to the {cpp} memory model modifies the definition
 of some concepts from the {cpp} core language.
-For example: data races, the synchronizes-with relationship and sequential
-consistency must be defined in a way that accounts for atomic operations with
-differing (but compatible) scopes, in a manner similar to the <<opencl20, OpenCL
-2.0 specification>>.
-Efforts to formalize the memory model of SYCL are ongoing, and a formal memory
-model will be included in a future version of the SYCL specification.
 ====
```
TApplencourt, line 1134: Move the `note` to `issue`. 
Thanks!

adoc/chapters/architecture.adoc
VerenaBeckham: As per WG discussion I have removed the rest of the note. I opened https://github.com/KhronosGroup/SYCL-Docs/issues/1041 to track addressing the TODOs described in it.

adoc/chapters/architecture.adoc
VerenaBeckham: Done: https://github.com/KhronosGroup/SYCL-Docs/issues/1041

adoc/chapters/architecture.adoc
```diff
@@ -1102,17 +1102,26 @@ If a wider scope is supplied, the behavior is as-if the narrowest scope
...
+  * _A_ and _B_ are executed by work-items in the same kernel instance, and the
+    memory scope of _A_ and _B_ is [code]#sycl::memory_scope::device# or wider.
+  * _A_ and _B_ are executed by work-items on the same device, and the memory
+    scope of _A_ and _B_ is [code]#sycl::memory_scope::device# or wider.
+  * _A_ and _B_ are executed by a host thread or by work-items, and the memory
+    scope of _A_ and _B_ is [code]#sycl::memory_scope::system#.
+
+In SYCL 2020 the memory scope of _A_ and _B_ must additionally be the same for
+the operations to have an inclusive scope.
+From SYCL {SYCL_VERSION} the memory scope of _A_ and _B_ may be different.
```
gmlueck, line 1124: I wonder if we should add a non-normative note here to explain the rationale.  Something like:

> [Note: The SYCL definition of inclusive scope follows the OpenCL definition.  In OpenCL versions prior to 3.1, the memory scopes of _A_ and _B_ must be the same for their scopes to be inclusive, but this was relaxed in OpenCL version 3.1.  As a result, implementations of SYCL 2020 using an OpenCL backend may use the older OpenCL semantic, but implementations of {SYCL_VERSION} must use the OpenCL 3.1 semantic. -- end note]

VerenaBeckham: I added this. @gmlueck is there a way of highlighting a special technical term in the spec, like in italic? "inclusive scope" would perhaps benefit from something like that, if we have it.

gmlueck: I think I'd just put double quotes around "inclusive scope" to denote it as a technical term.  However, I would use italics for the letters A and B because they are italicized in the text above.  (The underbars are the Asciidoc markup for italics.)

Since we're talking about formatting ... can you change this to use the new-style non-normative note format?  To do this, use the `{note}` and `{endnote}` attributes.  Search for "endnote" and you will see many examples.

VerenaBeckham: Ah, I was not aware of the new note formatting, thanks! Made all these changes now.

gmlueck (approved):

TApplencourt (approved):

TApplencourt:
Merge as clarification of SYCL 2020 except 
> From SYCL {SYCL_VERSION} the memory scope of _A_ and _B_ may be different.

Which is a SYCL Next change.

PR #1044, opened 2026-08-22
Addition of explicit noexcept in common semantics

### LINKED ISSUE #1042: Explicit addition of `noexcept` for common functions
### Specification Version

SYCL 2020 (Revision 12)

### Section Number(s)

3.5.2. Common reference semantics
3.5.3. Common by-value semantics

### Issue Description

I would like to make a request to specifications for adding the `noexcept` keyword for the following:

1. All `~T()`
* Why?

* I believe it's good standard practice to mark destructors as noexcept as it addresses issues with stack unwinding for exceptions and solves the issue where you have multiple exceptions at once. If I recall correctly, a non-explicitly noexcept default constructor gets the noexcept trait either way from all modern compilers, but I think it's a good addition to add it explicitly nevertheless

2. For functions of classes that adhere to `Common reference semantics` & `Common by-value semantics`
* Why?

* For the `move constructor` & `move assignment operator`, I believe it would be beneficial to add the `noexcept` keyword, I believe that classes that adhere to `common reference semantics` would benefit more from this. The `STL` uses the following [move_if_noexcept](https://en.cppreference.com/cpp/utility/move_if_noexcept) function when deducing if a move can be performed instead of a copy where preferable, `noexcept` is needed to keep the strong exception guarantee of containers

I hope `clarification` issue type is the appropriate type for this. If it is desired for this addition to be accepted in the specifications and it's possible for me to contribute, I would very much like to do so!

Any comments are welcomed!
Thanks!

### Code Example (Optional)

```cpp
~T() noexcept = default;

T(T&&) noexcept;
T& operator=(T&&) noexcept;
```

Robertkq:
fixes #1042 
This PR adds the explicit `noexcept` to the following sections:
* 3.5.2. Common reference semantics
* 3.5.3. Common by-value semantics

for the following functions:
```cpp
T(T &&rhs) noexcept;
T &operator=(T &&rhs) noexcept;
```

The PR also contains some additional changes to make the tables prettier, it fixes alignment and just looks better overall but I can remove these changes / commit them for another PR if needed, otherwise, if these additional changes are welcomed, I'd also like to recommend either the removal or addition for both tables to contain the comments:
```cpp
  // If any of the following five special member functions are declared,
  // then all five of them should be explicitly declared (see rule of
  // five).
  //
  // Otherwise, none of them should be explicitly declared
  // (see rule of zero).
  ```
  
  I've done some analysis on the classes that use both by-value and by reference semantics and I found none that overwrite the behavior of the move constructor & assignment operator
  
  > This is my first PR to SYCL-Docs, please let me know if there are any inconsistencies between what is normally expected from contributors' PRs and mine, thanks!

### CODE REVIEW

**Comment by Robertkq** [comment]
Kindly pinging for reviews, @TApplencourt @gmlueck @tahonermann 

Thanks!

**Comment by TApplencourt** [comment]
Hi,

Sorry for the late reply. Amazing, thanks a lot! 

We will try to discuss it soon during our weekly call.  I will let you know ASAP,

Regards,
Thomas

File: adoc/headers/common-byval.h [hunk 1]
```diff
@@ -4,35 +4,32 @@
...
-      ...
+  // ...
 
-      friend bool
-      operator==(const T& lhs, const T& rhs) { /* ... */
-  }
+  friend bool operator==(const T &lhs, const T &rhs);
 
-  friend bool operator!=(const T& lhs, const T& rhs) { /* ... */ }
+  friend bool operator!=(const T &lhs, const T &rhs);
```
gmlueck: Our editorial style is to put the space after the `&`, not before.  So the original spacing was correct.  This line should be:

```suggestion
  friend bool operator!=(const T& lhs, const T& rhs);
```

The synopsis in the text has the correct format.  Could you change the spacing for the other declarations in this header and also "common-reference.h" to match this style?

Robertkq: I've changed this to the preferred editorial style. A small observation about this, I believe that the repository could benefit from a `.clang-format` file, just so that source code files that are used as reference can be modified with a higher confidence that coding style will remain consistent, don't think a CI job or anything should be needed to enforce it necessarely, but it would bring more convenience.

Anyways, hope it's good as is

gmlueck (approved):

adoc/headers/common-byval.h
gmlueck: @TApplencourt tried this once, but it changed a bunch of other formatting in ways we didn't like.  Maybe it's possible to use clang-format in a way that doesn't do this, but we didn't discover that way.

In any case, thanks for the updates!

TApplencourt: Yes, I tried but I failed so we removed it.  But I guess now with AI, maybe easier to make it iterate over all the 1000 clang-format combination to find the one who doesn't change too much our current code :) 

Thanks again! 

LGTM. So our next discussion will be:
- Should be back port this to SYCL 2020
- What whould be we do will all the `noexcept` function who have precondition

adoc/headers/common-reference.h
```diff
@@ -4,26 +4,25 @@
...
 
   T& operator=(const T& rhs);
 
-  T& operator=(T&& rhs);
+  T& operator=(T&& rhs) noexcept;
 
   ~T();
 
-  ...
+  // ...
```
TApplencourt, line 20: Thanks!  : ) I also prefer when our `.h` are valid c++

TApplencourt (approved):

adoc/headers/common-byval.h
gmlueck: > Should be back port this to SYCL 2020

We looked at the DPC++ implementation, and we do not see any problem add these `noexcept` to SYCL 2020.  Therefore, we'd be OK adopting this as a clarification / errata.

Robertkq:
Is there anything else required from my side for this PR to be merged? I see that there is a workflow waiting for approval.

Really happy to contribute to SYCL-Docs, if I see any other possible improvements or inconsistencies, I will file an issue and PR! Thanks for the approvals

UPD: didn't read latest comments, I'll wait for you guys to have the meeting and discuss it, no rush from my side! Out of curiosity, is this a public meeting?

adoc/headers/common-byval.h
Robertkq: > We looked at the DPC++ implementation, and we do not see any problem add these `noexcept` to SYCL 2020. Therefore, we'd be OK adopting this as a clarification / errata.

Does this require for this PR to be duplicated on the SYCL-2020 branch?

Also, out of curiosity, when is the expected release of SYCL-Next? Correct me if I am wrong, but SYCL-Next isn't just the next revision update for SYCL-2020, right? It's a separate set of specifications that up to date SYCL implementations will need to follow in the future?

gmlueck: > Does this require for this PR to be duplicated on the SYCL-2020 branch?

No.  If the WG approves this for SYCL 2020, I will make a separate PR to the SYCL 2020 branch.

TApplencourt:
>Really happy to contribute to SYCL-Docs, if I see any other possible improvements or inconsistencies, I will file an issue and PR! Thanks for the approvals

Amazing! Thanks again, always good to have another pair of eyes for things we missed ! :)  

> Out of curiosity, is this a public meeting?

Sadly, no. It's restricted to the member of Khronos :(  But your company are most than welcome to join : ) (https://www.khronos.org/members/). 

We have also a SYCL Advisory Panel this one is free to join if you are interested.

adoc/headers/common-byval.h
TApplencourt: > It's a separate set of specifications that up to date SYCL implementations will need to follow in the future?

Exactly. Like a new version. In a new SYCL version, we are allowed to do some breaking change (that we will try to minimize ofc), and where some deprecated feature will be removed,  etc  

In new rev, we just "back-port" bug fixes.

Robertkq:
> We have also a SYCL Advisory Panel this one is free to join if you are interested.

I am contributing independently to SYCL, so I will try to join / request to join the SYCL Advisory Panel! Thanks for letting me know!

TApplencourt:
SYCL Next: No objection so we can merge the PR
SYCL 2020 as a bug fix: Merge it into SYCL 2020 as a bug fixes. 

- CTS change required (we need to hold for change) 
  - https://github.com/KhronosGroup/SYCL-CTS need to be extended to verify `no-except` for the common-reference  and common-by-value semantic

Robertkq:
Happy to hear this :)
As far as I understand, I am not responsible to duplicate the PR on the SYCL2020 branch, although I don't mind if necessary.

About CTS change, I'd be more than happy to look into this, might take a bit in order to get familiar enough to add the checks properly but I hope to contribute more in the future so I wouldn't mind learning. Is it expected / OK if I take the responsability of adding the CTS checks?

nliber:
Thought (and it would be a separate PR): do we want to make the comparison operators `noexcept` also?  It would be weird for those operations to throw, as they are both non-modifying and shouldn't have preconditions.

nliber (approved):

Robertkq:
> Thought (and it would be a separate PR): do we want to make the comparison operators `noexcept` also? It would be weird for those operations to throw, as they are both non-modifying and shouldn't have preconditions.

I also believe they shouldn't have preconditions, for which then it would be easy to add `noexcept` and keep that promise, as it's ABI changing. Would you like to move forward with this and open a new issue yourself? If not, I do not mind looking more into it an opening a new issue.

Also thanks for the review!

PR #1048, opened 2026-09-01
[Editorial] Fix mention of the removed function

KornevNikita:
The `group` class doesn't have the `get_linear_id()` function. It's not clear how to correctly update this sentence. Given that this sentence is an example for deprecated function, I've simply dropped this mention.


adoc/chapters/programming_interface.adoc
```diff
@@ -13716,7 +13716,7 @@ representing the work-item computation.
 
 Note that the logical range does not need to be uniform
 across all work-groups in a kernel.  For example the logical range may depend on
-a work-group varying query (e.g. [code]#group::get_linear_id)#,
+a work-group varying query (e.g. [code]#group::get_group_linear_id)#,
```
gmlueck: I agree there is no `group::get_linear_id`, so there is some typo here.  Maybe the intent was `get_group_linear_id` as you have here, but there is also `get_local_linear_id`.  To be honest, neither makes any sense to me.

Assuming the intent was `get_group_linear_id`, why would you want to define the size of the logical local range in terms of the physical range's position in the global nd-range?

KornevNikita: Yes, that's confusing. Should we just drop this?
```suggestion
a work-group varying query,
```

gmlueck: Does anyone in the SYCL committee understand the intent here?  @illuhad, you've been on the committee longer than me, and I think you were once a supporter of these hierarchical APIs?  Do you understand the intent of this example?

TApplencourt: Oh good catch! 

I agree with @KornevNikita pragmatic suggestion to drop it. It's a example, inside a node, inside a deprecated API.

I thinks it's a relicate of SYCL 1.2.1. If my archeology is correct in 1.2.1  `group::get_linear_id ` existed . Also the marckup is wrong ( the `#` and `)` )

I think the indent what to show that this code is possible:
```
cgh.parallel_for_work_group(range<1>(2), [=](group<1> g) {
  g.parallel_for_work_item(range<1>(g.get_group_linear_id() ? 4 : 8),
                           [=](h_item<1> it) { /* ... */ });
});
```
So for `parallel_for_work_item`,  `range` can different for each work-group. If we really want we can said: 

```
Note that the logical range does not need to be the same for every work-group in
a kernel.  For example, the logical range may be computed from the work-group's
own id (e.g. [code]#group::get_group_linear_id#), so that different work-groups
in the same kernel invocation execute different logical range sizes.
```

( and to be honest, I don't know if we should said `kernel` or `SYCL kernel function`, or  even `kernel handler`)

KornevNikita: > It's a example, inside a node, inside a deprecated API.

I've proceeded with drop e13e4b2ffc6afad697e24fd47795d894df87388b

gmlueck (approved):

TApplencourt (approved):

VerenaBeckham (approved):

TApplencourt:
WG Agreed to merge as a bug fix in SYCL 2020. 
Thanks!

